PR #4, opened 2018-11-07
Add no-op check to Chuck utility class

olivierperez:
Add a no-op check in `Chuck` utilities class, which can be used to check if we currently have a no operation instance of chuck.

### DIFF
--- a/library-no-op/src/main/java/com/readystatesoftware/chuck/Chuck.java
+++ b/library-no-op/src/main/java/com/readystatesoftware/chuck/Chuck.java
@@ -26,4 +26,9 @@ public class Chuck {
     public static Intent getLaunchIntent(Context context) {
         return new Intent();
     }
+
+    public static boolean isNoOp() {
+        return true;
+    }
+
 }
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/readystatesoftware/chuck/api/Chuck.java
```diff
@@ -69,6 +69,15 @@ public static void dismissErrorsNotification() {
         new NotificationHelper(context).dismissErrorsNotification();
     }
 
+    /**
+     * Check if this instance is the operation one or no-op.
+     *
+     * @return true if this is a the operation instance.
```
cortinico: Typo here:

```suggestion
     * @return true if this is the operation instance.
```

olivierperez: Ok

cortinico:
@olivierperez apologize for the delay. I was sure I already approved those PRs. Looks like I hadn't 😅

PR #6, opened 2018-11-07
Issue #43 Ask for confirmation when delete history

olivierperez:
Show an AlertDialog when user try to delete HTTP calls or Errors to confirm his action.

### DIFF
--- a/library/src/main/java/com/readystatesoftware/chuck/internal/ui/error/ErrorListFragment.java
+++ b/library/src/main/java/com/readystatesoftware/chuck/internal/ui/error/ErrorListFragment.java
@@ -1,6 +1,7 @@
 package com.readystatesoftware.chuck.internal.ui.error;
 
 import android.content.Context;
+import android.content.DialogInterface;
 import android.database.Cursor;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
@@ -9,6 +10,7 @@ import android.support.v4.app.Fragment;
 import android.support.v4.app.LoaderManager;
 import android.support.v4.content.CursorLoader;
 import android.support.v4.content.Loader;
+import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.DividerItemDecoration;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/readystatesoftware/chuck/internal/ui/error/ErrorListFragment.java
```diff
@@ -97,6 +98,24 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        new AlertDialog.Builder(getContext())
+            .setTitle(R.string.chuck_clear)
+            .setMessage(R.string.chuck_clear_error_confirmation)
+            .setPositiveButton(R.string.chuck_clear, new DialogInterface.OnClickListener() {
+                @Override
+                public void onClick(DialogInterface dialog, int which) {
+                    getContext().getContentResolver().delete(ChuckContentProvider.ERROR_URI, null, null);
+                }
+            })
+            .setNegativeButton(R.string.chuck_cancel, new DialogInterface.OnClickListener() {
```
cortinico: You can replace here with

```suggestion
            .setNegativeButton(R.string.chuck_cancel, null)
```

And remove all the lines below

library/src/main/java/com/readystatesoftware/chuck/internal/ui/transaction/TransactionListFragment.java
```diff
@@ -126,6 +127,25 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+            .setTitle(R.string.chuck_clear)
+            .setMessage(R.string.chuck_clear_http_confirmation)
+            .setPositiveButton(R.string.chuck_clear, new DialogInterface.OnClickListener() {
+                @Override
+                public void onClick(DialogInterface dialog, int which) {
+                    getContext().getContentResolver().delete(ChuckContentProvider.TRANSACTION_URI, null, null);
+                    NotificationHelper.clearBuffer();
+                }
+            })
+            .setNegativeButton(R.string.chuck_cancel, new DialogInterface.OnClickListener() {
```
cortinico: As before:

```suggestion
            .setNegativeButton(R.string.chuck_cancel, null)
```

library/src/main/java/com/readystatesoftware/chuck/internal/ui/error/ErrorListFragment.java
olivierperez: Ok

library/src/main/java/com/readystatesoftware/chuck/internal/ui/transaction/TransactionListFragment.java
olivierperez: Ok

PR #8, opened 2018-11-08
Udpate Gradle version to 4.10.1

cortinico:
Time to modernize Gradle a bit 😅


olivierperez:
Ho! I've just look at the release notes, there is a 4.10.2. Let's go?

PR #10, opened 2018-11-11
Updating the ListItem layout to use CL

cortinico:
I've refactored a bit the layout of the list items:

* Now uses a `ConstraintLayout` instead of a `RelativeLayout` (+ nested layouts)
* Time TextView had a static size and was wrapping the line on my device (annyoing and causing extra space to be used).
* Fixed the alignment so now the list will look more "ordered"
* Added color for the SSL icon 
* Updated IDs to have `chucker_` as a prefix

Regarding the last point, we should probably do it for all the ids otherwise will create a mess on IDs of client projects


library/src/main/res/layout/chucker_list_item_transaction.xml
```diff
@@ -0,0 +1,90 @@
...
+
+    <TextView
+        android:id="@+id/chucker_host"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="@dimen/chucker_half_grid"
+        app:layout_constraintBottom_toTopOf="@+id/chucker_ssl"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toEndOf="@+id/chucker_ssl"
+        app:layout_constraintTop_toBottomOf="@+id/chucker_ssl"
```
olivierperez: I guess you mean
```xml
app:layout_constraintBottom_toBottomOf="@+id/chucker_ssl"
app:layout_constraintTop_toTopOf="@+id/chucker_ssl"
```

library/src/main/res/layout/chucker_list_item_transaction.xml
```diff
@@ -0,0 +1,90 @@
...
+        tools:text="example.com" />
+
+    <ImageView
+        android:id="@+id/chucker_ssl"
+        android:layout_width="@dimen/chucker_doub_grid"
+        android:layout_height="@dimen/chucker_doub_grid"
+        android:contentDescription="@string/chuck_ssl"
+        android:src="@drawable/chuck_ic_https_grey_24dp"
+        android:tint="@color/chuck_colorPrimary"
+        android:visibility="gone"
```
olivierperez: Can you add `tools:visibility="visibile"`, to see the icon in editor.

library/src/main/res/layout/chucker_list_item_transaction.xml
cortinico: Yup :)

cortinico:
@olivierperez now should be ready to go

PR #12, opened 2018-11-12
Prepare 2.1.0

olivierperez:
I set the version number to **2.1.0-SNAPSHOT**, it will allow us to publish unstable versions _if needed_.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -16,9 +16,9 @@ org.gradle.jvmargs=-Xmx1536m
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 
-VERSION_NAME=2.0.1
-# 2*10*10 + 0*10 + 1 => 20001
-VERSION_CODE=20001
+VERSION_NAME=2.1.0-SNAPSHOT
+# 2*100*100 + 1*100 + 0 => 20100
+VERSION_CODE=20100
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,8 +30,8 @@ repositories {
 
 ```gradle
 dependencies {
-  debugImplementation 'fr.o80.chucker:library:2.0.1'
-  releaseImplementation 'fr.o80.chucker:library-no-op:2.0.1'
+  debugImplementation 'fr.o80.chucker:library:2.1.0'
```
cortinico: Are you sure we want to update this as well?
If we're publishing `-SNAPSHOT` versions on maven, the stable version (`2.1.0`) will not be yet available right?

I'd rather keep `-SNAPSHOT` only on `gradle.properties` and we update the README as soon as we release a new version

olivierperez: I've merged your work about Jitpack+SNAPSHOT, what do I have to do with this change?

cortinico: So the idea is to merge this, as soon as we're ready with `2.1.0` right?
If so, just remove the `-SNAPSHOT` from the `VERSION_NAME` field and everything will be fine.

olivierperez: Ok, I've removed the "-SNAPSHOT" part of the version in `gradle.properties`.

olivierperez:
The next version released from `develop` (for improvements) will be **2.1.0**, and the next from `master` (for fixes) will be **2.0.3**.

PR #21, opened 2018-11-27
Polish the MainActivity

cortinico:
* Updated Main Activity to use Kotlin only
* Added an intro message/title
* Tweaked the layout a bit


sample/src/main/java/com/readystatesoftware/chuck/sample/MainActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+import com.readystatesoftware.chuck.api.RetentionManager
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+
+class MainActivity : AppCompatActivity() {
+
+    private var collector: ChuckCollector? = null
```
olivierperez: You may use a `lateinit` here, collector is created in the onCreate, and is not reset to null ever.

sample/src/main/java/com/readystatesoftware/chuck/sample/MainActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_main)
+
+        findViewById<View>(R.id.do_http).setOnClickListener { doHttpActivity() }
```
olivierperez: Can we use `kotlin-android-extensions` to access to views?

olivierperez:
Ok for me, let's fix conflicts and go!

PR #22, opened 2018-11-28
Modernize the project icon

cortinico:
* Updated the sample app icon to use adaptive icons
* Updated the notification icon to do NOT use `VectorDrawable` as we're supporting minSdk 16 (all the other icons should be updated as well).
* Rebalanced the notification icon size (was smaller than a usual notification icon).

I just turned the icon 45' degree on the side. I'm happy to other suggestions if you have some :)
ic_launcher-web


cortinico:
> Maybe prefix "ic_launcher_*" with "chucker" too.

@olivierperez That's not really needed as the `sample` module (the app) is not published so the resource are not exported. I can do it for the sake of _clarity_ but is not really needed.

olivierperez:
You're right :wink: 

Let's fix the conflicts, and go!

PR #24, opened 2018-11-28
Redact header (from okhttp logging)

alorma:
Hi.

I've ported this feature from OkHttp Logging Interceptor:

https://github.com/square/okhttp/tree/master/okhttp-logging-interceptor

Now, there's an option to **redact** / **remove** a specific header content and replace by an empty string, so sensitive information is not set on the database


library/src/main/java/com/readystatesoftware/chuck/api/ChuckInterceptor.java
```diff
@@ -73,6 +80,14 @@ public ChuckInterceptor maxContentLength(long max) {
         return this;
     }
 
+    public ChuckInterceptor redactHeader(String name) {
+        Set<String> newHeadersToRedact = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        newHeadersToRedact.addAll(headersToRedact);
+        newHeadersToRedact.add(name);
+        headersToRedact = newHeadersToRedact;
+        return this;
```
olivierperez: Why don't you just `headersToRedact.add(name)`?

library/src/main/java/com/readystatesoftware/chuck/api/ChuckInterceptor.java
```diff
@@ -73,6 +80,14 @@ public ChuckInterceptor maxContentLength(long max) {
         return this;
     }
 
+    public ChuckInterceptor redactHeader(String name) {
```
olivierperez: Some developers may need to redact multiple headers, we should have this method instead:
`public ChuckInterceptor redactHeaders(String names...) {`

library/src/main/java/com/readystatesoftware/chuck/api/ChuckInterceptor.java
alorma: Done

library/src/main/java/com/readystatesoftware/chuck/api/ChuckInterceptor.java
alorma: Done

library/src/main/java/com/readystatesoftware/chuck/api/ChuckInterceptor.java
olivierperez: Add it to **library-no-op** too please :wink:

alorma:
rebased against developm fixed `gradle.properties` conflict

PR #29, opened 2019-01-14
Replacing Cupboard with Room

cortinico:
Took some time but I managed to remove Cupboard completely :)
Unfortunately the Pull Request is pretty big :( Let me know if I should split it somehow to make it easier to review.

Summary of changes:

- Removed all the usage 
- Removed the Content Provider
- Removed the DbOpenHelper
- Updated entities to be Kotlin Data classes
- Introduced the Repository pattern (this will allow other devs to swap Room with something else, e.g. Realm, InMemoryRepository, etc.).
- Removed several static refences to Context 🤣
- Removed all the nested CursorAdapters


library/src/main/java/com/readystatesoftware/chuck/internal/data/entity/HttpTransaction.kt
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    private fun toHttpHeaderList(headers: Headers): List<HttpHeader> {
+        val httpHeaders = ArrayList<HttpHeader>()
+        var i = 0
+        val count = headers.size()
+        while (i < count) {
+            httpHeaders.add(HttpHeader(headers.name(i), headers.value(i)))
+            i++
+        }
```
olivierperez:
```kotlin
for (i in 0 until headers.size()) {
    httpHeaders.add(HttpHeader(headers.name(i), headers.value(i)))
}
```

library/src/main/java/com/readystatesoftware/chuck/internal/data/entity/HttpTransaction.kt
```diff
@@ -0,0 +1,205 @@
...
+        return httpHeaders
+    }
+
+    private fun formatBody(body: String, contentType: String?): String {
+        return if (contentType != null && contentType.toLowerCase().contains("json")) {
```
olivierperez: Can you use `when` insteadof `if` ?

library/src/main/java/com/readystatesoftware/chuck/internal/data/repository/ChuckerDatabaseRepository.kt
```diff
@@ -0,0 +1,67 @@
...
+        executor.execute { database.throwableDao().deleteAllThrowables() }
+    }
+
+    override fun getSortedThrowablesTuples(): LiveData<List<RecordedThrowableTuple>> {
+        return database.throwableDao().getSortedThrowablesTuples()
+    }
+
+    private val executor: Executor = Executors.newSingleThreadExecutor()
+
+    override fun insertTransaction(transaction: HttpTransaction) {
```
olivierperez: I'm not really fan of this kind of method that have a side effect (here, changing `transaction.id`). For later we should think about it.

library/src/main/java/com/readystatesoftware/chuck/internal/data/repository/ChuckerRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Repository Interface reprensenting all the operations that are needed to let Chucker work
+ * properly. Please use [ChuckerDatabaseRepository] that uses Room and SqLite to run those
+ * operations.
+ */
+internal interface ChuckerRepository {
+
+    /*
+     * CRUD for HttpTransaction[Tuple]
```
olivierperez: Can we split the Repository in 2 ? Transaction VS Throwable

library/src/main/java/com/readystatesoftware/chuck/internal/data/room/ChuckerDatabase.kt
```diff
@@ -0,0 +1,26 @@
+package com.readystatesoftware.chuck.internal.data.room
+
+import android.arch.persistence.room.Database
+import android.arch.persistence.room.Room
+import android.arch.persistence.room.RoomDatabase
+import android.content.Context
+import com.readystatesoftware.chuck.internal.data.entity.HttpTransaction
+import com.readystatesoftware.chuck.internal.data.entity.RecordedThrowable
+
+@Database(entities = [RecordedThrowable::class, HttpTransaction::class], version = 1, exportSchema = false)
```
olivierperez: The database version is going back to `1`? What about the previous data?

library/src/main/java/com/readystatesoftware/chuck/internal/data/room/RecordedThrowableDao.kt
```diff
@@ -0,0 +1,28 @@
...
+import android.arch.persistence.room.Insert
+import android.arch.persistence.room.Query
+import com.readystatesoftware.chuck.internal.data.entity.RecordedThrowable
+import com.readystatesoftware.chuck.internal.data.entity.RecordedThrowableTuple
+
+@Dao
+internal interface RecordedThrowableDao {
+
+    @Query("SELECT id,tag,date,clazz,message FROM throwables ORDER BY date DESC")
+    fun getSortedThrowablesTuples(): LiveData<List<RecordedThrowableTuple>>
```
olivierperez:
```suggestion
    fun getTuples(): LiveData<List<RecordedThrowableTuple>>
```

library/src/main/java/com/readystatesoftware/chuck/internal/data/room/RecordedThrowableDao.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @Insert()
+    fun insertRecordedThrowables(throwable: RecordedThrowable): Long?
```
olivierperez:
```suggestion
    fun insert(throwable: RecordedThrowable): Long?
```

library/src/main/java/com/readystatesoftware/chuck/internal/data/room/RecordedThrowableDao.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @Query("DELETE FROM throwables")
+    fun deleteAllThrowables()
```
olivierperez:
```suggestion
    fun deleteAll()
```

library/src/main/java/com/readystatesoftware/chuck/internal/data/room/RecordedThrowableDao.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @Query("SELECT * FROM throwables WHERE id = :id")
+    fun getRecordedThrowable(id: Long): LiveData<RecordedThrowable>
```
olivierperez:
```suggestion
    fun getById(id: Long): LiveData<RecordedThrowable>
```

library/src/main/java/com/readystatesoftware/chuck/internal/data/room/RecordedThrowableDao.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @Query("DELETE FROM throwables WHERE date <= :threshold")
+    fun deleteOldThrowables(threshold: Long)
```
olivierperez:
```suggestion
    fun deleteBefore(threshold: Long)
```

library/src/main/java/com/readystatesoftware/chuck/internal/ui/transaction/TransactionActivity.java
```diff
@@ -118,29 +114,11 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-    }
-
-    @Override
-    public void onLoaderReset(@NonNull Loader<Cursor> loader) {
-    }
-
-    private void populateUI() {
-        if (transaction != null) {
-            title.setText(transaction.getMethod() + " " + transaction.getPath());
+    private void populateUI(HttpTransaction observed) {
```
olivierperez: `observed` => `transaction` ?

library/src/main/java/com/readystatesoftware/chuck/internal/ui/transaction/TransactionListFragment.java
```diff
@@ -128,62 +107,51 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-            .setPositiveButton(R.string.chuck_clear, new DialogInterface.OnClickListener() {
-                @Override
-                public void onClick(DialogInterface dialog, int which) {
-                    getContext().getContentResolver().delete(ChuckContentProvider.TRANSACTION_URI, null, null);
-                    NotificationHelper.clearBuffer();
-                }
-            })
-            .setNegativeButton(R.string.chuck_cancel, null)
-            .show();
+        new AlertDialog.Builder(getActivity())
```
olivierperez: Why don't you use `getContext()` instead of `getActivity()`?

library/src/main/java/com/readystatesoftware/chuck/internal/data/room/ChuckerDatabase.kt
cortinico: So for this, I won't spend much time doing migrations as this is going to be a breaking change.
We also need to change the package name so I don't expect users complaining because their transactions/throwables are not stored.

What do you think?

library/src/main/java/com/readystatesoftware/chuck/internal/data/entity/HttpTransaction.kt
```diff
@@ -0,0 +1,205 @@
...
+
+    val status: Status
+        get() = when {
+            error != null -> Status.Failed
+            responseCode == null -> Status.Requested
+            else -> Status.Complete
+        }
+
+    val requestDateString: String?
+        get() = requestDate?.toString()
```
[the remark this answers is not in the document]
cortinico: So this can easily be removed. The problem is unfortunately wider.
So far I tried to make sure `HttpTransaction.kt` had a similar api with `HttpTransaction.java`.

I'll say we can probably refactor all this stuff outside of the data class and have it inside say FormatUtils?
I'll anyway keep those refactorings outside of this PR.

library/src/main/java/com/readystatesoftware/chuck/internal/data/repository/ChuckerDatabaseRepository.kt
cortinico: Wasn't really able to find better alternatives. Here the problem is that the `id` field is auto-generated and we should update the fields. AFAIK, the auto-generated id is returned as result value of the `insert` function.

cortinico:
Hey @olivierperez so I've addressed all your comments + left a couple of questions.
The PR should be good to go IMHO

library/src/main/java/com/readystatesoftware/chuck/internal/data/room/ChuckerDatabase.kt
olivierperez: Ok for me. Actually Chucker is a dev tool, nothing of this should go to prod so we don't really care of migration.
BTW if I understand well the previous data is still there (in antoher DB), can we deleted it ? If not... don't worry, this is just a nice to have :-)

cortinico: > BTW if I understand well the previous data is still there (in antoher DB), can we deleted it ? If not... don't worry, this is just a nice to have :-)

So yeah, I pushed a commit to delete the old db: e04c085
I'm just afraid this will make Chucker incompatible with Chuck if used simultaneously. I don't really know if this is a real use-case. But probably something worth noting.

PR #35, opened 2019-02-22
Cleanup form the old Chuck names

cortinico:
Mostly done using automated tools:

* Package name is now `com.chuckerteam.chucker...`
* All the previous occurrences of Chuck have been renamed.


library/src/main/java/com/chuckerteam/chucker/api/internal/data/room/ChuckerDatabase.kt
```diff
@@ -18,7 +18,7 @@ internal abstract class ChuckerDatabase : RoomDatabase() {
         private val DB_NAME = "chucker.db"
 
         fun create(context: Context): ChuckerDatabase {
-            // We eventually delete the old DB if a previous version of Chuck/Chucker was used.
+            // We eventually delete the old DB if a previous version of Chucker/Chucker was used.
```
olivierperez:
```suggestion
            // We eventually delete the old DB if a previous version of Chucker was used.
```

library/src/main/java/com/chuckerteam/chucker/api/internal/support/SQLiteUtils.java
```diff
@@ -47,8 +47,8 @@ private static String extractDatabase(Context context) {
             File external = context.getExternalFilesDir(null);
             File data = Environment.getDataDirectory();
             if (external != null && external.canWrite()) {
-                String dataDBPath = "data/" + context.getPackageName() + "/databases/chuck.db";
-                String extractDBPath = "chuckdb.temp";
+                String dataDBPath = "data/" + context.getPackageName() + "/databases/chucker.db";
+                String extractDBPath = "chuckerdb.temp";
```
olivierperez: Do we still use this method? `extractDatabase`

cortinico: Not really, I'm removing that class. You'll find it in a commit in this PR.

library/src/main/java/com/chuckerteam/chucker/api/internal/data/room/ChuckerDatabase.kt
cortinico: Actually the original was correct.

Issue #31, opened 2019-02-03
Hide the RetentionManager in the API?


psh:
I believe that it would be possible to hide the `RetentionManager` from the public API and treat it as an implementation detail, so users would only need to supply a _retention period_ and internally the `ChuckCollector` manages an instance of it.
```kotlin
collector = ChuckCollector(this)
        .showNotification(true)
        .retentionManager(RetentionManager(this, ChuckCollector.Period.ONE_HOUR))
```
becoming 
```kotlin
collector = ChuckCollector(this)
        .showNotification(true)
        .retentionPeriod(ChuckCollector.Period.ONE_HOUR)
```
I am assuming that you would always want to use the same `context` for both the collector and the retention manager.  Can you imagine any circumstance where this wouldn't be the case and the exposed constructor would be useful?

Do you feel this would be a good way to evolve the API?

**Do you want to develop this feature yourself?**
Yes, I'd be happy to.

cortinico:
Hey @psh 

This shoulds like a good idea to me :)
Please note that we also recently migrated all the retention mechanism to Room, so it should be easier to integrate with your proposal. Would be happy/free to code this feature?

cortinico:
Closed by #36

PR #36, opened 2019-02-24
Hide the RetentionManager in the API (Issue #31)

psh:
This PR starts the process of hiding the `RetentionManager` in the public API by deprecating the existing method (with appropriate Javadoc) and internally it treats the `RetentionManager` as an implementation detail, so users would only need to supply a _retention period_ and the `ChuckCollector` manages an instance of it.  This means that the API effectively changes from
```kotlin
collector = ChuckCollector(this)
        .showNotification(true)
        .retentionManager(RetentionManager(this, ChuckCollector.Period.ONE_HOUR))
```
to
```kotlin
collector = ChuckCollector(this)
        .showNotification(true)
        .retentionPeriod(ChuckCollector.Period.ONE_HOUR)
```
I took a conservative approach (by deprecating the existing method) as I didn't know if you already had production usages of that piece of code.

### DIFF
--- a/library-no-op/src/main/java/com/chuckerteam/chucker/api/ChuckerCollector.kt
+++ b/library-no-op/src/main/java/com/chuckerteam/chucker/api/ChuckerCollector.kt
@@ -8,7 +8,7 @@ import android.content.Context
 class ChuckerCollector @JvmOverloads constructor(
     context: Context,
     var showNotification: Boolean = true,
-    var retentionManager: Any? = null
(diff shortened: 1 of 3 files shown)


cortinico:
Hey @psh 

Sorry for the really late reply.

So the API of Chucker changed significantly and we now migrated to use Kotlin + named parameters rather than a Java builder pattern. 

Can I ask you to rebase your change on top of `develop`?

psh:
I've rebased and redone the change in light of the new API.  It was easier to make the retention manager private rather than making it _deprecated_ as I initially suggested.

PR #39, opened 2019-04-16
Prevent potential XXE attacks from XML formatting

cortinico:
This PR replicates https://github.com/jgilfelt/chuck/pull/84

The fix will prevent the library from loading external resources when parsing an XML from a response body.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/internal/support/FormatUtils.java
+++ b/library/src/main/java/com/chuckerteam/chucker/api/internal/support/FormatUtils.java
@@ -1,18 +1,3 @@
-/*
- * Copyright (C) 2017 Jeff Gilfelt.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
 package com.chuckerteam.chucker.api.internal.support;
 
 import android.content.Context;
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/chuckerteam/chucker/api/internal/support/FormatUtils.java
```diff
@@ -73,7 +59,11 @@ public static String formatJson(String json) {
 
     public static String formatXml(String xml) {
         try {
-            Transformer serializer = SAXTransformerFactory.newInstance().newTransformer();
+            SAXTransformerFactory transformerFactory = (SAXTransformerFactory) SAXTransformerFactory.newInstance();
+            transformerFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+            transformerFactory.setAttribute("http://javax.xml.XMLConstants/property/accessExternalDTD", "");
+            transformerFactory.setAttribute("http://javax.xml.XMLConstants/property/accessExternalStylesheet", "");
```
olivierperez: I'm not sure to understand this 2 lines. It's like we are allowing the usage of external DTDs and Stylesheets. If I'm right, the security is still there, the transformer will download the external resources.

cortinico:
So the documentation of the properties is here:
https://docs.oracle.com/javase/tutorial/jaxp/properties/properties.html

The implementation I followed is basically the one suggested on StackOverflow here:
https://stackoverflow.com/questions/32178558/how-to-prevent-xml-external-entity-injection-on-transformerfactory
I just replaced the variables with the effective value (given that those variables are missing in the Android version of `javax.xml.XMLConstants`).

> If I'm right, the security is still there, the transformer will download the external resources.

To answer your question, with those `setAttribute` we are restricting the set of domains where the external resources (DTDs and Stylesshets are allowed). In this case we set them to the empty string so no domain will be allowed.

There is a specific explanation on the javadoc for that methods here:

https://docs.oracle.com/javase/8/docs/api/javax/xml/transform/TransformerFactory.html#setAttribute-java.lang.String-java.lang.Object-

olivierperez:
Thanks for explaination ;-) It's OK for me.

PR #40, opened 2019-05-05
Reorganize dependency versions

cortinico:
I refactored the way how we organise deps versions. Now gradle plugin and production dependencies are together. Moreover I updated some versions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         gsonVersion = '2.8.5'
         junitGradlePluignVersion = '1.3.1.1'
         junitVersion = '5.3.2'
-        kotlinVersion = '1.3.20'
+        kotlinVersion = '1.3.31'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -1,24 +1,33 @@
-// Top-level build file where you can add configuration options common to all sub-projects/modules.
-
 buildscript {
-    ext.kotlin_version = '1.3.20'
-    ext.dokka_version = '0.9.17'
+    ext {
+        androidGradleVersion = '3.4.0'
+        androidMavenGradleVersion = '2.1'
+        constraintLayoutVersion = '1.1.3'
+        cupboardVersion= '2.2.0'
```
olivierperez: Do we still use Cupboard?

cortinico: You're right I fixed it

PR #41, opened 2019-05-05
Switch Release Artifact to JitPack

cortinico:
As discussed at AndroidMakers, let's move both artifacts to JitPack.
This makes easier for everyone to release new version (Just `git tag`) and we don't have GPG keys to handle etc.

I think we just need to tag `2.0.4` as it was never tagged so is not available on JitPack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ try {
 
 # Snapshots
 
-Development of Chucker happens in the [develop](https://github.com/ChuckerTeam/chucker/tree/develop) branch. You get get `SNAPSHOT` versions directly from Jitpack if needed.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -81,7 +81,7 @@ try {
 
 # Snapshots
 
-Development of Chucker happens in the [develop](https://github.com/ChuckerTeam/chucker/tree/develop) branch. You can switch to the development branch by using artifacts available on [jitpack.io](https://jitpack.io/#ChuckerTeam/chucker)
+Development of Chucker happens in the [develop](https://github.com/ChuckerTeam/chucker/tree/develop) branch. You get get `SNAPSHOT` versions directly from Jitpack if needed.
```
olivierperez: Get get ?

cortinico:
Sure now we should be good to go

olivierperez:
Looks good to me :+1:

PR #42, opened 2019-05-13
 Kotlinize the Public API of Chucker

cortinico:
I converted to Kotlin all the public exposed classes of Chucker.
I also converted to kotlin the `no-op` package to make it easier to
maintain once we do further api changes.

IMPORTANT: Please note that this change introduce some breaking changes
and requires a major version bump.

Summary of Changes:
* Converted all the classes in `com.chuckerteam.chucker.api` to Kotlin
* The `no-op` module now is all in Kotlin and uses the Kotlin android
plugin
* Refactored the historical builder pattern of Interceptor/Collector to
use named parameters instead
* Moved the `PERIOD` enum to be inside the `RetentionManager`
* Updated visibility of multiple methods to be `internal`


library-no-op/build.gradle
```diff
@@ -19,6 +20,7 @@ artifacts {
 
 dependencies {
     implementation "com.squareup.okhttp3:okhttp:$okhttp3Version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
```
olivierperez: If we merge this PR after #40, $kotlin_version should become $kotlinVersion.

library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * No-op implementation.
+ */
+object Chucker {
+
+    val SCREEN_HTTP = 1
+    val SCREEN_ERROR = 2
+
+    val isOp: Boolean get() = false
```
olivierperez: Instead of defining the Getter, should we directly set the value ?
`val isOp: Boolean = false`

library-no-op/src/main/java/com/chuckerteam/chucker/api/ChuckerCollector.kt
```diff
@@ -0,0 +1,16 @@
+package com.chuckerteam.chucker.api
+
+import android.content.Context
+
+/**
+ * No-op implementation.
+ */
+class ChuckerCollector @JvmOverloads constructor(
+        context: Context?,
```
olivierperez: I guess Context shouldn't be nullable. Am I right?

library-no-op/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -0,0 +1,27 @@
...
+import android.content.Context
+import okhttp3.Interceptor
+import okhttp3.Response
+import java.io.IOException
+
+/**
+ * No-op implementation.
+ */
+class ChuckerInterceptor @JvmOverloads constructor(
+        context: Context?,
```
olivierperez: Nullable context?

library-no-op/src/main/java/com/chuckerteam/chucker/api/RetentionManager.kt
```diff
@@ -0,0 +1,23 @@
+package com.chuckerteam.chucker.api
+
+import android.content.Context
+
+/**
+ * No-op implementation.
+ */
+class RetentionManager @JvmOverloads constructor(
+        context: Context?,
```
olivierperez: Nullable context?

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Chucker methods and utilities to interact with the library.
+ */
+object Chucker {
+
+    /**
+     * Check if this instance is the operation one or no-op.
+     * @return `true` if this is the operation instance.
+     */
+    val isOp: Boolean get() = true
```
olivierperez: Directly set the value?

library/src/main/java/com/chuckerteam/chucker/api/RetentionManager.kt
```diff
@@ -0,0 +1,101 @@
...
+) {
+
+    // The actual retention period in milliseconds (default to
+    private val period: Long
+    // How often the cleanup should happen
+    private val cleanupFrequency: Long
+    private val prefs: SharedPreferences
+
+    init {
+        period = toMillis(retentionPeriod)
```
olivierperez: Those lines can be done in the property definition.

`private val period: Long = context.get......`

library-no-op/build.gradle
cortinico: Done ✅

PR #48, opened 2019-05-21
Make sure dependency examples have the proper case

cortinico:
Looks like having the dependency coordinates all lowercase is not working.
Updating to use the one copy-pasted from Jitpack.


cortinico:
> Does it change something?
Yup, the former was not always working 😅

PR #49, opened 2019-05-21
Add missing @JvmOverload on the ChuckerInterceptor

cortinico:
This was causing the library being harder to use from Java,
as you would need to provide all the parameters to initialise a
ChuckerInterceptor.


cortinico:
Yup we have it on the no-op. Was missing on the operation side

PR #51, opened 2019-05-24
Setup KtLint and Detekt

cortinico:
I set up KtLint and Detekt for the project in order to have a consistent styleguide across the project.
Here the summary of changes:

f8e1423 Fix KtLint Issues after the first Run
3abbc4d Fix Detekt Issues after the first Run
0dc44e3 Setup Detekt and KtLint for Chucker


cortinico:
It's correct as I'm using
```
buildUponDefaultConfig = true
```

Whatever is not specified is taken from the default detekt file

PR #53, opened 2019-06-04
Remove usage of methods from okhttp3.internal

### LINKED ISSUE #52: ChuckerInterceptor uses internal OkHttp APIs
This will break with OkHttp 4
See https://github.com/jgilfelt/chuck/issues/88

cortinico:
Chucker was using one methods from the aforementioned package.
Those methods are risky to use as they may be broken between releases
and I'm cleaning them up.

Specifically the method `hasBody` has now been moved to an extension
function inside our internal package. This will make easier to migrate
to OkHttp 4.x as the above method has been removed.

Fixes #52

Thanks @ZacSweers for the report <3

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -61,6 +61,7 @@ dependencies {
     testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
     testImplementation "org.junit.jupiter:junit-jupiter-params:$junitVersion"
+    testImplementation "org.mockito:mockito-core:$mockitoVersion"
 
     implementation "android.arch.persistence.room:runtime:$roomVersion"
     kapt "android.arch.persistence.room:compiler:$roomVersion"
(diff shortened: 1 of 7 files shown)


library/build.gradle
```diff
@@ -61,6 +61,7 @@ dependencies {
     testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
     testImplementation "org.junit.jupiter:junit-jupiter-params:$junitVersion"
+    testImplementation "org.mockito:mockito-core:$mockitoVersion"
```
olivierperez: There's another lib I use in Kotlin development to replace Mockito -> MockK.
I prefer its DSL to right Kotlin tests.
Try it and tell me what you think about.

BTW it won't block the PR

library/src/test/java/com/chuckerteam/chucker/api/internal/support/OkHttpUtilsTest.kt
```diff
@@ -0,0 +1,115 @@
...
+
+        assertEquals(42L, mockResponse.contentLenght)
+    }
+
+    @Test
+    fun isChunked_withNotChunked() {
+        val mockResponse = mock(Response::class.java)
+        whenever(mockResponse.header("Transfer-Encoding")).thenReturn("gzip")
+
+        assertEquals(false, mockResponse.isChunked)
```
olivierperez: You may use `assertFalse` instead (or `assertTrue` for the other tests).

library/build.gradle
cortinico: Yup let's give it a try :)

PR #62, opened 2019-07-04
LeakCanary memory leak report

uOOOO:
- AndroidX migration

- [LeakCanary](https://github.com/square/leakcanary) reports some memory leaks on Transaction*Fragment's adapter. I think the cause is that viewpager adapter holds child fragment's reference in arraylist. So I've changed some codes.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionActivity.java
+++ b/library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionActivity.java
@@ -53,7 +53,7 @@ public class TransactionActivity extends BaseChuckerActivity {
     }
 
     TextView title;
-    Adapter adapter;
+    PagerAdapter adapter;
 
     private long transactionId;
     private HttpTransaction transaction;
@@ -127,7 +127,7 @@ public class TransactionActivity extends BaseChuckerActivity {
     }
 
     private void setupViewPager(ViewPager viewPager) {
-        adapter = new Adapter(getApplicationContext(), getSupportFragmentManager());
+        adapter = new PagerAdapter(getApplicationContext(), getSupportFragmentManager());
         viewPager.setAdapter(adapter);
         viewPager.addOnPageChangeListener(new SimpleOnPageChangedListener() {
             @Override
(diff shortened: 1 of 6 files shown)


cortinico:
Hey @uOOOO 
Thanks for the PR. Can I ask you to split it in 2?

Ideally we would love to release Chucker 3.0 without the AndroidX migration so at least we have a working stable version with all the bugfixes we developed in the last months but without AndroidX.

I'll be happy to merge the LeakCanary fix into a separate PR.
Please let me know if you can do it or if I should cherry pick your CRs

uOOOO:
It's done! I've removed AndroidX commit. But in sample app, due to LeanCanary has AndroidX dependency since 2.x, I had to downgrade its version to 1.6.3 and SampleApplication is added for initializing it.

library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionActivity.java
```diff
@@ -148,32 +147,46 @@ private void share(String content) {
         startActivity(Intent.createChooser(sendIntent, null));
     }
 
-    static class Adapter extends FragmentPagerAdapter {
-        final List<TransactionFragment> fragments = new ArrayList<>();
-        private final List<String> fragmentTitles = new ArrayList<>();
+    static class Adapter extends FragmentStatePagerAdapter {
```
cortinico: Can I ask you to have this class as non static?
Moreover `PagerAdapter` is probably a better name

library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionActivity.java
```diff
@@ -148,32 +147,46 @@ private void share(String content) {
...
+        private final WeakReference<Context> context;
+        private static final int[] TITLE_RES_IDS = {
```
cortinico: Same here, this can be non-static

uOOOO:
I was waiting for travis build a few minutes ago, but it's already been merged. @cortinico Thank you.

PR #70, opened 2019-07-25
Add support for GraphQL

### LINKED ISSUE #69: Add support for GraphQL
I don’t know the details at first glance. Because every request is a POST method to `/graphql`

ex)

screenshot

I want to change `POST /graphql`  to used Query or Mutation name

**Describe alternatives you've considered**
Add processing to change the string in TransactionAdapter class when endpoint equals `/graphql` 
If there are other good ways, please suggest 🙏 

ref: https://graphql.org/learn/serving-over-http/#uris-routes

**Do you want to develop this feature yourself?**
yes

rnitame:
fix #69 

Screenshot_1567082567


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -70,6 +75,8 @@ class ChuckerInterceptor @JvmOverloads constructor(
             if (io.isPlaintext(buffer)) {
                 val content = io.readFromBuffer(buffer, charset, maxContentLength)
                 transaction.requestBody = content
+                transaction.operationName =
+                    Gson().fromJson(content, HttpRequestBody::class.java).operationName
```
cortinico: This will be really inefficient :(
You're going to recreate a `Gson` instance for every request.
Please use a `JsonConvertor.getInstance()`.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -164,4 +172,8 @@ class ChuckerInterceptor @JvmOverloads constructor(
     companion object {
         private val UTF8 = Charset.forName("UTF-8")
     }
+
+    data class HttpRequestBody(
+        val operationName: String?
```
cortinico: GSON annotation is missing here.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -164,4 +172,8 @@ class ChuckerInterceptor @JvmOverloads constructor(
     companion object {
         private val UTF8 = Charset.forName("UTF-8")
     }
+
+    data class HttpRequestBody(
```
cortinico: Naming can be improved here:

```suggestion
    data class GraphQLRequestBody(
```

library/src/main/java/com/chuckerteam/chucker/api/internal/data/entity/HttpTransaction.kt
```diff
@@ -118,13 +119,16 @@ internal class HttpTransaction(
             return when (status) {
                 Status.Failed -> " ! ! !  $method $path"
                 Status.Requested -> " . . .  $method $path"
-                else -> responseCode.toString() + " " + method + " " + path
+                else -> responseCode.toString() + " " + method + " " + path + if (isGraphql) operationName else ""
```
cortinico: Can you add a space between path and operation name?

rnitame: done!

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
rnitame: Since the location of operation name was different depending on the library, it was absorbed using `JsonParser`

[apollo-android](https://github.com/apollographql/apollo-android/)  : 

```json
{
  "operationName": "xxx"
}
```

[retrofit-graphql](https://github.com/AniTrend/retrofit-graphql) :

```json
{
  "queryContainer" : {
       "operationName": "xxx"
   }
}
```

rnitame:
@cortinico Thanks for your comments!

> Would you be able to provide a sample public endpoint to include inside the Sample app?

I added implementation using SWApi. http://graphql.org/swapi-graphql/ :)
Please review again.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -161,6 +170,17 @@ class ChuckerInterceptor @JvmOverloads constructor(
         return response.body()!!.source()
     }
 
+    private fun findOperationName(content: JsonObject): String? {
+        val key = "operationName"
+        if (content.has(key)) return content.get(key).asString
+
+        return if (content.has("queryContainer")) {
```
cortinico: Hey @rnitame 
Apparently here there is a problem. GraphQL request should have the `operationName` at the top level and should not be wrapped inside a `queryContainer` object:

Screenshot_1565099160

This is an invalid GraphQL request.

The problem is caused by the initialisation of the retrofit interface:
```
        Retrofit retrofit = new Retrofit.Builder()
                .baseUrl("https://graphql.org")
                .addConverterFactory(GsonConverterFactory.create())
                .client(client)
                .build();
```
Although we could fix it by using the Converter factory provided by retrofit-graphql, I think it's better if we change approach here.

Can I ask you to:

a) Don't use `retrofit-graphql`. Is yet another dependency that we probably don't need.
b) Just create a simple `data class` and do a GraphQL request without external library (as described here: https://graphql.org/learn/serving-over-http/#post-request)
c) Remove all the .graphql file?

cortinico:
Let me know when this is ready for review again :)

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -71,7 +77,13 @@ class ChuckerInterceptor @JvmOverloads constructor(
             }
             if (io.isPlaintext(buffer)) {
                 val content = io.readFromBuffer(buffer, charset, maxContentLength)
+                val graphQLRequestBody = JsonConverter.instance.fromJson<GraphQLRequestBody>(
+                    content,
+                    object : TypeToken<GraphQLRequestBody>() {
+                    }.type
```
olivierperez: We can use this trick 

```kotlin
// Somewhere smart in the lib
inline fun <reified T> genericType() = object: TypeToken<T>() {}.type
```

```kotlin
// Usage
genericType<GraphQLRequestBody>()
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionAdapter.java
```diff
@@ -86,26 +86,33 @@ public void setData(List<HttpTransactionTuple> httpTransactions) {
...
             size = view.findViewById(R.id.chucker_size);
+            graphql = view.findViewById(R.id.chucker_graphql);
             ssl = view.findViewById(R.id.chucker_ssl);
         }
 
         void bind(final HttpTransactionTuple transaction) {
             path.setText(String.format("%s %s", transaction.getMethod(), transaction.getPath()));
+            operationName.setText(transaction.getOperationName());
+            operationName.setVisibility(transaction.isGraphQL() ? View.VISIBLE : View.GONE);
+            graphql.setVisibility(transaction.isGraphQL() ? View.VISIBLE : View.GONE);
```
olivierperez: All those lines are about GraphQL, can do something to group them ?

```kotlin
if (transaction.isGraphQL()) {
    operationName.setText(transaction.getOperationName());
    operationName.setVisibility(View.VISIBLE);
    graphql.setVisibility(View.VISIBLE);
}
```

And set them to View.GONE by default.

sample/build.gradle
```diff
@@ -36,6 +36,10 @@ android {
     }
 }
 
+repositories {
+    maven { url 'https://jitpack.io' }
+}
+
```
olivierperez: Can you add the repository in `/build.gradle` instead? Under `allprojects.repositories`.

sample/src/main/java/com/chuckerteam/chucker/sample/MainActivity.kt
```diff
@@ -97,6 +99,19 @@ class MainActivity : AppCompatActivity() {
             cache("Mon").enqueue(cb)
             cache(30).enqueue(cb)
         }
+
+        with(SampleApiService.getGraphQLInstance(getClient(this))) {
```
olivierperez: The method `getClient` doesn't only get the HTTP client, it also create it. It should be named `buildHttpClient` or something, it would explicitly show that something is built by this method. And maybe we shouldn't abuse of calls to this method.

sample/build.gradle
olivierperez: Actually, why do we need this repository? Can we just remove it?

cortinico: Correct, this can be removed entirely

sample/proguard-rules.pro
```diff
@@ -17,3 +17,6 @@
 #}
 -dontwarn okio.**
 -dontwarn retrofit2.**
+
+# retrofit-graphql
```
cortinico: Please remove this

sample/src/main/AndroidManifest.xml
```diff
@@ -20,6 +20,8 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
 
+    <uses-sdk tools:overrideLibrary="io.github.wax911.library" />
```
cortinico: Please remove this

library/src/main/res/drawable/chucker_ic_graphql_24dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
+      android:fillColor="#e535ab"
```
cortinico: Please export this color

sample/src/main/java/com/chuckerteam/chucker/sample/SampleApiService.java
```diff
@@ -50,53 +61,103 @@ static HttpbinApi getInstance(OkHttpClient client) {
...
         Call<Void> deny();
+
         @GET("/cache")
         Call<Void> cache(@Header("If-Modified-Since") String ifModifiedSince);
+
         @GET("/cache/{seconds}")
         Call<Void> cache(@Path("seconds") int seconds);
     }
+
+    interface GraphQLSWApi {
```
cortinico: Can you move this to a separate class, like a `SimpleGraphQLService` or something similar

vbuberen:
@rnitame 
Could you give some timeline on when you plan to finish this PR? 

I would love to proceed with `3.1.0` release in some nearest time, so if it would take too much time we might need to postpone it to `3.2.0`.

Agree @cortinico @olivierperez ?

cortinico:
> I would love to proceed with `3.1.0` release in some nearest time, so if it would take too much time we might need to postpone it to `3.2.0`.
> 
> Agree @cortinico @olivierperez ?

Definitely, let's move this to `3.2.0` 👍

vbuberen:
@rnitame @cortinico
Hey guys, could someone provide some info on the status of this feature and what is required to make it to production?
Is any help or discussion required?

vbuberen:
Also, let's consider points from #116 here.
Probably, it is better to outline some plan with help of @psh to improve implementation as much as we can?

cortinico:
> Probably, it is better to outline some plan with help of @psh to improve implementation as much as we can?

Agree here. I generally like the approach in this PR by @rnitame. I believe we could reuse the UI part at least.

On the other hand I believe GraphQL initialisation should be provided by the developer when creating the `ChuckerInterceptor`, rather than having the library trying to _infer_ if a request is actually a valid GraphQL request. So on that side we should probably change our approach.

quentin41500:
This is such a nice feature! Anyway I can help getting that merge? I'm not the branch owner so I don't think I can push to it, but I could make a new branch if that's ok.

Let me know!

cortinico:
> This is such a nice feature! Anyway I can help getting that merge? I'm not the branch owner so I don't think I can push to it, but I could make a new branch with the desired changes if that's ok.
> 
> Let me know!

You can definitely send a new PR branching from this one :)

vbuberen:
@quentin41500 That is actually cool. I believe that you could cooperate with @rnitame to do a new version of PR. 

Also, I would like to join you in these efforts, since from what I see there were not much attention to this PR and it is quite old already, while the feature is really useful. So ping me here or in [Chucker Slack channel ](https://kotlinlang.slack.com/archives/CRWD6370R) and let's make it happen in `3.2.0` 😄

PR #72, opened 2019-08-02
Fetch payload in AsyncTask

### LINKED ISSUE #55: Chucker is crashing opening responses with big bodies
**Describe the bug**

I do have an app that is calling an endpoint that returns a JSON Body of nearly 700kb.
Chucker is blocking the UI thread trying to parse the body and is causing an ANR/Crash if trying to inspect the response body.

**To Reproduce**
1. Use Chucker to call an endpoint with a big response
2. Click on the Chucker Push notification
3. Open the HTTP Request from the list
4. Click on the "Response" tab

**Expected behavior**
The tab should open immediately

**Actual behavior**
The Chucker UI is completely frozen and eventually crashes

**Smartphone**
 - Emulator - Nexus 5X
 - OS: API27
 - Lib version: 3.0.0-alpha1

redwarp:
To avoid the formating of json or xml stalling the UI, the payload will be loaded and formatted in an async task instead.
Fixes #55

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
@@ -155,11 +155,18 @@ internal class TransactionPayloadFragment : Fragment(), TransactionFragment, Sea
         }
 
         override fun onPostExecute(result: UiPayload) {
-            fragment.setBody(result.headersString, result.bodyString, result.isPlainText, result.image)
+            with(result) {
+                fragment.setBody(headersString, bodyString, isPlainText, image)
+            }
         }
     }
 
-    private data class UiPayload(val headersString: String, val bodyString: String?, val isPlainText: Boolean, val image: Bitmap?)
+    private data class UiPayload(
+        val headersString: String,
+        val bodyString: String?,
+        val isPlainText: Boolean,
+        val image: Bitmap?
+    )
 
     companion object {
 
--- a/library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
@@ -15,6 +15,7 @@
  */
 package com.chuckerteam.chucker.api.internal.ui.transaction
 
+import android.graphics.Bitmap
 import android.graphics.BitmapFactory
 import android.os.AsyncTask
 import android.os.Bundle
@@ -28,7 +29,6 @@ import android.view.MenuInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
-import android.view.*
 import android.widget.TextView
 import com.chuckerteam.chucker.R
 import com.chuckerteam.chucker.api.internal.data.entity.HttpTransaction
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -121,6 +117,31 @@ internal class TransactionPayloadFragment : Fragment(), TransactionFragment, Sea
...
+                )
+                else -> Triple(
+                    transaction.getResponseHeadersString(true),
+                    transaction.getFormattedResponseBody(),
+                    transaction.isResponseBodyPlainText
+                )
+            }
+        }
+
+        override fun onPostExecute(result: Triple<String, String, Boolean>) {
```
cortinico: How about having a progress bar and hiding it in the `onPreExecute`?

cortinico:
You can use this Charles rewrite rule:
[BigGetResponse.xml.zip](https://github.com/ChuckerTeam/chucker/files/3473049/BigGetResponse.xml.zip)

Plus you need to remove SSL to let it work properly here:
https://github.com/ChuckerTeam/chucker/blob/1e893c2a6bfa16e3d41bd9488f84a698d060e13e/sample/src/main/java/com/chuckerteam/chucker/sample/SampleApiService.java#L35-L36

redwarp:
Verified with a big query that it actually solves the issue.

library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
redwarp: I rebased. I think a progress bar could be nice, but can be done separately.

library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -129,19 +129,45 @@ internal class TransactionPayloadFragment : Fragment(), TransactionFragment, Sea
...
+
+            val bitmap = transaction.responseImageData?.let { imageData ->
+                BitmapFactory.decodeByteArray(imageData, 0, imageData.size)
+            }
+
+            return when (type) {
+                TYPE_REQUEST -> UiPayload(
+                    transaction.getRequestHeadersString(true),
+                    transaction.getFormattedRequestBody(),
+                    transaction.isRequestBodyPlainText, bitmap
```
cortinico: Can you put it on 2 lines?
```suggestion
                    transaction.isRequestBodyPlainText, 
                    bitmap
```

library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -129,19 +129,45 @@ internal class TransactionPayloadFragment : Fragment(), TransactionFragment, Sea
...
+                )
+                else -> UiPayload(
+                    transaction.getResponseHeadersString(true),
+                    transaction.getFormattedResponseBody(),
+                    transaction.isResponseBodyPlainText, bitmap
```
cortinico: Can you put it on 2 lines?
```suggestion
                    transaction.isResponseBodyPlainText, 
                    bitmap
```

library/src/main/java/com/chuckerteam/chucker/api/internal/ui/transaction/TransactionPayloadFragment.kt
redwarp: Damn. Sorry, our work project's setup make that happen automatically when formatting, I though the same would happen here :-P

PR #85, opened 2019-08-09
Fix race conditions in NotificationHelper.

koral--:
Access to `transactionBuffer` is not properly synchronized which may cause NPEs.
https://github.com/jgilfelt/chuck/issues/69

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/internal/support/NotificationHelper.java
+++ b/library/src/main/java/com/chuckerteam/chucker/api/internal/support/NotificationHelper.java
@@ -49,22 +49,26 @@ public class NotificationHelper {
     private final Context context;
     private final NotificationManager notificationManager;
 
-    public static synchronized void clearBuffer() {
-        transactionBuffer.clear();
-        transactionIdsSet.clear();
+    public static void clearBuffer() {
+        synchronized (transactionBuffer) {
+            transactionBuffer.clear();
+            transactionIdsSet.clear();
+        }
     }
 
-    private static synchronized void addToBuffer(HttpTransaction transaction) {
+    private static void addToBuffer(HttpTransaction transaction) {
         if (transaction.getId() == 0) {
             // Don't store Transactions with an invalid ID (0).
             // Transaction with an Invalid ID will be shown twice in the notification
             // with both the invalid and the valid ID and we want to avoid this.
             return;
         }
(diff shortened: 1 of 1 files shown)


cortinico:
Were you able to reproduce? or is this a best guess fix?

koral--:
I think neither of that cases. It's rather effect of code analysis.
I've got this crash report from my app:
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.readystatesoftware.chuck.internal.data.HttpTransaction.getNotificationText()' on a null object reference
       at com.readystatesoftware.chuck.internal.support.NotificationHelper.show + 96(NotificationHelper.java:96)
```
Line 96 looks like that:
`inboxStyle.addLine(transactionBuffer.valueAt(i).getNotificationText());`
So `transactionBuffer.valueAt(i)` must have returned `null`. This may happen if value at key `i` has been removed.
Code which can remove elements is located in line 50, in `clearBuffer` method.

Look at the methods signatures:
`public synchronized void show(HttpTransaction transaction)`
`public static synchronized void clearBuffer()`

`show` is non-static while `clearBuffer` is static. They are both synchronized but on different objects (instance vs class). So synchronization is invalid and those methods may be executed by different threads at the same time.

library/src/main/res/values/strings.xml
```diff
@@ -46,7 +46,7 @@
     <string name="chucker_tab_errors">Errors</string>
     <string name="chucker_notification_category">Chucker notifications</string>
     <string name="chucker_share_error_title">Error report</string>
-    <string name="chucker_share_error_content"><![CDATA[Date: %1$s\nException: %2$s\nTag: %3$s\nMessage: %4$s\n\n%5$s]]></string>
+    <string name="chucker_share_error_content"><![CDATA[Date: %s\nException: %s\nTag: %s\nMessage: %s\n\n%s]]></string>
```
cortinico: Apparently this should not be in the diff.
Can you please remove it?

koral--: Sure. Done.

wwwss:
Which version will have this code

cortinico:
> Which version will have this code

Will be in 3.1.0 that is going to be released soon (still no ETA sorry).
If you need, you can use the SNAPSHOT version from the `develop` branch that contains this code.

PR #96, opened 2019-08-16
Limit size of binary image to 1 million bytes.

### LINKED ISSUE #94: Image that are too big will make chucker crash
**Describe the bug**
Because the image data is stored in database, if an image is too big, it crashes

**To Reproduce**
1. Use Chucker to download a really big image, like https://i.ibb.co/znbM2dp/IMG-20190812-194359.jpg

**Expected behavior**
Chucker should work

redwarp:
It prevents the SQLiteBlobTooBigException that would happen when downloading big images.

Fixes #94

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -16,6 +16,8 @@ import okhttp3.Response
 import okio.Buffer
 import okio.BufferedSource
 
+private const val MAX_BLOB_SIZE = 1000_000L
+
 /**
  * An OkHttp Interceptor which persists and displays HTTP activity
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -126,7 +126,7 @@ class ChuckerInterceptor @JvmOverloads constructor(
             } else {
                 transaction.isResponseBodyPlainText = false
 
-                if (transaction.responseContentType?.contains("image") == true) {
+                if (transaction.responseContentType?.contains("image") == true && buffer.size() < 1000_000L) {
```
cortinico: Can we export `1000_000L` to a constant?

redwarp:
Extracted 1000_000L to a constant.

PR #101, opened 2019-08-22
Move the .internal package outside of .api

cortinico:
The public API of Chucker is currently inside the `com.chuckerteam.chucker.api`
package (those are the classes were we ensure strict semantic
versioning).

Classes inside the `com.chuckerteam.chucker.api.internal` package are
not part of the public API. For historical reason the .internal package
is a subpackage of the .api package. This makes things really confusing.

I'm now moving the internal package outside of the api package to:
`com.chuckerteam.chucker.internal`.


redwarp:
Good idea!

cortinico:
> Good idea!

Are we merging it @redwarp ? :)

PR #111, opened 2019-09-08
Removing SimpleOnPageChangedListener

cortinico:
We probably don't need this interface as it's just used in one place.
The only affected class is `TransactionActivity` that will be
simplified even more once we migrate it to Kotlin as well.


redwarp:
I think the idea was to mimick this class: https://developer.android.com/reference/android/support/v4/view/ViewPager.SimpleOnPageChangeListener.html

But it's not super useful indeed.

PR #123, opened 2019-09-26
Kotlinify classes in .internal.ui.error package

cortinico:


library/src/main/java/com/chuckerteam/chucker/internal/ui/MainActivity.java
```diff
@@ -92,7 +92,8 @@ private CharSequence getApplicationName() {
 
     @Override
     public void onErrorClick(long throwableId, int position) {
-        ErrorActivity.start(this, throwableId);
+        Intent intent = ErrorActivity.newInstance(this, throwableId);
```
redwarp: You can probably inline that.

cortinico: done

PR #129, opened 2019-10-05
Websockets Support

psh:
tl;dr - this PR adds support for Chucker to monitor _websocket_ traffic.

image

If I wanted to write a _ticker_ app that displayed the current price of BitCoin, I might write
```kotlin
    private var coinbaseSocket: WebSocket? = null

    fun start() {
        coinbaseSocket = coinbaseSocket().also {
            // Subscribe to get Bitcoin ticker updates
            it.send("{ \"type\": \"subscribe\", \"product_ids\": [\"BTC-USD\"], \"channels\": [\"ticker\"] }")
        }
    }

    fun stop() {
        if (coinbaseSocket != null) {
            coinbaseSocket?.close(1000, "Thanks!")
            coinbaseSocket = null
        }
    }

    private fun coinbaseSocket(): WebSocket {
        return httpClient.newWebSocket(
            Request.Builder().url("wss://ws-feed.pro.coinbase.com").build(),
            object : WebSocketListener() { /* Display prices */ }
        )
    }
```
Sadly OkHttp bypasses the normal `Interceptor` framework when running websockets; Chucker would display a single `GET` request as OkHttp negotiated switching from HTTP over to the Websockets protocol, and then would be missing all of the lifecycle and asynchronous messages that flow over the websocket.

This PR introduces a new API that injects monitoring by wrapping the underlying websocket and supplied websocket listener.  Client code would need to switch from calling `httpClient.newWebSocket()` to `httpClient.newLoggedWebSocket()`
```kotlin
    private fun coinbaseSocket(): WebSocket {
        return httpClient.newLoggedWebSocket(
            Request.Builder().url("wss://ws-feed.pro.coinbase.com").build(),
            object : WebSocketListener() {}
        )
    }
```

Also this PR re-writes the HTTP traffic display to handle 
* HTTP transaction rows
* Websocket lifecycle rows (nothing interesting to view, so no detail screen if tapped)
* Websocket data rows (showing what was sent / received) 

Along the way tests were added and several Java files were converted to Kotlin.


redwarp:
Hey there, I think this change is doing waay too much at once, between kotlinifying stuff, bumping libraries, supporting websockets...

Can you maybe break that down in several pull requests?

Can you also create a ticket for the support of web socket, so we can have a discussion on there?

Thanks, I'm looking forward to discuss about it and having support for web-socket in Chucker!

psh:
Thanks for the feedback - I'll close this PR for now.

PR #133, opened 2019-10-07
Updating Detekt to 1.1.0

cortinico:


cortinico:
> You removed the when, is it because of Detekt?

Sorry I forgot to answer you, but yes was because of a newly enabled detekt rule.

PR #122, opened 2019-09-25
Enhancement - Update NotificationHelper to kotlin

ullas-jain:
A small step to convert project into kotlin.

`This is inspired by - @christopherniksch ` [PR](https://github.com/ChuckerTeam/chucker/pull/120)

PR #120, opened 2019-09-19
Kotlinify FormatUtils

christopherniksch:
This PR is for the general goal of converting the project over to Kotlin

PR #135, opened 2019-10-11
Kotlinify classes in .internal.ui package

yoavst:
As other prs like #120 and #122 , this pr aims to convert the Java classes of this project to kotlin.


library/src/main/java/com/chuckerteam/chucker/internal/ui/HomePageAdapter.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentStatePagerAdapter
+import com.chuckerteam.chucker.R
+import com.chuckerteam.chucker.internal.ui.error.ErrorListFragment
+import com.chuckerteam.chucker.internal.ui.transaction.TransactionListFragment
+import java.lang.ref.WeakReference
+
+/**
+ * @author Olivier Perez
+ */
```
redwarp: You can remove this comment while you are at it, we are not really adding authorship like that anymore.

Issue #82, opened 2019-08-09
Large json responses result in bad user experience


ColtonIdle:
Some of the backends I work with send really large amounts of data.

Here is a sample: https://data.cityofnewyork.us/api/views/kku6-nxdu/rows.json

Moving from the request to response tab is slow, and trying to "Share as text" is also impossible. It'd be great to keep this on the roadmap for improvement.

Edit: Occurs in latest stable and beta

Edit 2: Maybe just show the first x amount of lines, but still provide a way to export or view the json as a file?

cortinico:
Part of this issue was addressed by @redwarp in #72 

> Edit 2: Maybe just show the first x amount of lines, but still provide a way to export or view the json as a file?

You can set the `maxContentLength` on the `ChuckerInterceptor` to truncate the body if needed (but still you won't be able to export or share the full body afterwards).

Curious to hear what others think about this

ColtonIdle:
Gotcha. I think after a certain size maybe 500k or something the text just gets treated as a file. Then in the response tab it can say "Response exceed x" with a button that opens the file in another file viewer. Also, the share as text would turn into a "Share as file" button?

cortinico:
Sounds more like as a feature-request.
Definitely valuable to implement :)

redwarp:
Actually, being able to open any response as a file with another viewer sounds like a great idea!

kmayoral:
Just tried out this library and it looks great, but the one thing keeping me from migrating over from Chuck is this issue where large json responses cause Chucker to hang and then eventually crash after you try and swipe from the Overview tab to either the Request or Response tab (probably when the data starts to get loaded into the Response tab).

Chuck is able to show these responses (even if there is no ability to search) so for now, I'll have to stay with Chuck. 

Looking forward to trying this library again in the future, it looks great!

ColtonIdle:
Just bumping this. Tried using Chucker on a new project and it's completely useless because of large json. It just ANRs. 😭

cortinico:
> Just bumping this. Tried using Chucker on a new project and it's completely useless because of large json. It just ANRs.

Thanks for bumping. Is this still happening? If yes, which version are you using? `3.0.1`?

ColtonIdle:
@cortinico yes this is still happening. I am using 3.0.1

cortinico:
[...]

PR #138, opened 2019-10-12
Add an option to save the request or response body

yoavst:
## :camera: Screeshots
Screenshot

## :page_facing_up: Context
On issue #82, it says:
> Actually, being able to open any response as a file with another viewer sounds like a great idea!

However, There were few problems implementing that:
Android have no standard way to use `ACTION_VIEW` intent with binary data from an app. One has to save it to file, but on android 7.0 it doesn't work anymore, and so on...
Therefore, The simplest idea was to provide an easy way to save the response to a file. I assumes this is good enough as a starter.

In addition, this implementation only works on 4.4+, since it uses SAF framework to get a `Uri` to a file. On older version there is no standard dialog for that.

**Also, and probably most importantly, I'm not sure the implementation is correct, so instead of thinking on it alone, I'm sending this PR so everybody could comment on it**

## :pencil: Changes
I've added a menu item to transaction activity, and added a handler for it on `TransactionPayloadFragment`. The handler launches an `ACTION_CREATE_DOCUMENT` intent, and on receiving `Uri`, launch an `AsyncTask` to save the data to the file.

## :pencil: How to test
Maybe i should extract the IO saving functionality to a method, so we can test it using a unit test.
Otherwise, I think integration test is the way to check it.
Also, I'm sure it doesn't works with multi-part requests

## :crystal_ball: Next steps before merge
- [x] Confirm the implementation is correct
- [ ] Add support for 4.2, 4.3
- [ ] Add support for multipart requests
- [ ] Add tests


cortinico:
Thanks for the PR @yoavst <3

Just kicking off the discussion here. (Also I'd love to get @redwarp input as he initially mentioned the idea of saving to file).

> In addition, this implementation only works on 4.4+, since it uses SAF framework 
> to get a Uri to a file. On older version there is no standard dialog for that.

How about using the request timestamp to save a file inside the external storage (e.g. `/chucker/<request_timestamp>`) ? I also feel like that having to pick a file for the user is a bit more annoying. You anyway have to pick the file again once you want to share it.

We should probably show a loading dialog or something similar to the user while saving the file.

> Maybe i should extract the IO saving functionality to a method, so we can test 
> it using a unit test.
> Otherwise, I think integration test is the way to check it.

Sounds great to me :+1: 

> Confirm the implementation is correct

Is there something that makes you think this implementation is not correct?

yoavst:
> How about using the request timestamp to save a file inside the external storage
The problem is that it requires storage permission which I tried to avoid.

cortinico:
Looks like you have a problem on your branch @yoavst.
I currently experience a crash every time I open any transaction:

<details>
<summary>Stacktrace</summary>
<code>
2019-10-28 11:19:17.160 5121-5121/com.chuckerteam.chucker.sample E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.chuckerteam.chucker.sample, PID: 5121
    java.lang.IllegalStateException: findViewById(R.id.image) must not be null
        at com.chuckerteam.chucker.internal.ui.transaction.TransactionPayloadFragment.onCreateView(TransactionPayloadFragment.kt:66)
        at androidx.fragment.app.Fragment.performCreateView(Fragment.java:2600)
        at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:881)
        at androidx.fragment.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManagerImpl.java:1238)
        at androidx.fragment.app.BackStackRecord.executeOps(BackStackRecord.java:434)
        at androidx.fragment.app.FragmentManagerImpl.executeOps(FragmentManagerImpl.java:2079)
        at androidx.fragment.app.FragmentManagerImpl.executeOpsTogether(FragmentManagerImpl.java:1869)
        at androidx.fragment.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManagerImpl.java:1824)
        at androidx.fragment.app.FragmentManagerImpl.execSingleAction(FragmentManagerImpl.java:1696)
        at androidx.fragment.app.BackStackRecord.commitNowAllowingStateLoss(BackStackRecord.java:299)
        at androidx.fragment.app.FragmentStatePagerAdapter.finishUpdate(FragmentStatePagerAdapter.java:259)
        at androidx.viewpager.widget.ViewPager.populate(ViewPager.java:1244)
        at androidx.viewpager.widget.ViewPager.populate(ViewPager.java:1092)
        at androidx.viewpager.widget.ViewPager.onMeasure(ViewPager.java:1622)
        at android.view.View.measure(View.java:22071)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6602)
        at androidx.coordinatorlayout.widget.CoordinatorLayout.onMeasureChild(CoordinatorLayout.java:743)
        at com.google.android.material.appbar.HeaderScrollingViewBehavior.onMeasureChild(HeaderScrollingViewBehavior.java:99)
        at com.google.android.material.appbar.AppBarLayout$ScrollingViewBehavior.onMeasureChild(AppBarLayout.java:1882)
        at androidx.coordinatorlayout.widget.CoordinatorLayout.onMeasure(CoordinatorLayout.java:813)
        at android.view.View.measure(View.java:22071)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6602)
        at android.widget.FrameLayout.onMeasure(FrameLayout.java:185)
        at androidx.appcompat.widget.ContentFrameLayout.onMeasure(ContentFrameLayout.java:143)
        at android.view.View.measure(View.java:22071)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6602)
        at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1514)
        at android.widget.LinearLayout.measureVertical(LinearLayout.java:806)
        at android.widget.LinearLayout.onMeasure(LinearLayout.java:685)
        at android.view.View.measure(View.java:22071)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6602)
        at android.widget.FrameLayout.onMeasure(FrameLayout.java:185)
        at android.view.View.measure(View.java:22071)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6602)
        at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1514)
        at android.widget.LinearLayout.measureVertical(LinearLayout.java:806)
        at android.widget.LinearLayout.onMeasure(LinearLayout.java:685)
        at android.view.View.measure(View.java:22071)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6602)
        at android.widget.FrameLayout.onMeasure(FrameLayout.java:185)
        at com.android.internal.policy.DecorView.onMeasure(DecorView.java:724)
        at android.view.View.measure(View.java:22071)
        at android.view.ViewRootImpl.performMeasure(ViewRootImpl.java:2422)
        at android.view.ViewRootImpl.measureHierarchy(ViewRootImpl.java:1504)
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:1761)
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1392)
        at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:6752)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:911)
        at android.view.Choreographer.doCallbacks(Choreographer.java:723)
        at android.view.Choreographer.doFrame(Choreographer.java:658)
2019-10-28 11:19:17.161 5121-5121/com.chuckerteam.chucker.sample E/AndroidRuntime:     at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:897)
        at android.os.Handler.handleCallback(Handler.java:790)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6494)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
</code>
</details>

Looks like you're inflating the wrong View ID here:
https://github.com/ChuckerTeam/chucker/pull/138/files#diff-82fdd209bbae8c34145cb5e71831d8efR74

cortinico:
Ping @yoavst What's the status? Please let me know if me, or someone else, should take this PR over :) I'm happy to support

yoavst:
@cortinico , unfortunately I don't have a computer in reach for the next month, so I'd be happy if you would take over this PR.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -78,15 +88,27 @@ internal class TransactionPayloadFragment :
...
+                searchMenuItem.isVisible = true
+                val searchView = searchMenuItem.actionView as SearchView
+                searchView.setOnQueryTextListener(this)
+                searchView.setIconifiedByDefault(true)
+            }
+
+            val saveMenuItem = menu.findItem(R.id.save_body)
+
+            // SAF is not available on pre-Kit Kat so let's hide the icon.
+            saveMenuItem.isVisible = Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT
```
vbuberen: I would suggest adding braces for logical conditions in such cases to make it easier to read.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -105,11 +127,12 @@ internal class TransactionPayloadFragment :
...
     private fun populateUI() {
         if (isAdded && transaction != null) {
-            UiLoaderTask(this).execute(Pair(type, transaction!!))
+            uiLoaderTask = UiLoaderTask(this).apply { execute(Pair(type, transaction!!)) }
         }
     }
 
     private fun setBody(headersString: String, bodyString: String?, isPlainText: Boolean, image: Bitmap?) {
+        uiLoaderTask = null
         headers.visibility = if (headersString.isEmpty()) View.GONE else View.VISIBLE
```
vbuberen: I see `isVisible()` from KTX being used earlier in this file. Suggest to replace this line with `isVisible()` too for clarity.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -164,6 +230,47 @@ internal class TransactionPayloadFragment :
...
+                return false
+            }
+            return true
+        }
+
+        override fun onPostExecute(result: Boolean) {
+            fragment.fileSaverTask = null
+            fragment.context?.let {
+                if (result) {
+                    Toast.makeText(it, R.string.chucker_file_saved, Toast.LENGTH_SHORT).show()
```
vbuberen: What about idea of defining only the resource name in this conditional check and write `Toast` creation only once?
Something like:
```
val toastResource = if (result) {
    R.string.chucker_file_saved
} else {
    R.string.chucker_file_not_saved
}
Toast.makeText(it, toastResource, Toast.LENGTH_SHORT).show()
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -164,6 +230,47 @@ internal class TransactionPayloadFragment :
...
+                e.printStackTrace()
+                return false
+            } catch (e: IOException) {
+                e.printStackTrace()
+                return false
+            }
+            return true
+        }
+
+        override fun onPostExecute(result: Boolean) {
```
vbuberen: I suggest to rename the variable to match its type.

library/src/main/res/values/strings.xml
```diff
@@ -38,6 +38,13 @@
     <string name="chucker_share">Share</string>
     <string name="chucker_share_as_text">Share as text</string>
     <string name="chucker_share_as_curl">Share as curl command</string>
+    <string name="chucker_save">Save body to file</string>
+    <string name="chucker_empty_request_body">The request\'s body is empty</string>
+    <string name="chucker_empty_response_body">The response\'s body is empty</string>
+    <string name="chucker_save_is_only_supported_on_19">This feature is currently available only for Android 4.4+</string>
```
vbuberen: I guess, this is redundant now, because users won't see an option on devices with older versions of Android.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -164,6 +230,47 @@ internal class TransactionPayloadFragment :
...
+        AsyncTask<Triple<Int, Uri, HttpTransaction>, Unit, Boolean>() {
+
+        @Suppress("NestedBlockDepth")
+        override fun doInBackground(vararg params: Triple<Int, Uri, HttpTransaction>): Boolean {
+            val (type, uri, transaction) = params[0]
+            try {
+                val context = fragment.context ?: return false
+                context.contentResolver.openFileDescriptor(uri, "w")?.use {
+                    FileOutputStream(it.fileDescriptor).use { fos ->
+                        if (type == TYPE_REQUEST) {
```
vbuberen: I suggest to replace with `when`

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
cortinico: I'm not sure I got it 🤔

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
cortinico: Is actually `isVisible()` from `MenuItem` and not from KTX so we can't use that method.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
vbuberen: In this particular case we have Task for saving the result into file, so we could name the parameter as `isSaved` or `wasSaved` or `isSuccessful`. After such change the condition below would look like:
```
if (isSaved) {
....
}
```
which looks quite more readable.

cortinico: Got it :) Yup I'm changing it

vbuberen:
Just tested this thing - looks good for me in terms of functionality.
In terms of code - will leave another comment, which more a nitpick than an issue.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -83,15 +96,31 @@ internal class TransactionPayloadFragment :
...
+        if ((type == TYPE_RESPONSE || type == TYPE_REQUEST)) {
+            if (body.text.isNotEmpty()) {
+                val searchMenuItem = menu.findItem(R.id.search)
+                searchMenuItem.isVisible = true
+                val searchView = searchMenuItem.actionView as SearchView
+                searchView.setOnQueryTextListener(this)
+                searchView.setIconifiedByDefault(true)
+            }
+
+            val saveMenuItem = menu.findItem(R.id.save_body)
```
vbuberen: What if we follow the same principle of not showing an icon in case of empty body, like it is done with search feature? If we do so, we can also remove complex conditions from `viewBodyExternally`, because there will be no need for toasts with errors.

cortinico: Agree, I've added it. Dont' you mind taking a look?

vbuberen: Sure

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -83,15 +96,39 @@ internal class TransactionPayloadFragment :
...
+            if (body.text.isNotEmpty()) {
+                val searchMenuItem = menu.findItem(R.id.search)
+                searchMenuItem.isVisible = true
+                val searchView = searchMenuItem.actionView as SearchView
+                searchView.setOnQueryTextListener(this)
+                searchView.setIconifiedByDefault(true)
+            }
+
+            val transaction = viewModel.transaction.value
+            val saveMenuItem = menu.findItem(R.id.save_body)
```
vbuberen: Let's move `findItem` under `if (showSaveMenuItem)...`, because we might not need this menu item at all.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -83,15 +96,39 @@ internal class TransactionPayloadFragment :
...
+            val showSaveMenuItem = when {
+                // SAF is not available on pre-Kit Kat so let's hide the icon.
+                Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT -> false
```
vbuberen: Let's add braces for the condition here, since it is quite hard to read such things fast.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -83,15 +96,39 @@ internal class TransactionPayloadFragment :
...
+                else -> true
+            }
+
+            if (showSaveMenuItem) {
+                saveMenuItem.isVisible = true
+                saveMenuItem.setOnMenuItemClickListener {
+                    viewBodyExternally()
+                    true
+                }
+            } else {
```
vbuberen: Maybe we could set `visible` property in menu xml and remove `else` part here?

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -122,6 +160,38 @@ internal class TransactionPayloadFragment :
...
+        }
+    }
+
+    override fun onActivityResult(requestCode: Int, resultCode: Int, resultData: Intent?) {
+        if (requestCode == GET_FILE_FOR_SAVING_REQUEST_CODE && resultCode == Activity.RESULT_OK) {
+            val uri = resultData?.data
+            if (uri != null) {
+                val transaction = viewModel.transaction.value
+                fileSaverTask = FileSaverTask(this).apply {
+                    execute(Triple(type, uri, transaction!!))
```
vbuberen: I don't like forcing non-null, since we really can't be 100% sure about value of the `LiveData` being not null. 
Let's move `val transaction` outside `if` and change the check to 
`if (uri != null && transaction !=null)`

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -158,6 +228,47 @@ internal class TransactionPayloadFragment :
...
+        override fun doInBackground(vararg params: Triple<Int, Uri, HttpTransaction>): Boolean {
+            val (type, uri, transaction) = params[0]
+            try {
+                val context = fragment.context ?: return false
+                context.contentResolver.openFileDescriptor(uri, "w")?.use {
+                    FileOutputStream(it.fileDescriptor).use { fos ->
+                        when {
+                            type == TYPE_REQUEST ->
+                                transaction.requestBody?.byteInputStream()?.copyTo(fos)
+                            transaction.responseBody != null ->
```
vbuberen: I suggest to keep all condition checks in braces, as well as add curly braces for outcomes, because it is hard to read in such cases.

cortinico:
Thanks for the review @vbuberen Now is ready for another pass :)

vbuberen:
LGTM
Thanks @yoavst for implementation and @cortinico for finishing this PR.

PR #139, opened 2019-10-13
Remove deprecations

vbuberen:
This PR replaces some deprecations both in code and build configs.
Also, minor changes to drop things like @JvmStatic, which aren't required anymore after activities switched to Kotlin.


library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionActivity.kt
```diff
@@ -127,8 +126,8 @@ internal class TransactionActivity : BaseChuckerActivity() {
         startActivity(Intent.createChooser(sendIntent, null))
     }
 
-    internal inner class PagerAdapter(context: Context, fm: FragmentManager) : FragmentStatePagerAdapter(fm) {
-        private val context: WeakReference<Context> = WeakReference(context)
+    internal inner class PagerAdapter(fm: FragmentManager) :
+        FragmentStatePagerAdapter(fm, BEHAVIOR_RESUME_ONLY_CURRENT_FRAGMENT) {
```
olivierperez: Removing the `context` parameter and the WekRefenrence may produce memory leak.

Actually, it would be better to keep it and to remove the `inner` keyword. `inner` adds a hidden reference to the outside class.

vbuberen: Could you, please, explain how the removal of `context` could produce memory leak in this case?

As to `inner` - agree.

olivierperez: I mean, removing `context`+`WeakReference` forces you to use `inner class` + `this@TransactionActivity` to access a context.

So, **after compiling**, the class `PagerAdapter` will have a reference to `this@TransactionActivity` that is NOT a WeakReference.

vbuberen: Thanks, now it is clear to me. Yeah, agree. Will update this part.

olivierperez:
Thanks for contribution 👍

Issue #88, opened 2019-08-10
Provide an option to remove "errors" tab


ColtonIdle:
I use a few other crash reporters and crash viewers built into my application. The fact that Chucker added this onto Chuck is nice, but I want to get rid of it so my team doesn't get confused about whether or not we should be using it vs our other error reporters. During Chucker setup, it'd be nice to just opt-out of having the error feature showing/working.

cortinico:
@ColtonIdle 
Are you suggesting a way to hide the "Errors" tab right?

ColtonIdle:
Yep. Basically to only have the original functionality of chuck. Which is to inspect network traffic because I already use other tools for errors.

cortinico:
> Yep. Basically to only have the original functionality of chuck. Which is to inspect network traffic because I already use other tools for errors.

I see. As a side note: Chucker is not capturing crashes unless you explicitly call `collector.onError()`. So the Error tab won't be populated automatically.

uOOOO:
I'll PR after #98 is merged.

vbuberen:
Would like to return to this question. I haven't seen much feedback or requests about error reporting feature since I joined the team, so it just gives some points to my suggestion of removing this functionality at all and making Chucker focused on network only. 

I believe that it is better to focus on something in such tool and do it great. 
So I suggest to remove `Errors` tab at all.

cortinico:
I'm also in for removing it. Maybe we want to loop in with @olivierperez since he initially developed it. Are you actually using this feature? I would love to understand what's the adoption rate of it.

olivierperez:
Hi, I'm ok too. This feature seems to be unused.
Can we at least make a poll on Twitter to ask people if the want to keep it or drop it?

vbuberen:
Pool is a great idea. Let's do it :)

cortinico:
I'd rather create a poll as a Github Issue as the majority of our users are there. We can set the issue as sticky and have a deadline say 2/3 weeks. People can vote using reactions like:

👍 Let's keep the feature
👎 Let's remove the feature
👀 See comment below

vbuberen:
@cortinico Ok, I will create such issue. But I also would like to ask you to also post the link to tbis issue in Twitter, since your account is the most popular among us.

cortinico:
Sure let's do that 👍

vbuberen:
Closing this one due to deprecation of `Errors` tab

PR #141, opened 2019-10-15
Feature management for #88

olivierperez:
### Kotlin
```kotlin
configureChucker {
    http {
        enabled = true
        showNotification = true
        retentionPeriod = RetentionManager.Period.ONE_HOUR
        maxContentLength = 250000L
        headers {
            redact("Authorization")
            redact("Auth-Token")
            redact("User-Session")
        }
    }
    error {
        enabled = true
    }
}
```

### Java
```java
HashSet<String> headersToRedact = new HashSet<>();
headersToRedact.add("Authorization");
headersToRedact.add("Auth-Token");
headersToRedact.add("User-Session");

List<TabFeature> features = Arrays.asList(
        new HttpFeature(
                true,
                true,
                RetentionManager.Period.ONE_HOUR,
                DEFAULT_MAX_CONTENT_LENGTH,
                headersToRedact
        ),
        new ErrorsFeature(
                true,
                true
        )
);

ChuckerJavaConfig.configure(features);
```

## :page_facing_up: Context
* Some integrators want to be able to disable the errors feature #88 

## :pencil: Changes
* `FeatureManager` is an object that handle the **Features** system
  * It helps to know which features are enabled, and get their configurations
* `Feature` interface describe what is a Feature
  * Implemented by `HttpFeature` and `ErrorsFeature`
* Every code that concerns HTTP or ERRORS are now calling `FeatureManager` to check if feature is enabled

## :paperclip: Related PR
* #100 Was offering another way to achieve it, but last was done on Aug, 20

## :warning: Breaking
*None*

## :pencil: How to test
* Change configuration in sample -> ChuckerApplication

## :crystal_ball: Next steps
- [x] I still have to add the DSL in library-no-op :man_shrugging:


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -47,36 +51,13 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
-                charset = contentType.charset(UTF8) ?: UTF8
-            }
-            if (io.isPlaintext(buffer)) {
-                val content = io.readFromBuffer(buffer, charset, maxContentLength)
-                transaction.requestBody = content
-            } else {
-                transaction.isResponseBodyPlainText = false
-            }
-        }
+        return if (httpFeature.enabled) captureTransaction(request, chain)
```
vbuberen: What do you think about keeping the same style for `if else` everywhere? 
For example, in the same file there are lines the conditions have braces for `if` and `else` cases (like on line 127).
I believe that it could improve readability a little bit.

olivierperez: :+1:

olivierperez:
:warning: I haven't put the `maxContentLength` and `headersToRedact`, maybe I should. What to you think of it @cortinico @redwarp?

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
cortinico: We can enable the `MandatoryBracesIfStatements` detekt rule if you think it's the case

README.md
```diff
@@ -127,8 +117,8 @@ It is intended for **use during development**, and not in release builds or othe
 You can redact headers that contain sensitive information by calling `redactHeader(String)` on the `ChuckerInterceptor`.
 
 ```kotlin
-interceptor.redactHeader("Auth-Token");
-interceptor.redactHeader("User-Session");
```
cortinico: ++

README.md
```diff
@@ -113,7 +103,7 @@ Chucker supports also collecting and displaying **Throwables** of your applicati
 ```kotlin
 try {
     // Do something risky
-} catch (IOException exception) {
+} catch (exception: IOException) {
```
cortinico: :O my bad

library/src/main/java/com/chuckerteam/chucker/internal/support/ClearDatabaseService.kt
```diff
@@ -9,7 +9,7 @@ internal class ClearDatabaseService : IntentService(CLEAN_DATABASE_SERVICE_NAME)
 
     override fun onHandleIntent(intent: Intent?) {
         System.err.println("onHandleIntent")
-        when (val a = intent?.getSerializableExtra(EXTRA_ITEM_TO_CLEAR)) {
```
cortinico: You probably need to rebase here.

library/src/main/java/com/chuckerteam/chucker/api/config/Feature.kt
```diff
@@ -0,0 +1,14 @@
+package com.chuckerteam.chucker.api.config
+
+import android.content.Context
+import androidx.annotation.StringRes
+import androidx.fragment.app.Fragment
+
+interface Feature {
+    @get:StringRes
+    val name: Int
+    val tag: Int
```
cortinico: Tag is not really clear here what exactly it is

olivierperez: I've renamed it to `id`, is it clearer? Each feature should have a unique `id`.

olivierperez:
@cortinico 
I've read your comments and proposed a fix almost each of them.

I still have to:
- [ ] find a nice way avoid big breaking changes
- [ ] offer a API for Java developers to configure Chucker

cortinico:
Awesome, please let me know if you need some support :)

olivierperez:
@cortinico Done !
I've:
- Add `ChuckerJavaConfig` (see Example above)
- Reworked the DSL to redact headers (tell me what version do you prefer)

library/src/main/java/com/chuckerteam/chucker/api/ChuckerCollector.kt
```diff
@@ -48,8 +63,10 @@ class ChuckerCollector @JvmOverloads constructor(
      * @param transaction The HTTP transaction sent
      */
     internal fun onRequestSent(transaction: HttpTransaction) {
+        if (!httpFeature.enabled) return
```
vbuberen: I suggest to add braces here for readability and to match other code lines, like below where condition with only one line in it has braces.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerCollector.kt
```diff
@@ -61,8 +78,10 @@ class ChuckerCollector @JvmOverloads constructor(
      * @param transaction The sent HTTP transaction completed with the response
      */
     internal fun onResponseReceived(transaction: HttpTransaction) {
+        if (!httpFeature.enabled) return
```
vbuberen: Same suggestion about braces as above.

library-no-op/build.gradle
```diff
@@ -21,6 +21,7 @@ artifacts {
 dependencies {
     implementation "com.squareup.okhttp3:okhttp:$okhttp3Version"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    implementation "androidx.fragment:fragment:$fragmentVersion"
```
vbuberen: Maybe we could think about some other way of implementing feature management, which would allow to not include this dependency?

cortinico:
Thanks again for the big work @olivierperez 

I have to be honest, I'm a bit hesitant on merging this. I have the feeling the whole approach is a bit over engineered. My point of view is that is a bunch of extra code that we will have to maintain for a limited benefit.

Ideally we should be able to have some sort of key-value storage used to store/retrieve configurations for specific features.

Your approach is based on having a `FeatureManager` that holds a set of `TabFeature`. Here we're already making the assumption that every feature is rendered using a `Fragment` in a `ViewPager`. This has several side effects:

* The public API got way bigger (6 files were added in the `-no-op` library artifact).
* The `androidx.fragment:fragment` dependency leaked inside the `-no-op` library. We should probably avoid having `Fragments` as part of the public API.
* Having features that doesn't have a corresponding tab will require a refactoring, as well as features that are shared between two tabs.

Moreover, the DSL could be moved to a `library-dsl` artifact, and have it as a pluggable add-on. 

I generally have the feeling that we should simplify a bit the whole approach.

On the other hand I strongly believe that we need some sort of modular infrastructure to make easy to Chucker extend.

cortinico:
Also I invite the community, as well as @vbuberen @redwarp to provide their point of view on the topic :) Happy to discuss

library/src/main/java/com/chuckerteam/chucker/api/ChuckerCollector.kt
olivierperez: I use braces when there is something to do.
My point of view is this is an early return, in order to do nothing. If I do nothing I have no braces 🤷‍♂ 

What do you think of that?

cortinico: > I use braces when there is something to do.

Agree with @olivierperez 
Also the style guide here https://developer.android.com/kotlin/style-guide#braces allows single-line if statements like:

```kotlin
if (condition) effect
```

vbuberen:
From my point of view, this is too much for Chucker at the moment. 
Having 2 tabs with transactions and errors isn't something too complex or requiring some additional efforts from users, who want just watch for transactions or for those, so want to only see throwable 
errors.
Moreover, talking about errors - now errors are tracked by default, while with these changes user should explicitly write 
`error {
        enabled = true
    }`
So I agree with @cortinico here on being hesitant on merging this thing.

olivierperez:
Hi, thanks for the review and the comments and sorry for the delay.
I think the way I choose is not really the right one. If we want to implement a way to configure features we should talk about it before I go develop it :-)

**I propose to close this PR.**

BTW _It was fun to develop it anyway_

cortinico:
> **I propose to close this PR.**
> BTW _It was fun to develop it anyway_

Let's close it 👍 I anyway believe that we might come back to this topic at a certain topic and your PR could be used as material to kickoff the discussion again :)

PR #142, opened 2019-10-17
Cleanup kotlin classes

rakshit444:
optimized kotlin code


library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
```diff
@@ -11,35 +11,37 @@ internal class HttpTransactionDatabaseRepository(private val database: ChuckerDa
 
     private val executor: Executor = Executors.newSingleThreadExecutor()
 
+    private val transcationDao = database.transactionDao()
```
cortinico: This is not equivalent to the previous code as you're retaining an instance of `HttpTransactionDao` inside the repository. This behavior is already offered by the Room generated code (inside `ChuckerDatabase_Impl`)

Please change with a property with no backing field:

```suggestion
    private val transcationDao get() = database.transactionDao()
```

cortinico:
Thanks for the contribution ❤️

PR #145, opened 2019-10-19
Introduce a shared view model into transaction details activity + fragments

psh:
## :pencil: Changes
* Added a shared view model (created in `TransactionActivity` also shared between `TransactionOverviewFragment` and `TransactionPayloadFragment`)
* Removed the previous callback based `TransactionFragment` where the view held onto state
* Light Kotlin cleanup (use of *let* rather than null checking)
* No need to hold a weak reference to a context in `TransactionPagerAdapter`
* Fixed pre-existing lint issues in `TransactionPagerAdapter`


cortinico:
Thanks for the PR @psh 

Can I ask you to provide a bit more context on why you would love to add a shared `ViewModel`. Is unclear to me if you're doing this change because you would love to get rid of the `WeakReference` or you prefer to introduce the `ViewModel` pattern in the library?

psh:
Two problems of holding state in the wrong place - 
* Data for the current transaction was being held in the View.  Since the project already used `LiveData` from _Architecture Components_ it seemed logical to adopt a `ViewModel` to hold on to state rather than any of the alternatives.  
* I removed the Weak Reference because the alternative (using the `Context` immediately) reduced complexity and reduced the small chance that we would end up with blank titles (if the Weak Reference went out of scope before tab titles were requested)

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
+import com.chuckerteam.chucker.internal.data.repository.RepositoryProvider
+import com.chuckerteam.chucker.internal.support.observeOnce
+
+class TransactionViewModel(private val transactionId: Long) : ViewModel() {
+    val transactionTitle: MutableLiveData<String> = MutableLiveData()
```
cortinico:
```suggestion
    internal val transactionTitle: MutableLiveData<String> = MutableLiveData()
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+    internal var transaction: MutableLiveData<HttpTransaction> = MutableLiveData()
```
cortinico:
```suggestion
    internal val transaction: MutableLiveData<HttpTransaction> = MutableLiveData()
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.Observer
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
+import com.chuckerteam.chucker.internal.data.repository.RepositoryProvider
+import com.chuckerteam.chucker.internal.support.observeOnce
+
+class TransactionViewModel(private val transactionId: Long) : ViewModel() {
```
cortinico:
```suggestion
internal class TransactionViewModel(private val transactionId: Long) : ViewModel() {
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+            .observeOnce(
+                Observer {
+                    transactionTitle.value = if (it != null) "${it.method} ${it.path}" else ""
+                    transaction.value = it
+                }
+            )
+    }
+}
+
+class TransactionViewModelFactory(private val transactionId: Long = 0L) :
```
cortinico:
```suggestion
internal class TransactionViewModelFactory(private val transactionId: Long = 0L) :
```

PR #147, opened 2019-10-23
Remove unused variable in ClearDatabaseService

cortinico:
Looks like we just have an unused variable that can definitely go


library/src/main/java/com/chuckerteam/chucker/internal/support/ClearDatabaseService.kt
```diff
@@ -9,7 +9,7 @@ internal class ClearDatabaseService : IntentService(CLEAN_DATABASE_SERVICE_NAME)
 
     override fun onHandleIntent(intent: Intent?) {
         System.err.println("onHandleIntent")
```
vbuberen: I believe that it is not correct to use `err` here, since it is called from normal action, not error. Suggest to replace it or remove.

PR #148, opened 2019-10-24
Another fix for notifications in Q

vbuberen:
Basically, this PR is just a duplicate of #144.
It appears that after another merged PR `NotificationsHelper.java` file with my fix was deleted and replaced by its Kotlin version. So I am just restoring that from #144 .

However, I also found that one of functions have `@JvmStatic` annotation, while not being called from Java class anywhere in the project. So removed this annotation.


cortinico:
Nice catch!

PR #149, opened 2019-10-24
Update/share util

vbuberen:
## :page_facing_up: Context
During work on previous PRs found out that `Share` action in `TransactionActivity` and in `ErrorActivity` is duplicated as well as has different style. So my changes should help unify creation of `Intent` to share details about transaction or error.

## :pencil: Changes
Added a new class called `ShareUtil` with extension functions, which call the same private function to create chooser to share text from transaction or error.


library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtil.kt
```diff
@@ -0,0 +1,41 @@
+package com.chuckerteam.chucker.internal.support
+
+import androidx.appcompat.app.AppCompatActivity
+import androidx.core.app.ShareCompat
+import com.chuckerteam.chucker.R
+
+fun AppCompatActivity.shareError(errorText: String) {
```
cortinico: I'm a bit against those extensions function. I would prefer to have those two as methods inside the `TransactionActivity` and `ErrorActivity`. First those two functions are public and will be exposed to users of Chucker (that is nothing we would love to do).

Moreover I'd try to the keep all the UI related code (e.g. `getString(R.string.chucker_share_error_title)`) closer to the Activity rather than inside a Util class.

vbuberen: Yeah, my mistake about exposing.
As to keep `R` class - agree as well. 

Will reiterate on this one.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionActivity.kt
```diff
@@ -114,16 +115,19 @@ internal class TransactionActivity : BaseChuckerActivity() {
...
-            action = Intent.ACTION_SEND
-            putExtra(Intent.EXTRA_TEXT, content)
-            type = "text/plain"
-        }
-        startActivity(Intent.createChooser(sendIntent, null))
+    private fun share(transactionDetailsText: String) {
+        startActivity(
+            ShareCompat.IntentBuilder.from(this)
+                .setType(MIME_TYPE)
+                .setChooserTitle(getString(R.string.chucker_share_error_title))
```
cortinico:
```suggestion
                .setChooserTitle(getString(R.string.chucker_share_transaction_title))
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/error/ErrorActivity.kt
```diff
@@ -75,20 +75,20 @@ internal class ErrorActivity : AppCompatActivity() {
...
             throwable.content
         )
-
         startActivity(
             ShareCompat.IntentBuilder.from(this)
-                .setType(TEXT_PLAIN)
-                .setSubject(getString(R.string.chucker_share_error_title))
-                .setText(text)
+                .setType(MIME_TYPE)
+                .setChooserTitle(getString(R.string.chucker_share_transaction_title))
```
cortinico:
```suggestion
                .setChooserTitle(getString(R.string.chucker_share_error_title))
```

vbuberen: Fixed.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionActivity.kt
vbuberen: Fixed.

cortinico:
Awesome, thanks for the PR <3

PR #143, opened 2019-10-17
Feature/dark theme

vbuberen:
## :pencil: Changes
This PR adds cool new thing for devs, who love to debug at night - dark theme!
Also, I did a migration to new Material theme completely, since Chucker already had corresponding library in dependencies. So, all the colors are defined in a new way as it is described in [Material guidelines ](https://material.io/design/color/dark-theme.html#anatomy).
Themes switch automatically according to device settings.

Updated app theme was tested on APIs down to 16 to make sure I didn't break anything.

I tried to preserve original colors, but for `night` theme completely followed guidelines and used less saturated colors to lower eye strain. This is also the reason why I changed `highlight` util function to receive colors for spans as parameters, because standard yellow and red colors were too saturated for `night` theme (compare screenshots below).

PR #156, opened 2019-11-09
Update assets to match changed design

vbuberen:
## :page_facing_up: Context
Since in #143 I updated themes and made Chucker look quite different, I feel that I should also take care of animation assets and update them as well. Also, old assets were not compressed, so were not very efficient in terms of loading.

## :pencil: Changes
No code changes for this PR. Only updated GIFs, which were freshly recorded and compressed to take as less space as it possible. 

## :crystal_ball: Next steps
Suggest to merge this PR along with release of `3.1.0`, which will have updated theme.


cortinico:
> Suggest to merge this PR along with release of 3.1.0, which will have updated theme.

Let's merge it now. We keep on `develop` the latest version (the one with Dark theme, etc.). We sync the `release` branch once we release a new version.

Issue #163, opened 2019-11-26
Crash when chucker is used with proguard enabled.


prithvibhola:
Chucker is crashing with these crash logs when clicked on the notification. It is working fine in the debug build when proguard is not enabled but crashing when the proguard is enabled.

2019-11-26 17:39:06.222 18381-18381/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.gojek.driver.bike.staging, PID: 18381
    android.view.InflateException: Binary XML file line #17: Binary XML file line #17: Error inflating class <unknown>
    Caused by: android.view.InflateException: Binary XML file line #17: Error inflating class <unknown>
    Caused by: java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
        at android.view.LayoutInflater.createView(LayoutInflater.java:647)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:790)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:730)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:492)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:423)
        at dark.Ɨſ.onCreateView(:65)
        at dark.ιʟ.performCreateView(:2439)
        at dark.Іɹ.ˊ(:1460)
        at dark.Іɹ.ॱॱ(:1784)
        at dark.Іɹ.ॱ(:1852)
        at dark.ϙ.ॱॱ(:802)
        at dark.Іɹ.ˋ(:2625)
        at dark.Іɹ.ˊ(:2411)
        at dark.Іɹ.ˋ(:2366)
        at dark.Іɹ.ˏ(:2243)
        at dark.ϙ.ʼ(:654)
        at dark.іɹ.finishUpdate(:146)
        at dark.ɨӏ.ˊ(:1244)
        at dark.ɨӏ.ˏ(:1092)
        at dark.ɨӏ.onMeasure(:1622)
        at android.view.View.measure(View.java:23355)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6758)
        at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1535)
        at android.widget.LinearLayout.measureVertical(LinearLayout.java:825)
        at android.widget.LinearLayout.onMeasure(LinearLayout.java:704)
[...]

PR #166, opened 2019-12-13
Crash fix when proguard is enabled

prithvibhola:
## :page_facing_up: Context
When the build is created with proguard enabled, the chucker crashes when clicked from the notification.
Issue number : #163 

## :pencil: Changes
Removed **app:layoutManager="LinearLayoutManager"** from chucker_fragment_transaction_list.xml and chucker_fragment_error_list.xml
As the LinearLayoutManager is already set in the kt files for Recyclerview, the app is working as expected after removing LayoutManger from the XML files.

## :pencil: How to test
Normal build with proguard enabled can be created with **-repackageclasses "package"** in the proguard file. This issue is replicable every time.

## :crystal_ball: Next steps
As the LayoutManager is already set in the kt files nothing else is required.

### DIFF
--- a/library/src/main/res/layout/chucker_fragment_error_list.xml
+++ b/library/src/main/res/layout/chucker_fragment_error_list.xml
@@ -25,7 +25,6 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:scrollbars="vertical"
-        app:layoutManager="LinearLayoutManager"
         tools:listitem="@layout/chucker_list_item_error" />
 
     <LinearLayout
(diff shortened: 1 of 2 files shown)


vbuberen:
Would to note here as well that the issue appears due to a specific use case @prithvibhola has in his application, which he already mentioned [here](https://github.com/ChuckerTeam/chucker/issues/163#issuecomment-565367770)
I suggested another solution in #167 which makes the code slightly cleaner as it seems for me.

cortinico:
Closing as this is superseded by @vbuberen's PR #167 
Thanks again for submitting this PR @prithvibhola and apologize for the major delay :pray:

PR #169, opened 2019-12-30
Minor code cleanup

vbuberen:
## :page_facing_up: Context
Most of the code is already migrated to Kotlin, which means we don't need `@JvmStatic` annotated functions anymore. Since this annotation generates additional methods during compilation, I believe Chucker should get rid of it.
Also, some code parts 

## :pencil: Changes
- Removed @JvmStatic annotations almost everywhere (a few usages left)
- Used String interpolation in a few places to support switching to unified code style across the project
- Added a few `const` keywords in database `companion object`

## :crystal_ball: Next steps
Remove other `@JvmStatic` annotations when Kotlinification of `FormatUtils` class will be merged into `develop`

### DIFF
--- a/library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
@@ -11,7 +11,7 @@ object Chucker {
     const val SCREEN_HTTP = 1
     const val SCREEN_ERROR = 2
 
-    const val isOp = false
+    val isOp = false
 
     @JvmStatic
     fun getLaunchIntent(context: Context, screen: Int): Intent {
--- a/library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
@@ -11,20 +11,24 @@ object Chucker {
     const val SCREEN_HTTP = 1
     const val SCREEN_ERROR = 2
 
-    val isOp = false
+    const val isOp = false
 
+    @JvmStatic
     fun getLaunchIntent(context: Context, screen: Int): Intent {
         return Intent()
     }
 
+    @JvmStatic
     fun registerDefaultCrashHandler(collector: ChuckerCollector) {
         // Empty method for the library-no-op artifact
     }
 
+    @JvmStatic
     fun dismissTransactionsNotification(context: Context) {
         // Empty method for the library-no-op artifact
     }
 
+    @JvmStatic
     fun dismissErrorsNotification(context: Context) {
         // Empty method for the library-no-op artifact
     }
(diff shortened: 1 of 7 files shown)


library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -13,19 +13,19 @@ object Chucker {
 
     val isOp = false
 
-    @JvmStatic fun getLaunchIntent(context: Context, screen: Int): Intent {
```
cortinico: Ideally we should keep those `@JvmStatic`, at least for all the methods inside `com.chuckerteam.chucker.api`.

Removing this annotation will generate a breaking change + will make Chucker less Java-friendly.

cortinico:
Other than the `@JvmStatic` comment, the code looks go to me 👍

library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
vbuberen: Yeah, somehow thought that it is only fo internal usages. Reverting.

library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -11,21 +11,25 @@ object Chucker {
     const val SCREEN_HTTP = 1
     const val SCREEN_ERROR = 2
 
-    val isOp = false
+    const val isOp = false
```
cortinico: I'm afraid you have to revert this line (unless we come up with a different solution).
I just realized that `const val` properties won't generate getters/setters in the Bytecode. This will result in a breaking change on the Java side of things.

tl;dr: Java users accessing `Chucker.isOp()` will have a compilation issue as now it should be accessed as `Chucker.isOp()`

We should consider maybe leaving a note over here and ship this change with Chucker 4.x

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -12,11 +12,14 @@ import com.chuckerteam.chucker.internal.ui.MainActivity
...
 
+    const val SCREEN_HTTP = 1
+    const val SCREEN_ERROR = 2
+
     /**
      * Check if this instance is the operation one or no-op.
      * @return `true` if this is the operation instance.
      */
-    val isOp = true
+    const val isOp = true
```
cortinico: As above

library-no-op/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
vbuberen: Fixed it. The possibility is quite low, but yes, better safe than sorry.

vbuberen: @cortinico There was a typo, probably in your tl;dr. Could you fix it, since you referenced this comment in the library code and we shouldn't confuse anyone.

cortinico: Yup thanks for noticing it

PR #172, opened 2020-01-02
Fixing handling of big body payloads

### LINKED ISSUE #82: Large json responses result in bad user experience

cortinico:
Currently Chucker UI hangs when the HTTP request/response body is too
big. The issue is caused by having to put a bunch of text inside a
single TextView.

This commit refactors the request/response body screen to use a
RecyclerView to load single lines. In this way the UI is much smoother
and the experience is generally better.

## :camera: GIF
bigpayload

## :page_facing_up: Context
Fixes #82

## :warning: Breaking
Only changes inside `.internal` so no breaking changes on the API.
On the other hand there are a couple of drawbacks:

- Text in the body is selectable only line by line.
- Search in the body works only line by line.
- Response/Request screen are not inside a `NestedScrollView` anymore (as this had major performance drawbacks) 

## :pencil: How to test
Ideally we should extend the sample app with a single call to https://data.cityofnewyork.us/api/views/kku6-nxdu/rows.json or to another big JSON api.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
@@ -61,13 +61,13 @@ internal class TransactionBodyAdapter(
                     item.line.clearSpans()
                     item.line = item.line.toString()
                         .highlightWithDefinedColors(newText, backgroundColor, foregroundColor)
-                    notifyItemChanged(index)
+                    notifyItemChanged(index + 1)
                 } else {
                     // Let's clear the spans if we haven't found the query string.
                     val spans = item.line.getSpans(0, item.line.length - 1, Any::class.java)
                     if (spans.isNotEmpty()) {
                         item.line.clearSpans()
-                        notifyItemChanged(index)
+                        notifyItemChanged(index + 1)
                     }
                 }
             }
@@ -80,7 +80,7 @@ internal class TransactionBodyAdapter(
                 val spans = item.line.getSpans(0, item.line.length - 1, Any::class.java)
                 if (spans.isNotEmpty()) {
                     item.line.clearSpans()
-                    notifyItemChanged(index)
+                    notifyItemChanged(index + 1)
                 }
             }
     }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
@@ -1,0 +1,130 @@
+package com.chuckerteam.chucker.internal.ui.transaction
+
+import android.graphics.Bitmap
+import android.text.SpannableStringBuilder
+import android.text.Spanned
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
+import com.chuckerteam.chucker.R
+import com.chuckerteam.chucker.internal.support.highlightWithDefinedColors
+
+/**
+ * Adapter responsible of showing the content of the Transaction Request/Response body.
+ * We're using a [RecyclerView] to show the content of the body line by line to do not affect
+ * performances when loading big payloads.
+ */
+internal class TransactionBodyAdapter(
+    private val bodyItems: List<TransactionPayloadItem>
+) : RecyclerView.Adapter<TransactionPayloadViewHolder>() {
+
+    override fun onBindViewHolder(holder: TransactionPayloadViewHolder, position: Int) {
+        holder.bind(bodyItems[position])
+    }
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): TransactionPayloadViewHolder {
+        val inflater = LayoutInflater.from(parent.context)
+        return when (viewType) {
+            TYPE_HEADERS -> {
+                val view = inflater.inflate(R.layout.chucker_transaction_item_headers, parent, false)
+                TransactionPayloadViewHolder.HeaderViewHolder(view)
+            }
+            TYPE_BODY_LINE -> {
+                val view = inflater.inflate(R.layout.chucker_transaction_item_body_line, parent, false)
+                TransactionPayloadViewHolder.BodyLineViewHolder(view)
+            }
+            else -> {
+                val view = inflater.inflate(R.layout.chucker_transaction_item_image, parent, false)
+                TransactionPayloadViewHolder.ImageViewHolder(view)
+            }
+        }
+    }
+
+    override fun getItemCount() = bodyItems.size
+
+    override fun getItemViewType(position: Int): Int {
+        return when (bodyItems[position]) {
+            is TransactionPayloadItem.HeaderItem -> TYPE_HEADERS
+            is TransactionPayloadItem.BodyLineItem -> TYPE_BODY_LINE
+            is TransactionPayloadItem.ImageItem -> TYPE_IMAGE
+        }
+    }
+
+    internal fun highlightQueryWithColors(newText: String, backgroundColor: Int, foregroundColor: Int) {
+        bodyItems.filterIsInstance<TransactionPayloadItem.BodyLineItem>()
+            .withIndex()
+            .forEach { (index, item) ->
+                if (newText in item.line) {
+                    item.line.clearSpans()
+                    item.line = item.line.toString()
+                        .highlightWithDefinedColors(newText, backgroundColor, foregroundColor)
+                    notifyItemChanged(index)
+                } else {
+                    // Let's clear the spans if we haven't found the query string.
+                    val spans = item.line.getSpans(0, item.line.length - 1, Any::class.java)
+                    if (spans.isNotEmpty()) {
+                        item.line.clearSpans()
+                        notifyItemChanged(index)
+                    }
+                }
+            }
+    }
+
+    internal fun resetHighlight() {
+        bodyItems.filterIsInstance<TransactionPayloadItem.BodyLineItem>()
+            .withIndex()
+            .forEach { (index, item) ->
+                val spans = item.line.getSpans(0, item.line.length - 1, Any::class.java)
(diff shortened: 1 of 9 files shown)


ColtonIdle:
Any chance that we can use the "old" method of just setting text if the response is less than 0.5MB or something. And if it's larger than that, then we go with the recyclerView route?

I would say that very large json is an exception, but for me it's not even that important to view it on the device... I just don't want it to freeze, and I don't want it to crash/freeze when sending the results to my teammate on slack, email, etc.

cortinico:
> Any chance that we can use the "old" method of just setting text if the response 
> is less than 0.5MB or something. And if it's larger than that, then we go with 
> the recyclerView route?

What's the reason behind this? Are you afraid of any of the drawbacks? I would prefer to have a single way of showing the body honestly rather than branching.

> I just don't want it to freeze, and I don't want it to crash/freeze when sending > the results to my teammate on slack, email, etc.

The proposed implementation won't freeze/crash in those scenarios. On the other hand, please consider that sharing via `Intent` (e.g. to Slack) is limited at around 500Kb. So sharing as plain text won't work in those scenarios. We would have to wait for #138 to offer a way to save the body in a file.

ColtonIdle:
>What's the reason behind this? Are you afraid of any of the drawbacks? I would prefer to have a single way of showing the body honestly rather than branching.

A use case I use often is long holding on the response text and copying that into slack for example.

>The proposed implementation won't freeze/crash in those scenarios. On the other hand, please consider that sharing via Intent (e.g. to Slack) is limited at around 500Kb. So sharing as plain text won't work in those scenarios. We would have to wait for #138 to offer a way to save the body in a file.

Yes! I'm aware that it wouldn't fix the sharing issue, but I was more-so thinking of the easiest solution for the "large response issue". The large response issue to me is that the app freezes and it can't share. The "solution" to that for me was to only show response if it's less than 0.5MB or 1MB (some kind of sane default) and then when sharing, it would create a file on the fly of the response. In my head we would have a boolean that was `isLargeResponse` and if it is, then we don't show the text response, and just have a message that says "Response too big" with a save button and offload the issue of showing a huge piece of text to a text editor or some kind of other app on the phone that knows how to deal with huge pieces of text.

I do like your RV solution! And I do like the solution of having a save icon. I do wish that share would implicitly save if it's above 0.5MB and share the file instead of crashing right now or me having to save and then share. Also, not being able to select the full text when long holding would be a bit inconvenient in my opinion... which is why I was saying that for 0.5MB and less it just works as it does because it works great. The exceptional case is a large payload, and in that case, make the user suffer with worse search and worse copy/paste ability.

Overall anything will be better than what happens in Chucker today. 😄

vbuberen:
I really like the approach of splitting the text body. 

But since we are targeting to reach best performance possible, what do you think if we also engage `PrecomputedText`? It might help for big body payloads.
Also, as to the adapter used - I would suggest to switch to `ListAdapter` instead of default `RecyclerView.Adapter`, so we could also use cool `AsyncListDiffer`.

If you would like I can help with both points.

cortinico:
> Overall anything will be better than what happens in Chucker today.

Agree. I suggest to try to merge this and eventually re-iterate further on branching and refining the user experience. With this PR at least we prevent freezes and crashes so it's already a huge improvements imho.

> But since we are targeting to reach best performance possible, what do you 
> think if we also engage PrecomputedText? It might help for big body payloads.

I played around with `PrecomputedText` before trying the RV approach and the result was not satisfying (as we anyway had a really long text). We can think about using `PrecomputedText` for every line, but I'd rather have it in a separate PR as this is getting bigger.

> Also, as to the adapter used - I would suggest to switch to ListAdapter instead 
> of default RecyclerView.Adapter, so we could also use cool AsyncListDiffer.

Again I'll keep it in a separate PR if possible (I don't think is a blocker for this one). I played also with `ListAdapter` and I don't see much benefit for our scenario given that:

* We would have to implement the diffing logic for classes like `Spanned`, `SpannableStringBuilder` and `Bitmap` that is not ideal.
* Our list is pretty static. The only change we apply is the highlight, and we call `notifyItemChanged` on the affected items only.

I'm happy to hand over the `ListAdapter` implementation to you @vbuberen if you wish :)

ColtonIdle:
@cortinico sounds good. Let's get this merged! Bye bye ANR!

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    companion object {
+        private const val TYPE_HEADERS = 1
+        private const val TYPE_BODY_LINE = 2
+        private const val TYPE_IMAGE = 3
+    }
+}
+
+internal sealed class TransactionPayloadViewHolder(view: View) : RecyclerView.ViewHolder(view) {
```
vbuberen: Could you please explain why this class is `sealed`?

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -194,40 +177,83 @@ internal class TransactionPayloadFragment :
...
+                result.add(
+                    TransactionPayloadItem.HeaderItem(
+                        HtmlCompat.fromHtml(headersString, HtmlCompat.FROM_HTML_MODE_LEGACY)
+                    )
                 )
             }
+
+            // The body could either be an image, binary encoded or plain text.
+            if (type == TYPE_RESPONSE && transaction.responseImageBitmap != null) {
+                transaction.responseImageBitmap?.let {
```
vbuberen: Isn't `responseImageBitmap` already checked for `null`?

library/src/main/res/layout/chucker_transaction_item_body_line.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/body_line"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:textIsSelectable="true"
+    android:minLines="1"
+    android:gravity="left"
```
vbuberen: Maybe replace with `start`?

vbuberen:
Found a few issues with search:
1. Highlighting works incorrectly.
2. Items blink during search (note `args` line of body in attached gif).

rv_search

vbuberen:
@cortinico 
I agree with your point about merging this and come back later. Please, fix issues I mentioned in the message above and we are good to go :)

As to `PrecomputedText` and `ListAdapter` - yes, I would love to try it out later.

Also, after looking through the discussion in the issue and here in the PR I am curious if we should consider evaluating other solutions for Json parsing in terms of performance.

ColtonIdle:
@vbuberen 

` I am curious if we should consider evaluating other solutions for Json parsing in terms of performance.`

Do you mean replacing Gson?

vbuberen:
> Do you mean replacing Gson?

Yes. For example, `Kotlinx.Serialization` might be a good candidate.
Here is one of articles with benchmarks: https://tech.olx.com/android-jetpack-benchmark-json-parsers-performance-1e76031a296b

This is not a top priority at the moment from my point of view, but definitely worth it.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
cortinico: `responseImageBitmap` is actually a property with a custom getter. Resolved storing the value in an immutable variable.

cortinico:
> Found a few issues with search

@vbuberen Can I ask you to verify again if you still have this issue (with the lasted commit on my branch). I also noticed those issues and I fixed them in d290e37fc8c0dbcddf941d84485f01fca116c8a9

> Do you mean replacing Gson?

> This is not a top priority at the moment from my point of view, but definitely worth it.

@vbuberen @ColtonIdle I'm in for having Moshi instead of Gson. This will probably have the smallest impact on user dependecies as the Retrofit + OkHTTP + Moshi stack is really common.

vbuberen:
@cortinico Sure, give me a few minutes to test.

As to JSON parsing alternatives - I also was for Moshi initially, but if you look at the results inside the article I shared above it seems that it is not that good in terms of performance. This was one of the reasons I mentioned `Kotlinx.Serialization` as an option.

Ideally, we should create our own set of tests and see the results 🤷‍♂😄

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
vbuberen: It is not a concern. I was just curious why it is `sealed`. But now I got it. Thanks.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    override fun getItemViewType(position: Int): Int {
+        return when (bodyItems[position]) {
+            is TransactionPayloadItem.HeaderItem -> TYPE_HEADERS
+            is TransactionPayloadItem.BodyLineItem -> TYPE_BODY_LINE
+            is TransactionPayloadItem.ImageItem -> TYPE_IMAGE
+        }
+    }
+
+    internal fun highlightQueryWithColors(newText: String, backgroundColor: Int, foregroundColor: Int) {
```
vbuberen: `nitpick` Maybe we could rename `newText` to `query` here?

vbuberen:
@cortinico No issues with search now. Also, re-checked the code - LGTM. 
If you plan no other changes, let's merge this thing :)

Issue #179, opened 2020-01-06
Internal classes are visible to outsiders


vbuberen:
## :loudspeaker: Describe the bug
Internal Chucker classes are visible for apps using this library.

### :wrench: Expected behavior
Internal implementation shouldn't be visible in projects with Chucker added as dependency.

## :camera: Screenshots
Screenshots from real project, which uses Chucker for debugging purposes
<img width="480" alt="Screen Shot 2020-01-06 at 15 45 28" src="https://user-images.githubusercontent.com/13467769/71822376-5d356300-309d-11ea-9f9c-3c2bcf853a07.png">
<img width="508" alt="Screen Shot 2020-01-06 at 15 46 38" src="https://user-images.githubusercontent.com/13467769/71822377-5d356300-309d-11ea-916c-d75bfd43fff4.png">

cortinico:
Yup that's actually a bug. Whatever is inside the `.internal` package should have the `internal` modifier.

vbuberen:
Closing as #180 was just merged.

Issue #134, opened 2019-10-10
More than one file was found with OS independent path 'META-INF/library_release.kotlin_module'


AlbertVilaCalvo:
Hi, when I add the library to the `app/build.gradle` file:

```
debugImplementation "com.github.ChuckerTeam.Chucker:library:3.0.1"
releaseImplementation "com.github.ChuckerTeam.Chucker:library-no-op:3.0.1"
```
I get the following build error:

```
Execution failed for task ':app:mergeDebugJavaResource'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
   > More than one file was found with OS independent path 'META-INF/library_release.kotlin_module'
```

Indeed, I have 2 libraries with this file, one of them is chucker:

![Captura de Pantalla 2019-10-10 a les 12 58 32](https://user-images.githubusercontent.com/4823860/66563306-cf111580-eb5d-11e9-8d06-878e5d5dc903.png)

There is a way to fix this with `packagingOptions`:

```
packagingOptions {
    exclude 'META-INF/library_release.kotlin_module'
}
```

However it seems that the root cause can be solved on the library, which is better because it avoids having everyone to add the exclude fix:

https://github.com/Takhion/android-extras-delegates/issues/1
https://github.com/Kodein-Framework/Kodein-DI/issues/138

Would be nice if this was fixed at the library :)

AlbertVilaCalvo:
I've found this related to the error: https://readyset.build/build-failures-kotlin-metadata-errors-f0090cdf948c

AlbertVilaCalvo:
I've opened an issue on the other library too: https://github.com/kizitonwose/CalendarView/issues/65

cortinico:
Thanks for the report @AlbertVilaCalvo 
Fix should be #146

jhamin0511:
It is work for me.

    packagingOptions {
        exclude 'META-INF/library_release.kotlin_module'
    }

https://stackoverflow.com/questions/44509608/duplicate-files-copied-in-apk-meta-inf-library-release-kotlin-module

PR #180, opened 2020-01-07
Visibility fix for internal members

vbuberen:
## :page_facing_up: Context
Fix for #179 

## :pencil: Changes
- Updated visibility modifiers
- Changed fun for starting TransactionActivity, since after merging PR with `ViewModel` this fun was reverted to state before #137 
- Fixed typo, which appeared after #146 
- Replaced deprecated calls for JsonParser

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -8,7 +8,7 @@ android {
     compileSdkVersion rootProject.compileSdkVersion
 
     compileOptions {
-        kotlinOptions.freeCompilerArgs += ['-module-name', "com.github.ChuckerTeam.Chucker.library-no-op"]
+        kotlinOptions.freeCompilerArgs += ['-module-name', "com.github.ChuckerTeam.Chucker.library"]
     }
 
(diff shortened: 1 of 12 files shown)


library/build.gradle
```diff
@@ -8,7 +8,7 @@ android {
     compileSdkVersion rootProject.compileSdkVersion
 
     compileOptions {
-        kotlinOptions.freeCompilerArgs += ['-module-name', "com.github.ChuckerTeam.Chucker.library-no-op"]
```
cortinico: ++

ColtonIdle: Unfamiliar with what's going on here. Can someone explain?

cortinico: Sure. This was a fix I introduced for #134 (in #146). I accidentally copy and pasted the same line for `library` and `library-no-op` without adapting the module name to the proper artifact name.

vbuberen:
@cortinico This PR is ready for review.

PR #181, opened 2020-01-10
Mask true type of headersToRedact and update documentation

sprohaszka:
## :page_facing_up: Context
When using ```ChuckerInterceptor```, you are forced to used a ```MutableSet``` for ```headersToRedact``` parameter.
It is better to mask the true type of ```headersToRedact```  and expose a simple ```Set```.

## :pencil: Changes
I change ```ChuckerInterceptor``` to accept a simple ```Set```.
I also upgrade *README* which was correct for this parameter.

### DIFF
--- a/sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
+++ b/sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
@@ -30,7 +30,7 @@ class HttpBinClient(
         context = context,
         collector = collector,
         maxContentLength = 250000L,
-        headersToRedact = emptySet()
+        headersToRedact = emptySet<String>()
     )
 
     private val httpClient =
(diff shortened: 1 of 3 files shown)


sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
```diff
@@ -29,7 +29,8 @@ class HttpBinClient(
     private val chuckerInterceptor = ChuckerInterceptor(
         context = context,
         collector = collector,
-        maxContentLength = 250000L
+        maxContentLength = 250000L,
+        headersToRedact = emptySet()
```
cortinico:
```suggestion
        headersToRedact = emptySet<String>()
```

cortinico:
Thanks for the PR @sprohaszka. Please note that your code is not compiling. Let's make the fix I suggested and then we should be good to go.

vbuberen:
Hey @sprohaszka 
Thanks for you contribution.
Will you be able to fix your PR?

sprohaszka:
It's done :)

vbuberen:
LGTM 🚢

Issue #154, opened 2019-11-07
Configure notification priority


vibin:
Hi, I would love to have a way to set Chuck's notification priority to NotificationCompat.PRIORITY_MIN (https://stackoverflow.com/a/16999675) so that Android can hide it from the status bar (it gets pretty annoying otherwise, for someone using a debug app all day :)).

Let me know if this sounds good. I can take a shot at this, this weekend.

vbuberen:
Why don't you just set priority for Chuck right on a device?

I believe that your use case is quite rare and device settings should be enough to handle it.
Chucker is a debug tool after all, so it is Ok to get notifications :)

vibin:
Notification priority set by user gets cleared whenever app is reinstalled - which can be very common in debug apps to change environment, etc. 

> Chucker is a debug tool after all, so it is Ok to get notifications

My issue is just about the status bar icon, not the notification itself :)

vibin:
@cortinico what do you think? Notification channels are immutable once created, so it's impossible to do this in the client code.

cortinico:
> @cortinico what do you think? Notification channels are immutable once created, so it's impossible to do this in the client code.

Hi @vibin 
Thanks for raising the issue and apologize for the extremely late reply.

To answer your question: you can re-define a Notification Channel using the same Channel ID set by Chucker (that is `chucker`):

https://github.com/ChuckerTeam/chucker/blob/012e6582872a211dcffd98fd59603495c1cf7666/library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt#L21

This will override the priority originally set by Chucker. 
Just call `createNotificationChannel` in your client in a way like this:

```kotlin
(context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager).createNotificationChannel(
                NotificationChannel(
                        "chucker",
                        "Overridden Chucker Channel",
                        NotificationManager.IMPORTANCE_MIN
                )
        )
```

vibin:
I didn't realise I could redefine the notification channel - will try that. Thanks a lot, @cortinico!

PR #182, opened 2020-01-10
Notifications update

vbuberen:
## :camera: Screenshots
Screen Shot 2020-01-10 at 17 45 39

## :page_facing_up: Context
Initially I planned to to just do some changes in order to get rid of notifications sound, but after looking through the code found some space for improvements, like flexibility in turning on/off notifications for transactions/errors and some code improvements.
Also, found out that icon for transactions notifications was the last raster drawable in the project and replaced it with vector.

## :pencil: Changes
- Created 2 separate notification channels for transactions and errors.
- Replaced `png` Chucker notification icon with `vector drawable`.
- Minor code cleanup.

## :crystal_ball: Next steps
- [x] Consider making notifications silent and without vibration, but be sure that they appear on devices with Android Q without problems, like in #144 (I believe it is caused by https://www.androidauthority.com/android-q-hidden-notifications-984317/ )


library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
```diff
@@ -17,10 +17,14 @@ import com.chuckerteam.chucker.internal.ui.BaseChuckerActivity
 import java.util.HashSet
 
 class NotificationHelper(val context: Context) {
+
     companion object {
-        private const val CHANNEL_ID = "chucker"
```
cortinico: Not a review, just a side note. Changing the Channel ID here will cause the fix we suggested here #154 to do not work anymore. Something that we want to point out somewhere.

As an alternative I suggest we keep the `TRANSACTIONS_CHANNEL_ID` as `"chucker"` as it's probably the most noisy.

vbuberen: I am not sure that this is kind of an issue which might required rolling back. 
It is better to have some meaningful names for all channels once we create them instead of having generic `chucker` for one and `chucker errors` for another.
Of course we will mention this change at least in #154

vbuberen:
@cortinico Wanted to hear your opinion on notifications importance. As I mentioned in the description of PR I wanted to get rid on notifications sound on Android 8 and higher (on lower versions there is just icon). But it seems that the only way is to set `IMPORTANCE_LOW` for a notification channel to get a silent push with an icon in the status bar. However, on Android Q such notifications will show no icon in the status bar due to https://www.androidauthority.com/android-q-hidden-notifications-984317/
So my question is - maybe it is acceptable to have no icon on Android Q, but forget about annoying multiple sounds for pushes on Android O and higher?

cortinico:
> So my question is - maybe it is acceptable to have no icon on Android Q, but 
> forget about annoying multiple sounds for pushes on Android O and higher?

If possible, I would avoid branching the notification importance based on the API level. I would keep the logic simple as stick to `IMPORTANCE_DEFAULT` for every channel. Maybe we can mention on the README how to override the notification importance as suggested in #154 

Moreover please note that the `ChuckerCollector` has the `showNotification` boolean property to hide/show notifications that should also cover the majority of the use case.

vbuberen:
Hmm, I created another emulator device with Android Q, but this time selected another image (the one with 29 on the attached screenshot) and on this image I see notification icons even for `IMPORTANCE_LOW`. Need to confirm on a real device.
Screen Shot 2020-01-12 at 22 23 01


> Moreover please note that the `ChuckerCollector` has the `showNotification` boolean property to hide/show notifications that should also cover the majority of the use case.
Probably, I wasn't clear on which icons I meant, so attaching a screenshot. The flag you mentioned controls whenever push will be fired. The issue I mentioned is that the notification is present in notification shade, but the icon is absent in the status bar (see the attached screenshot)
Screen Shot 2020-01-12 at 22 21 47

vbuberen:
@cortinico This PR is ready for another review.

PR #189, opened 2020-01-17
Headers redaction refactoring

vbuberen:
## :page_facing_up: Context
After merging #181 I decided to review some of the code, which works with headers and simplify it a little bit both in terms of usage and code quality.
During this work I found out that comparison of headers in `filterHeaders` was case-sensitive, while [the official HTTP standard](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) states that header field names are case-insensitive. So, it means that the library had a bug, which I also fixed here.

## :pencil: Changes
- Changed parameter type in `redactHeader()` function to accept `vararg` (note, that nothing changes for existing users).
- Replaced headers comparison with case-insensitive alternative
- Updated Readme and `ChuckerInterceptor` documentation to match changes.

## :warning: Breaking
No

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -41,6 +41,7 @@ class ChuckerInterceptor @JvmOverloads constructor(
     private val io: IOUtils = IOUtils(context)
     private val headersToRedact: MutableSet<String> = headersToRedact.toMutableSet()
 
+    /** Adds [headerName] into [headersToRedact] */
     fun redactHeader(vararg headerName: String) {
         headersToRedact.addAll(headerName)
     }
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -28,8 +28,8 @@ private const val MAX_BLOB_SIZE = 1000_000L
  * @param maxContentLength The maximum length for request and response content
  * before they are truncated. Warning: setting this value too high may cause unexpected
  * results.
- * @param headersToRedact List of headers that you want to redact. They will be not be shown in
- * the ChuckerUI but will be replaced with a `**`.
+ * @param headersToRedact Set of headers that you want to redact. They will be replaced
```
cortinico: `Set` is probably not the best word here. Ideally I would replace it with either:

- `Array` of headers... (as the `vararg` is compiled into a `String[]`)
- One or more headers...

vbuberen: Probably, I could use [Set] there, like I did in  #190, so it won't confuse anyone. What do you think?

cortinico: Yeah that makes sense 👍

vbuberen: Deal. Will do in the morning.

PR #190, opened 2020-01-18
ChuckerInterceptor refactoring

vbuberen:
## :page_facing_up: Context
During work on #189 saw that `intercept()` function is too hard to read and understand. This is the main reason of this PR. 

## :pencil: Changes
- Moved parts of code from `intercept()` into new functions responsible for request and response. Also cleaned up these code parts.
- Changed transaction time measurement method from manual to invocation of available fields in `Response` class.
- Updated documentation.

## :warning: Breaking
No


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -46,24 +44,47 @@ class ChuckerInterceptor @JvmOverloads constructor(
     }
 
     @Throws(IOException::class)
-    @Suppress("LongMethod")
```
cortinico: Thanks for making this world a better place :)

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -80,52 +101,40 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
             responseMessage = response.message()
 
             responseContentType = responseBody?.contentType()?.toString()
             responseContentLength = responseBody?.contentLength() ?: 0L
-            setResponseHeaders(filterHeaders(response.headers()))
-        }
 
-        val responseEncodingIsSupported = io.bodyHasSupportedEncoding(response.headers().get(CONTENT_ENCODING))
-        transaction.isResponseBodyPlainText = responseEncodingIsSupported
+            tookMs = (response.receivedResponseAtMillis() - response.sentRequestAtMillis())
```
cortinico: This is definitely a different metric than the one we used before. I don't think we will have major problem now, but it's probably worth to investigate if metrics are consistent with the calculation we used before.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -80,52 +101,40 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
-
-        collector.onResponseReceived(transaction)
-
-        return response
     }
 
     /**
-     * Private method to process the HTTP Response body and populate the corresponding response fields
-     * of a the [HttpTransaction].
+     * Processes a [ResponseBody] and populates corresponding fields of a [HttpTransaction].
```
cortinico: Since you're refactoring this class, maybe you can extract the `.contains("image")` and move the string literal to the `companion object`?

vbuberen: Sure, will do

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
vbuberen: My point was that this metric should be more accurate, since it is provided by OkHttp itself.

cortinico: Yup. I've done some testing and we should be fine with the new approach 👍

Issue #151, opened 2019-10-29
Response Search Improvements


kmayoral:
## :warning: Is your feature request related to a problem? Please describe.

One related problem is that when you perform a search in the response tab, any highlighted matches will be unhighlighted as you scroll through the body of the response. Furthermore, the act of searching does not auto scroll to the first occurrence of the search result as I would expect.

## :dart: Describe the solution you'd like

I propose the following:
- As you type a search query, the body should auto scroll to the first occurrence of the search term
- Proper highlighting be maintained for search results as you scroll through the body of the response
- Additional search controls be introduced to allow for navigating between next and previous search results with additional meta data displaying the current visited search result index and the total count of search results shown.

## :bar_chart: Describe alternatives you've considered

No other alternatives considered at this time

## :page_facing_up: Additional context

Thanks for a great product and for your consideration of this request. 

## :pencil: Do you want to develop this feature yourself?
- [ ] Yes
- [X] No

cortinico:
Sounds like a great idea :) This is open for development

vbuberen:
I would like to try to implement it in about 2 weeks when I have more time. 
If someone can do it earlier - please, leave a note here, so I will be aware of it.

vbuberen:
Also, I am not sure if this feature should go into `3.1.0` release, since it already has pretty serious amount of changes. @cortinico Agree?

cortinico:
Yeah, let's push it to 3.2.x or further.

vbuberen:
Moving further to be able to release `3.2.0`

vbuberen:
Update here - I am on it and going to present the solution pretty soon
Here is a sneak peak of how the UI would look like during the search (that view for navigating will also have number of occurrences and number of the currently selected one:
![Screenshot 2020-03-30 at 15 54 57](https://user-images.githubusercontent.com/13467769/77914612-dad4cd80-729e-11ea-8978-a00194190be1.png)

JFalky:
Hello! I'm really looking forward to this feature. How quickly can you complete it? =)

ArjanSM:
@vbuberen would you need any help finishing this feature?

vbuberen:
I am not working on it for quite a long time already as switched to Flutter development and lost motivation for it. 
So it is free to grab.

cortinico:
> @vbuberen may I be of any help for finishing this feature?

I've assigned this to you 👍

PR #195, opened 2020-01-21
Added ability to iterate through search results in response pane

kmayoral:
This uses the  IME search action button as the method for iterating through search resultt

## :camera: Screenshots
Video here:
iterate

## :page_facing_up: Context
I was looking for an easy way to iterate between multiple search results in long json responses viewed through Chucker.

## :pencil: Changes
The transaction adapter/fragment code. Added a few mutable variables to keep track of state, open to any suggestions for clean up, I was just getting something easy in for my purposes.

## :paperclip: Related PR
*PR that blocks this one, or the ones blocked by this PR*

## :warning: Breaking
None that I could spot, the search IME action button was previously unused.
## :pencil: How to test
Just follow the video example
## :crystal_ball: Next steps
None currently, thanks!

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
@@ -1,5 +1,6 @@
 package com.chuckerteam.chucker.internal.ui.transaction
 
+import android.content.Context
 import android.graphics.Bitmap
 import android.text.SpannableStringBuilder
 import android.text.Spanned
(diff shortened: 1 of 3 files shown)


vbuberen:
Hey @kmayoral 
Thanks for your contribution. 
However, in #151 you set a checkbox that you don't want to develop it. 
From my point of view it is now the best way to do the search an iterate through results.

I also started my implementation recently. But in my case I am doing like it is done in Chrome browser - with an additional view to show amount of results and buttons to navigate (will attach a screenshot below).
If you would like,  I would gladly hand over this feature to you, so you could implement it that way. If no, I am going to finish it pretty soon, but don't want to include into `3.1.0`, since we already have pretty serious amount of changes there.

vbuberen:
Here is what I meant when mentioned Chrome browser.
search_example

kmayoral:
Hi @vbuberen, I definitely prefer your approach as well, I was just moving with a quick solution to unblock my current use case. I'm happy to let you continue with your development on your solution and will close this PR for now. Thanks for this great product!

vbuberen:
Thanks for you feature suggestion :)

Issue #203, opened 2020-01-24
Version 3.1.0 always claims that response body is empty


Nekromancer:
## :loudspeaker: Describe the bug
After upgrading chucker to version 3.1.0 every single response shows 0 byte size and null body. In app everything works as expected. Downgrading to 3.0.1 fixes the issue

## :bomb: To Reproduce
Do some api request. Go to chucker and click on it. Request is properly shown, but response body is "null". Response size is always 0 bytes"

### :wrench: Expected behavior
Body should be always properly shown.

## :camera: Screenshots
![image](https://user-images.githubusercontent.com/3933348/73060655-9c82f280-3e98-11ea-960c-c250565c5e66.png)
![image](https://user-images.githubusercontent.com/3933348/73060689-b290b300-3e98-11ea-8cc7-63efa6cd32ab.png)

## :iphone: Smartphone
 - Device: Android Emulator
 - OS: Android 9
 - Lib version: 3.1.0

## :page_facing_up: Additional context
Whole request copied as text:
```
URL: <redacted>
Method: GET
Protocol: http/1.1
Status: Complete
Response: 200 OK
SSL: Yes

Request time: Fri Jan 24 10:58:58 GMT+01:00 2020
Response time: Fri Jan 24 10:58:58 GMT+01:00 2020
Duration: 67 ms

Request size: 0 B
Response size: 0 B
Total size: 0 B

---------- Request ----------

appInstallId: cc5a34e6-9ccf-46d4-9297-04f74a18355e
User-Agent: Android: 1.0.7-rc-dbg-dev (5) ; debug ; debuggableDev
Device-Name: AOSP On IA Emulator
HMAC: 6a2b11
Authorization: Bearer <token here>



---------- Response ----------

content-type: application/json; charset=utf-8
vary: Accept-Encoding
server: envoy
api-supported-versions: 1.0
date: Fri, 24 Jan 2020 09:58:56 GMT
x-envoy-upstream-service-time: 22

null
```

My chucker setup is fairly typical:
```kotlin
// setup collector
val chuckerCollector = ChuckerCollector(
            context = applicationContext,
            showNotification = false,
            retentionPeriod = RetentionManager.Period.ONE_DAY
        )

// adding interceptor
builder.addInterceptor(
            ChuckerInterceptor(
                context,
                collector = chuckerCollector
            )
        )
```

I have quite a few interceptors along the way. ChuckerInterceptor is almost last in the chain, after it there is only `RetryInterceptor`

[...]

PR #205, opened 2020-01-24
Fix/response body

vbuberen:
## :page_facing_up: Context
Fix for #203 
Now we are creating a complete response copy to pass further the chain of interceptors to be sure that Chucker doesn't break anything during processing.

## :pencil: Changes
- Removed redundant response size reassigning (to avoid 0 in size)
- Added `newBuilder()` for response to get a copy of original response.

## :warning: Breaking
Nope


vbuberen:
Closing in favor of #206

PR #198, opened 2020-01-21
Clone the Response Buffer rather than using it directly

cortinico:
##  :page_facing_up: Context

Fixing a potential `IllegalStateException: closed` if other OkHttp Interceptors are accessing the Response Buffer. The bug was introduced in #165. See the discussion in #192 for further context.

## :pencil: Changes
Let's close the buffer early upfront rather than opening/closing it directly.

## :paperclip: Related PR
Fixes #192

PR #206, opened 2020-01-24
Fix for Empty Response Body and 0 Content Lenght

### LINKED ISSUE #203: Version 3.1.0 always claims that response body is empty

cortinico:
Fixes a bug introduce by me in #198, when implementing clone of the `okio.Buffer` from the HTTP Response.

Looks like the problem on our end what that we were using `ChuckerInterceptor` as first interceptors, so the Sample App showed valid results. Switching order of the interceptors showed that responses are `null` and the approach followed for the `.clone()` was not correct. I reimplemented it following the `HttpLoggingInterceptor` body.

# How to test

Remove the fix I applied on `ChuckerInterceptor` and keep only the changes I applied in the sample app. It should break and show `null` in the response body.

Fixes #203
Closes #205

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -16,6 +16,7 @@ import okhttp3.Response
 import okhttp3.ResponseBody
 import okio.Buffer
 import okio.BufferedSource
+import okio.GzipSource
 
 /**
  * An OkHttp Interceptor which persists and displays HTTP activity
@@ -108,7 +109,7 @@ class ChuckerInterceptor @JvmOverloads constructor(
      * Processes a [Response] and populates corresponding fields of a [HttpTransaction].
      */
     private fun processResponse(response: Response, transaction: HttpTransaction) {
-        val responseBody = response.body()
+        val responseBody = response.body()!!
         val responseEncodingIsSupported = io.bodyHasSupportedEncoding(response.headers().get(CONTENT_ENCODING))
 
         transaction.apply {
@@ -123,8 +124,8 @@ class ChuckerInterceptor @JvmOverloads constructor(
             responseCode = response.code()
             responseMessage = response.message()
 
-            responseContentType = responseBody?.contentType()?.toString()
-            responseContentLength = responseBody?.contentLength() ?: 0L
+            responseContentType = responseBody.contentType()?.toString()
+            responseContentLength = responseBody.contentLength()
 
             tookMs = (response.receivedResponseAtMillis() - response.sentRequestAtMillis())
         }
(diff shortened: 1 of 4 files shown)


sample/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <debug-overrides>
+        <trust-anchors>
```
vbuberen: Why did you add this config into the sample?

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -137,32 +134,34 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
+    private fun processResponseBody(response: Response, responseBody: ResponseBody, transaction: HttpTransaction) {
+        val contentType = responseBody.contentType()
+        val charset: Charset = contentType?.charset(UTF8) ?: UTF8
+        val contentLength = responseBody.contentLength()
+
+        val source = responseBody.source()
+        source.request(Long.MAX_VALUE) // Buffer the entire body.
+        var buffer = source.buffer()
+
+        if (CONTENT_ENCODING_GZIP.equals(response.headers()[CONTENT_ENCODING], ignoreCase = true)) {
```
vbuberen: We already have this check in as `io.bodyIsGzipped()`. You could use it here.
However, I like you approach with constant. Could I ask you to modify `IOUtils` class as well to follow the same approach as here?

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -137,32 +134,34 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
+            GzipSource(buffer.clone()).use { gzippedResponseBody ->
```
vbuberen: Maybe we should move it into `IOUtils`? We have `getNativeSource()` there, which currently used only for request handling. I would move this code there, so both request and response use the same approach.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -137,32 +134,34 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
-            val buffer = source.buffer()
-            var charset: Charset = UTF8
-            val contentType = responseBody?.contentType()
-            if (contentType != null) {
-                try {
-                    charset = contentType.charset(UTF8) ?: UTF8
-                } catch (e: UnsupportedCharsetException) {
-                    return
-                }
+    private fun processResponseBody(response: Response, responseBody: ResponseBody, transaction: HttpTransaction) {
```
vbuberen: Since you started changes here, will drop a small concern.
I don't understand why can't we just pass only `response` here and retrieve `responseBody` inside this fun? 

Or another possible change - if we look for `response` usages we might see that is used only for headers check. So, maybe it is worth to pass just headers into this fun?

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -137,32 +134,34 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
+            if (contentLength != 0L) {
+                transaction.responseBody = buffer.clone().readString(charset)
+            }
+        } else {
+            transaction.isResponseBodyPlainText = false
 
-                if (isImageContentType && buffer.size() < MAX_BLOB_SIZE) {
-                    transaction.responseImageData = buffer.readByteArray()
-                }
+            val isImageContentType = (transaction.responseContentType?.contains(CONTENT_TYPE_IMAGE) == true)
```
vbuberen: Headers are case-insensitive. I know that it is hard to find an API, which sends `Image` or somehow else, except `image`, but let's follow the standard.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
cortinico: > I don't understand why can't we just pass only response here and retrieve responseBody inside this fun?

Yeah we could. Should we make this change in another PR?

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
vbuberen: Works for me 🆗

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
vbuberen: Yeah, but it is just one word to add here.
```suggestion
            val isImageContentType = (transaction.responseContentType?.contains(CONTENT_TYPE_IMAGE, ignoreCase = true) == true)
```

Issue #117, opened 2019-09-18
Better RTL support


eitanaviv:
**Is your feature request related to a problem? Please describe.**
It would be nicer if Chucker could have a better support for RTL devices.

**Describe the solution you'd like**
ViewPager works the opposite way on RTL layout and some texts are overlapping each other.

**Describe alternatives you've considered**
Better RTL support.

**Additional context**
![Screenshot_1568793405](https://user-images.githubusercontent.com/35567702/65129159-6eca0080-da03-11e9-94c5-56776ccd8739.png)
![Screenshot_1568793426](https://user-images.githubusercontent.com/35567702/65129161-7093c400-da03-11e9-87e9-2944fa251fd8.png)

**Do you want to develop this feature yourself?**
No unfortunately.

cortinico:
Thanks for the report @eitanaviv 

Having full support for RTL will require to update the `ViewPagers` and `TabLayouts` to handle both LTR and RTL.

To fix your situation, I've currently forced all the layouts to be LTR. This will prevent strings from overlapping and will make sure the ViewPagers are working in the right direction.
Once `ViewPager2` will be released and stable, we can consider providing full RTL support.

vbuberen:
Would like to reopen and assign myself to this issue. Going to update my #159 with RTL support, since forcing LTR isn't an option in long-time perspective. Targeting for `3.2.0` for release.

vbuberen:
There are some serious changes to be done in layouts, like `Overview` tab, so decided to make it in a separate branch to not mix with `ViewPager2` code updates.

vbuberen:
Hi @eitanaviv 
Am I right that you use RTL on a daily basis?
If so, I would like to ask you to help me a little bit by providing feedback on my implementation of RTL support.
You can try it out by adding a snapshot as dependency into your app:
`implementation 'com.github.ChuckerTeam:Chucker:feature~rtl_support-SNAPSHOT'`

Also, I would like to ask how the response body in `Response` tab should look like. Should it be mirrored as well?

vbuberen:
Closing this issue, since #208 is already merged. In case any troubles found with updated RTL implementation - open new issues.

PR #121, opened 2019-09-19
Force all layouts to be LTR

cortinico:
Developing full support for RTL would require to adapt all
the ViewPagers/TabLayouts to work correctly. I'm currently
fixing all the layouts to LTR and we'll wait for ViewPager2
to provide full RTL support.

| List | Transaction |
| -- | -- | 
| Screenshot_1568907113 | Screenshot_1568907116 |

Fixes #117

PR #208, opened 2020-01-25
RTL Support

### LINKED ISSUE #241: Text constraints issue in transactions list
In case of long transaction url when `url` textview becomes 2 lines high it overlaps with `path` textview.
See attached screenshot (had to cover some text due to privacy reasons).

## :wrench: Expected behavior
Texts don't overlap

## :camera: Screenshots
texts_overlap

vbuberen:
## :camera: Screenshots

| Before | After |
| --- | --- |
| Before | After |
transactions_list_rtl_before | transactions_list_rtl

## :page_facing_up: Context
PR addresses feature request #117 
Since Chucker currently has min API 16 as supported, while RTL properties are available from 17, I added all RTL related things into styles in `values-v21`, which means that RTL works only from Android 5.0 (API 21) and newer.

## :pencil: Changes
- Added tags to define text direction and alignment in `styles.xml`
- Replaced `TableLayout` with `ConstrainLayout` in `TransactionOverviewFragment` layout.
- Added `useVectorDrawables = true` in `build.gradle` to avoid PNG generation during builds and use real vectors.
- Replaced `android:src` with `app:srcCompat` to use vector drawable via AndroidX library.
- Fixed constraints to close #241.

## :pencil: How to test
Set `Force RTL` in Developer options or switch to some RTL locale.

## :crystal_ball: Next steps
Would like to hear some feedback from real RTL users, who could tell if I did everything right. Since I am not an RTL user, can't claim that it works as it should.


vbuberen:
This PR now closes #241 as well.

vbuberen:
Let's check and merge it. It seems we won't get feedback from the original reporter of this feature request.
Also, the PR has some additional fixes, like fix #241 

Finally, due to #227 I updated ids for layout items to be 100% sure we are not clashing with ids in users projects.

library/src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.chuckerteam.chucker">
-    <application>
+    <application
+      android:supportsRtl="true">
```
cortinico: This should be avoided. Ideally we should not set a `android:supportsRtl` in the library as this gets merged in the final manifest. The default value for `supportsRtl` is false.

Moreover we might break builds for users that have `supportsRtl` set to `false` on the host app. See also https://github.com/ChuckerTeam/chucker/pull/121#issuecomment-533480441

library/src/main/res/values-v21/styles.xml
```diff
@@ -1,5 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
+
+   <style name="Chucker.Theme" parent="Theme.MaterialComponents.DayNight.NoActionBar">
```
cortinico: Can I ask you to refactor to have a:

`<style name="Chucker.BaseTheme" parent="Theme.MaterialComponents.DayNight.NoActionBar">`

inside the `res/values/styles.xml` with all the common theme attributes?

We can then have the `Chucker.Theme` defined in both `values` and `values-v21` so we don't need to copy over the theme twice.

library/src/main/res/values-v21/styles.xml
```diff
@@ -9,4 +26,15 @@
...
+    <style name="Chucker.TextAppearance.Value" parent="android:TextAppearance.Small">
+        <item name="android:paddingStart">16dp</item>
+        <item name="android:textIsSelectable">true</item>
+    </style>
+
+    <style name="Chucker.TextAppearance.TransactionTitle" parent="TextAppearance.AppCompat.Widget.ActionBar.Title">
+        <item name="android:gravity">top</item>
+        <item name="android:textSize" tools:ignore="SpUsage">16dp</item>
+        <item name="android:maxLines">2</item>
+    </style>
```
cortinico: Why is this needed?

library/src/main/res/values-v21/styles.xml
vbuberen: Sure. I just thought that we might bump minSDK to 21 pretty soon due to Retrofit updates, etc.

library/src/main/res/values-v21/styles.xml
vbuberen: Just left it from original styles. It is used in `TransactionActivity` and `ErrorActivity` for toolbar title. Actually, I am against using dps for text sizes here. Will update as well.

library/src/main/res/values-v21/styles.xml
cortinico: Yeah true, but we don't have a PR pending for that (plus would be a major version bump). I think it's better if we keep the theme clean. Moreover having this setup with a `BaseTheme` could be handy for future similar scenarios

gradle.properties
```diff
@@ -16,6 +16,8 @@ org.gradle.jvmargs=-Xmx1536m
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 org.gradle.parallel=true
 
+android.useAndroidX=true
```
vbuberen: Added, because `3.6.0` plugin throws error without it.

vbuberen:
@cortinico It is updated and ready for a new round of review.

library/src/main/AndroidManifest.xml
```diff
@@ -8,9 +8,11 @@
             android:launchMode="singleTask"
             android:taskAffinity="com.chuckerteam.chucker.task"
             android:theme="@style/Chucker.Theme" />
+
         <activity android:name="com.chuckerteam.chucker.internal.ui.transaction.TransactionActivity"
             android:theme="@style/Chucker.Theme"
             android:parentActivityName="com.chuckerteam.chucker.internal.ui.MainActivity"/>
+
```
cortinico: nit: whitespace change here that can be reverted

vbuberen: I`ve added it for readability.

vbuberen:
Just tested it myself - there is an issue due to `supportVectorDrawables = true` - Chucker crashes on devices with API 16-19. Will fix it later today. Do not merge, please.

vbuberen:
Issues is fixed. In case someone interested in reasons here is the link to comment in the issue similar to what was seen: 
https://stackoverflow.com/a/38999661/7339798

PR #210, opened 2020-01-27
Fix for redundant toolbar collapsing

vbuberen:
## :camera: Screenshots
| Before | After |
| --- | --- |
|toolbar_before |toolbar_after |

## :page_facing_up: Context
Both in Chuck and Chucker I was always seeing the same bug where switching from `Overview` to `Response` tab made the toolbar collapse. Scrolling the content won't return the toolbar and user has to do a downwards scroll on `TabLayout` to bring the toolbar back. I believe that for this screen there is no need in toolbar hiding, so remove scroll flags from it.

## :pencil: Changes
- Removed scroll flags from `Toolbar` in `TransactionActivity`
- Fixed the position of `ProgressBar` in `Response` tab.

### DIFF
--- a/library/src/main/res/layout/chucker_activity_transaction.xml
+++ b/library/src/main/res/layout/chucker_activity_transaction.xml
@@ -33,13 +33,13 @@
             android:id="@+id/toolbar"
             android:layout_width="match_parent"
             android:layout_height="?attr/actionBarSize"
-            app:layout_scrollFlags="scroll|enterAlways|snap"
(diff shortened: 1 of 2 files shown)


vbuberen:
I see the same issue when trying to build locally.
Tried to find, but was only able to find this: https://github.com/Kotlin/dokka/issues/213#issuecomment-513522265
Since Travis and I use different connections for sure, it seems that is the issue with getting documentation from `developer.android.com`

PR #216, opened 2020-01-29
Resources renaming

vbuberen:
## :page_facing_up: Context
Chucker contains `chucker_` prefix for every resource name. I suspect that it is due to historical reasons, since Chuck also had prefix in resources. Author must have added it, so users won't mix it with resources in their app. But after recent PR #188 all Chucker resources are now private and won't appear in app code for users. Due to this fact I suggest to remove `chucker_` prefix everywhere for more concise and meaningful naming. 

## :pencil: Changes
Renamed resources


vbuberen:
You could check it in action to see if I missed anything. I tried to save prefixes in things like intent to open Chucker directly, notifications channels ids, etc. 
Might have missed something

Issue #217, opened 2020-01-30
No method found for ViewModel


Nekromancer:
## :loudspeaker: Describe the bug
Screen from Firebase crash report. This only occurs with `minifyEnabled true`. Is there some config missing in pro-guard setup?

![image](https://user-images.githubusercontent.com/3933348/73430736-14409980-433f-11ea-8f4a-1386501cb839.png)

## :bomb: To Reproduce
I got this crash report from UAT environment and i cannot reach author to find out reproduction path. I'm pretty sure that this happens when http request fails.

### :wrench: Expected behavior
It should not crash

## :iphone: Smartphone
 - Device: Galaxy Note10
 - OS: Android 10
 - Lib version: 3.1.1

vbuberen:
Thanks for your report.
It seems that we have issues with custom `TransactionViewModelFactory()` in our code.
I am going to handle it.

vbuberen:
Closing due to #219 being merged.

PR #219, opened 2020-01-30
Fix R8 minification crash on TransactionViewModel creation.

MiSikora:
## :page_facing_up: Context
https://github.com/ChuckerTeam/chucker/issues/217

## :pencil: Changes
`TransactionViewModelFactory` uses now ViewModel's constructor instead of obtaining it via reflection.

## :warning: Breaking
No. Only internal API changes.

## :pencil: How to test
Run a sample app that is minified with R8 in full mode. It will crash when one wants to open Chucker.

## :crystal_ball: Next steps
No.


vbuberen:
LGTM 🚢

Checked with both `proguard-android` and `proguard-android-optimize` set of rules and everything works like a charm.

Thanks for your contribution.

PR #220, opened 2020-01-30
ViewModel refactoring

vbuberen:
## :page_facing_up: Context
Initially planned as just an update for ViewModel dependency.
However, after taking a closer look found a few not so optimal implementation details in ViewModel, like exposing `MutableLiveData`, while it is better expose only `LiveData`, usage of `this` instead of `viewLifecycleOwner` in fragments, redundant extension function `observeOnce`.

My refactored approach exposes `LiveData` only. Also, due to switch to `Transformations` we have got some improvements. Earlier if user opened transaction, which is still in progress, the content won't appear till user close/open this transaction again. Now user can open the transaction in progress and the content will appear as soon as transaction finishes.

UPD: Occasionally forgot to switch to a new branch when created `MainViewModel` and transferred main screen fragments to get data from view model, so this PR also contains these changes as well.

## :pencil: Changes
- Updated `lifecycle` dependency to a newer version.
- Replaced deprecated artifact `lifecycle-extensions` with 2 new ones (for `ViewModel` and `LiveData`).
- Switched from deprecated `ViewModelProviders.of()` to `ViewModelProvider`.
- Removed extension function `observeOnce` in favor of standard `Transformations`.
- Switched from `MutableLiveData` to `LiveData` for exposing to views.
- Replaced `this` with `viewLifecycleOwner` in fragments.
- Tried to clean up dependencies list with meaningful blocks.
- Created new `MainViewModel` for the main screen.
- Started to retrieve data via view model in `TransactionListFragment` and `ErrorListFragment`.

## :pencil: How to test
No special testing required. For long taking transaction updated `drip....` transaction from sample app can be used.


library/src/main/java/com/chuckerteam/chucker/internal/ui/error/ErrorListFragment.kt
```diff
@@ -42,27 +45,29 @@ internal class ErrorListFragment : Fragment() {
...
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        viewModel.errors.observe(
+            viewLifecycleOwner,
+            Observer { errors ->
+                adapter.setData(errors)
+                tutorialView.visibility = if (errors.isNullOrEmpty()
+                ) {
```
cortinico: Please don't go to the new line here.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    val errors: LiveData<List<RecordedThrowableTuple>> =
+        Transformations.map(
+            RepositoryProvider.throwable().getSortedThrowablesTuples()
+        ) {
+            it
+        }
+
+    fun getFilteredItems(searchQuery: String) {
```
cortinico: Please rename to something like `refreshFilteredItems` or `updateFilteredItems`. Having `get` in the method name is misleading as this method has not return value.

PR #221, opened 2020-02-02
Fix for crash on Save transaction action

vbuberen:
## :page_facing_up: Context
Found an error, which caused crash during `Save` action. If user opened a long transaction (which isn't finished at the moment when user clicks on it and tried to click `Save` in `Response` tab Chucker crashes on saving operation. The error was due to wrong condition set in `when` - since `transaction.type != TYPE_REQUEST` and `transaction.responseBody` was `null` we ended up in `else` branch of `when`, which tried to get `transaction.responseImageData` and received NPE.
Finally, all 3 conditions worked wrongly also due to safe calls for fields without returning `false` as an AsyncTask result which caused cases with saving of invalid data (without response), but still showing `File saved successfully`, which it incorrect. Fixed it as well by returning false in case any required field can't be obtained.

Also, with such long transactions, especially after merging #220, users could click on `Share` as text or curl for transactions in progress and share text without response data (in case of sharing as text), which makes this share invalid. To avoid such things added guarding and toast to inform user that transaction isn't ready.

## :pencil: Changes
- Fixed `when` condition in file saving action.
- Added toasts to inform user that the transaction isn't ready for share.

## :pencil: How to test
To test `Save` crash transaction like `drip..` from the sample app can be used. Just add some delay in it, open it when it is still in progress and try to click `Save` in `Response` tab.


library/src/main/res/values/strings.xml
```diff
@@ -47,4 +47,5 @@
     <string name="check_readme"><a href="https://github.com/ChuckerTeam/chucker/blob/develop/README.md#configure-">Check the setup instructions on GitHub</a></string>
     <string name="setup">Setup</string>
     <string name="binary_data">binary data</string>
+    <string name="request_not_ready">The request isn`t ready for sharing or saving</string>
```
cortinico:
```suggestion
    <string name="request_not_ready">The request isn't ready for sharing or saving</string>
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -251,15 +251,19 @@ internal class TransactionPayloadFragment :
                 val context = fragment.context ?: return false
                 context.contentResolver.openFileDescriptor(uri, "w")?.use {
                     FileOutputStream(it.fileDescriptor).use { fos ->
-                        when {
-                            type == TYPE_REQUEST -> {
-                                transaction.requestBody?.byteInputStream()?.copyTo(fos)
+                        when (type) {
+                            TYPE_REQUEST -> {
+                                transaction.requestBody?.byteInputStream()?.copyTo(fos) ?: throw IOException()
```
cortinico: (commenting once but you have 3 occurrences of this).

Mayne we can add a short message in the `IOException()`?

vbuberen: Good idea.

Issue #168, opened 2019-12-30
Null Pointer Exception


RazibKani:
## :loudspeaker: Describe the bug
*App crash because null exception*

## :bomb: Stacktrace
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.chuckerteam.chucker.api.internal.data.entity.HttpTransaction.getNotificationText()' on a null object reference
       at com.chuckerteam.chucker.api.internal.support.NotificationHelper.show + 96(NotificationHelper.java:96)
       at com.chuckerteam.chucker.api.ChuckerCollector.onResponseReceived$library_release + 66(ChuckerCollector.kt:66)
       at com.chuckerteam.chucker.api.ChuckerInterceptor.intercept + 138(ChuckerInterceptor.kt:138)
       at okhttp3.internal.http.RealInterceptorChain.proceed + 112(RealInterceptorChain.kt:112)
       at okhttp3.internal.http.RealInterceptorChain.proceed + 87(RealInterceptorChain.kt:87)
       at okhttp3.logging.HttpLoggingInterceptor.intercept + 215(HttpLoggingInterceptor.kt:215)
       at okhttp3.internal.http.RealInterceptorChain.proceed + 112(RealInterceptorChain.kt:112)
       at okhttp3.internal.http.RealInterceptorChain.proceed + 87(RealInterceptorChain.kt:87)
       at com.happyfresh.snowflakes.hoverfly.interceptors.ClientInterceptor.intercept + 60(ClientInterceptor.kt:60)
       at okhttp3.internal.http.RealInterceptorChain.proceed + 112(RealInterceptorChain.kt:112)
       at okhttp3.internal.http.RealInterceptorChain.proceed + 87(RealInterceptorChain.kt:87)
       at okhttp3.RealCall.getResponseWithInterceptorChain + 184(RealCall.kt:184)
       at okhttp3.RealCall$AsyncCall.run + 136(RealCall.kt:136)
       at java.util.concurrent.ThreadPoolExecutor.runWorker + 1167(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run + 641(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run + 764(Thread.java:764)

## :iphone: Smartphone
 - Device: *Nokia 6.1*
 - OS: *9*
 - Lib version: *3.0.1*

vbuberen:
[...]

PR #223, opened 2020-02-02
Fix for NPE in NotificationHelper

vbuberen:
## :page_facing_up: Context
Fix, which should protect from issues, like #168 in `NotificationHelper`. Checks for `transaction` in `transactionBuffer` being `null` should protect from cases described in the mentioned issue.

## :pencil: Changes
Add explicit check for `null` value in `transactionBuffer`.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
@@ -107,7 +107,7 @@ internal class NotificationHelper(val context: Context) {
             synchronized(transactionBuffer) {
                 var count = 0
                 (transactionBuffer.size() - 1 downTo 0).forEach { i ->
-                    if (count < BUFFER_SIZE) {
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
```diff
@@ -107,7 +107,7 @@ internal class NotificationHelper(val context: Context) {
             synchronized(transactionBuffer) {
                 var count = 0
                 (transactionBuffer.size() - 1 downTo 0).forEach { i ->
```
cortinico: How about something like:

```kotlin
(transactionBuffer.size() - 1 downTo 0).forEach { i ->
                    val item = transactionBuffer.valueAt(i)
                    if (item != null && count < BUFFER_SIZE) {
                        if (count == 0) {
                            builder.setContentText(item.notificationText)
                        }
                        inboxStyle.addLine(item.notificationText)
                    }
                    count++
                }
```

PR #232, opened 2020-02-09
Add missing  section to release 3.1.1 and 3.1.2

cortinico:
Looks like we haven't credit the last two messages, let's do it 👍


vbuberen:
Yeah, missed it. Thanks for catching it up.

PR #233, opened 2020-02-09
Add switching between encoded and decoded URL formats.

### LINKED ISSUE #155: URL Encoder/Decoder
Please do provide url encode and decode for api request

MiSikora:
## :camera: Screenshots
*Show us what you've changed, we love images.*

Screenshot_1581281305
Screenshot_1581280198
Screenshot_1581280204
Screenshot_1581280210

## :page_facing_up: Context

There is an issue - https://github.com/ChuckerTeam/chucker/issues/155

## :pencil: Changes

From the exsiting code this PR touches mostly ViewModels and Entities. I added a way to display encoded or decoded URL (or its parts) with a wrapper `FormattedUrl` class.

## :warning: Breaking

No.

## :pencil: How to test

Run the sample app. Use checkbox in the top right corner to switch between encoded and decoded URLs. The option should persist between application restarts.

## :crystal_ball: Next steps

No.

Closes #155


vbuberen:
Thanks for your contribution.

I haven't checked the code yet, but would like to provide some feedback on UI.
For me it looks not good from UX point of view to have a checkbox in a dropdown. 
I would switch to some icon with 2 states (off/on) just inside toolbar to reduce number of actions required for switching.
We will have space there especially after my future planned update where share/save operations will move into FAB.

As to code review - will do in nearest days, since have quite limited availability at the moment.

P.S. Big thumbs up for tests)) Soon we will have coverage reports here in Chucker with requirements of some specific coverage level, so you are helping a lot.

vbuberen:
Side note: welcome to our #chucker channel in Kotlinlang Slack. We discuss some plans and ideas there as well as share some things to find out if it is worth implementation. 
We could discuss UX topic of your changes there if needed, for example.

MiSikora:
Ok, thanks for the feedback. I wanted to use a toggle icon button but couldn't figure out how to represent HTTP encoding with an image. Maybe I'll come up with something.

Also, I'm thinking now it makes sense to make this option configurable from code. Users will probably expect to have a default option on app reinstallation as well. Since this is a QA plugin the apps will be reinstalled all the time. Not sure what the place is for that. `Chucker` with a static function? Some option on `ChukcerInterceptor`? I'll have to update `README.md`.

vbuberen:
Yes, it makes sense to provide users with option to configure default preference while setting up Chucker, like they can with retention period.

vbuberen:
> Ok, thanks for the feedback. I wanted to use a toggle icon button but couldn't figure out how to represent HTTP encoding with an image. Maybe I'll come up with something.
I will look for some icons as well and share some ideas as well to not be just critic)))

cortinico:
> Also, I'm thinking now it makes sense to make this option configurable from code. Users will probably expect to have a default option on app reinstallation as well. Since this is a QA plugin the apps will be reinstalled all the time. Not sure what is the best place for that. `Chucker` with a static function? Some option on `ChukcerInterceptor`? I'll have to update `README.md`.

I'd rather avoid this approach. Offering this as a configurable feature has the side effect of extending the API surface for a limited benefit in terms of features.

I think we can just offer some UI controls (a button, checkbox, radio button, menu entry, etc...) to toggle the visibility of encoded/decoded URL.

vbuberen:
What about saving state of selection (show encoded or decoded URLs)? Will we save the state or just make users switch after every reinstall? 
Also, do you see it as a switch for every individual request or general in list of items?

MiSikora:
State is already saved to preferences on switching it. The main issue I see users have to switch it every time whenever app is reinstalled. Which I assume would be often given the nature of the library.

vbuberen:
Sorry @MiSikora, I addressed this question to @cortinico. I also referred case about reinstalls, which happen quite a lot during development.

cortinico:
> What about saving state of selection (show encoded or decoded URLs)? Will we save the state or just make users switch after every reinstall?

👍 for this approach. We can definitely use `SharedPreferences` or a table in the Room Database to store key value pairs with the current state.

> Also, do you see it as a switch for every individual request or general in list of items?

If I would have to develop this feature just for my self I'll do it stateless and only in the "detail" view (e.g. you click on the encoded URL and you magically decode it). I can't think of any meaningful use case that would push me to develop this feature for the list view as well (at least in my experience). I can anyway understand how this can be helpful for some so it's probably worth to investigate/consider.

Please also note that this has implications on how we search through the list (like if you decode URLs in, you probably want to offer search through the decoded URLs as well?)

vbuberen:
@MiSikora Sorry for not checking your code. Would like to come to some agreement regarding this feature first to not make you and me do some back-and-forth check/fix steps to not spend even more time. I will check as soon as we get the final vision of this feature flow.

@cortinico Makes sense. I agree that I would prefer to have this so-called switch `encode/decode` in each individual transaction, without saving states between sessions at all, since also can't remember scenario where I needed to decode all urls at once.
But I don't want to push too hard here, since we might be breaking the rule **`You are not your user`**.

Probably, we should do some sort of survey somewhere and get some meaningful responses.

cortinico:
> without saving states between sessions at all, since also can't remember scenario where I needed to decode all urls at once.

Exactly. That's my strong point. I'd try to make this stateless to avoid complication and just provide a simple way to decode a URL, nothing more.

vbuberen:
Ok, in such case let's skip the idea of doing survey etc. to not waste too much time.
@MiSikora could I ask you to update your code in a way that we will just have an options menu item in transaction details, which will switch encoded/decoded view without any state savings in SharedPreferences or db?

vbuberen:
Also, I promised to provide icon options for toggle.
My suggestion is to use eye icons:
https://materialdesignicons.com/icon/eye
https://materialdesignicons.com/icon/eye-off

The meaning behind this icon something like `convert item into human readable text` or not.

Having no state might be fine and we will definitely look for the feedback from users and change the way this feature works depending on it.

MiSikora:
Hey, sorry for being silent. I'm a bit occupied lately. I will update the code to reflect discussion some time soon

vbuberen:
@MiSikora Sure, take your time. We already appreciate you doing this PR and absolutely no pressure :)

MiSikora:
I adapted the code. One thing that was not clear to me was `HttpTransaction` handling from the database. In the title `LiveData` it was assumed that the value might be `null` but in the transaction `LiveData` it was not. I changed signature to expliclitly say that the value might be `null` in the database and handled it in the code.

Below are the screenshots with the redesigned approach. By default URLs are not encoded.

Screenshot_1581764109
Screenshot_1581764114

vbuberen:
Tried to run the sample app from your repo and wasn't able to see mentioned switch between encoding/decoding. Probably, I did something wrong?
Here is the sequence of actions:
1. Added ` redirectTo("https://ascii.cl?parameter=%22Click+on+%27URL+Encode%27%21%22").enqueue(cb)` into `HttpBinClient`.
2. Opened this transaction in the sample app and saw that it is already decoded and clicks on eye icon do nothing.
Here is how it looks like.
chucker_encoding

vbuberen:
Also, after trying it out myself it seems for me that we might improve the UX. Would be awesome if we could detect if url is encoded and show the eye icon only in such cases. For decoded or `normal` urls, like we have in our sample app show no `eye` icon, since I see no reason for willing to encode the url, which wasn't encoded originally.

vbuberen:
And it would be awesome if you add this url for check encode/decode functionality into the sample app.

MiSikora:
Wow, I'm sorry about that. It doesn't seems to work for me now as well. I must have messed up something with pushing and reverting commits. I'll double check my local commits and also I'll apply your suggestion about not showing eye icon for URLs which do not require decoding.

vbuberen:
No worries here. You are doing a great amount of work for Chucker and it is fine that there might be some typos.

MiSikora:
Ok, I rebased to accomodate for the `LiveDataRecord` in tests and fixed URL not displaying in encoded form. I also added mentioned suggestion about showin the icon.

vbuberen:
Yeah, can confirm that everything works.

Wouldn't you mind adding `redirectTo("https://ascii.cl?parameter=%22Click+on+%27URL+Encode%27%21%22").enqueue(cb)` into `HttpBinClient` in Chucker sample app so we could easily test this functionality on a regular basis?

Also found out that since the eye icon is available only in `Overview` after switching state user will see encoded/decoded url on all 3 tabs. It is not a big deal andI believe it is not worth it to add some additional logic for handling this, but just be aware.

MiSikora:
Ok, I added the samples.

Not switching back on other tabs is intentional as there is share functionality which also uses encoding/decoding based on the users' preference. Should this feature not affect sharing?

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -208,13 +208,27 @@ internal class HttpTransaction(
     }
 
     fun populateUrl(url: HttpUrl): HttpTransaction {
-        this.url = url.toString()
-        host = url.host()
-        path = ("/${url.pathSegments().joinToString("/")}${url.query()?.let { "?$it" } ?: ""}")
-        scheme = url.scheme()
+        val formattedUrl = FormattedUrl.fromHttpUrl(url, encoded = false)
+        this.url = formattedUrl.url
```
vbuberen: Maybe we could rename `url` parameter for this function, so there will be no need in using `this.url` for this line?

MiSikora: How about `httpUrl`?

vbuberen: Works for me 👍

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
```diff
@@ -1,28 +1,54 @@
...
-        Transformations.map(RepositoryProvider.transaction().getTransaction(transactionId)) {
-            if (it != null) "${it.method} ${it.path}" else ""
+    private val mutableEncodeUrl = MutableLiveData<Boolean>(false)
+
+    val encodeUrl: LiveData<Boolean> = mutableEncodeUrl
+
+    val transactionTitle: LiveData<String> = RepositoryProvider.transaction()
+        .getTransaction(transactionId)
+        .combineLatest(encodeUrl) { transaction, encodeUrl ->
+            if (transaction != null) "${transaction.method} ${transaction.getFormattedPath(encode = encodeUrl)}" else ""
```
vbuberen: This line seems too long.
```suggestion
            if (transaction != null) {
                "${transaction.method} ${transaction.getFormattedPath(encode = encodeUrl)}"
            } else {
                ""
            }
```

vbuberen:
Ok, I finished code review. Everything seems fine and works like it should.

vbuberen:
The only request for you - in the following PRs could you, please wait for the whole review to pass, so the reviewer could leave all his comments. I think it is not that convenient that you do multiple commits with single line changes as well as for reviewer there is also need to refresh PR after every such commit.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
MiSikora: It looks long to me as well but it does not break Detekt rules. I can change it but I would rather rely on static analysis for such stuff. Otherwise what's the point of it?

MiSikora:
Sure. But wouldn't it be easier for both sides to leave all comments at once instead of one by one? There is such a feature in GH while doing PR, and this way I know when the code reviewed as a whole. :)

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
vbuberen: Ok, as you wish. I will update rules for static analysis soon.

vbuberen:
I left comments one by one during one review. It took that much time, because I also checked some things on a device etc. 
I would leave the final review anyway after my whole review.

MiSikora:
Ok, makes sense. I'll try to be less eager with fixes in the future. :)

vbuberen:
I would use one of the following options anyway, so you will see everything and it won't be `one by one` as you say.
Screen Shot 2020-02-23 at 12 18 49

vbuberen:
Thanks for your time and for implementation of this feature request 👍 ❤️

vbuberen:
And yeah, I won't use `Add a single comment` button in future as well. Thanks for pointing out.

Issue #209, opened 2020-01-26
Payload size isn't correct for some transactions


vbuberen:
## :loudspeaker: Describe the bug
Some transactions show payload size as `-1`. It is totally fine, since OkHttp returns this value in case the size is unknown. However, when I wanted to handle this value in Chucker UI I decided to check what other network interceptors show and found out that [Stetho](https://facebook.github.io/stetho/) show sizes for transactions, which show `-1` in Chucker.

### :wrench: Expected behavior
Transactions show actual size instead of `-1`.

## :camera: Screenshots
<img width="920" alt="Screen Shot 2020-01-26 at 10 51 39" src="https://user-images.githubusercontent.com/13467769/73132894-f830b580-4029-11ea-96bd-c0c06151244e.png">

vbuberen:
After some basic investigation found out that Chucker solely relies on `contentLength` value in response to show the size, while Stetho does some calculations itself in case response headers have no `contentLength` value. Requests from the screenshot in my bug report are exactly such items with no mentioned value in their response.

So Chucker needs to implement the same mechanism.

MiSikora:
This could be fixed after merging #267. We would have to keep counting the bytes in `TeeSource` and report it back in the callback.

MiSikora:
I was thinking about this feature lately and see couple of things that could be done.

* If `Content-Length` is available Chucker could just use this value to determine the size of the response.
* If `Content-Length` is not available Chucker could wait for the response to be read in a `TeeSource` and be informed about bytes read once upstream is depleted. The problem occurs if the consumer does not read bytes fully like reported in #287. In this case there is no way of knowing what would be the size of the response.
* Chucker could display amount of data that was downloaded by the user and the size of the response if it is available (either form `Content-Length` or from `TeeSource` if exhausted). It would require some UI changes on most of the screens.
* Chucker could show gradual changes of downloaded data. So while bytes are being read from an upstream it would display `8 kB`, `16 kB`, `24 kB` and so on. Not really sure if there is a need for this. Also it would require UI changes mentioned in the previous point.

vbuberen:
Moving further, so we could release `3.2.0`

@MiSikora as to your ideas - agree on first 3 points, but the last one with progressive update of the request size is an overengineering, which won't bring much value to our users.

MiSikora:
Yes, I think so as well. Just wanted to propose it in case you find it valuable.

cortinico:
[...]

PR #234, opened 2020-02-09
Fix an issue with negative request/response sizes displayed.

MiSikora:
## :camera: Screenshots
*Show us what you've changed, we love images.*

Screenshot_1581268679
Screenshot_1581269196

## :page_facing_up: Context

I introduced a bug with one of the previous PRs.

## :pencil: Changes

I modified the response/request length on the UI level to display at lest `0` as a minimum value.

## :warning: Breaking

No.

## :pencil: How to test

Run the sample app. Requests and responses should display `0 B` instead of `-1 B` for some of the responses.

## :crystal_ball: Next steps

No.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
@@ -196,7 +196,7 @@ internal class HttpTransaction(
     }
 
     private fun formatBytes(bytes: Long): String {
-        return FormatUtils.formatByteCount(bytes, true)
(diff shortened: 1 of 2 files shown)


vbuberen:
Hi again @MiSikora.
It wasn't your bug that you see `-1` - this value is returned when `contentLength` size is unknown (API returned nothing in for this header).
You can see more info here: https://github.com/ChuckerTeam/chucker/issues/209
Till the time we introduce our own counting mechanism of size calculation, I would leave `-1`, since `0` isn't technically correct.

MiSikora:
Oh, ok. I thought I messed something up. I'll close this then. Thanks for the info.

PR #240, opened 2020-02-13
Fix HttpHeader serialization when obfuscation is enabled

technoir42:
## :page_facing_up: Context
I get this exception in `HttpTransaction.setRequestHeaders` when DexGuard is enabled:
```
java.lang.IllegalArgumentException: class o.bwy declares multiple JSON fields named ι
```

Mapping:
```
com.chuckerteam.chucker.api.internal.data.entity.HttpHeader -> o.bwy:
    java.lang.String name -> ι
    java.lang.String value -> Ι
```

## :pencil: Changes
Annotated fields in `HttpHeader` with `@SerializedName` so that Gson would know how to serialize that class when it's obfuscated.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpHeader.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpHeader.kt
@@ -1,3 +1,8 @@
 package com.chuckerteam.chucker.internal.data.entity
 
-internal data class HttpHeader(val name: String, val value: String)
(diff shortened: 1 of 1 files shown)


vbuberen:
I believe that we should consider annotating not this model only.

vbuberen:
Nevermind - other entities are annotated in Room with `columnName`, so there should be no problems.

cortinico:
> Nevermind - other entities are annotated in Room with `columnName`, so there should be no problems.

Also `HttpHeader` is the only class we actually serialize using GSON

PR #245, opened 2020-02-19
Add Action to validate the Gradle Wrapper

cortinico:
Let's add a simple GH action to validate the Gradle Wrapper just to play on the safe side.


vbuberen:
Awesome action. LGTM 🚢

PR #246, opened 2020-02-20
Do not refresh transaction when it is not being affected.

### LINKED ISSUE #244: Transaction data is updated constantly.
When user stays on some opened transaction and an app continues network activity in the background the whole transaction data on all 3 transaction tabs updates on each new transaction.
As a result, not only Chucker does redundant work, but also some flashing happens.

## :bomb: Steps to reproduce

1. Open any transaction details while app continues firing new requests in the background.
2. Observe screen flashing on each new transaction.

## :wrench: Expected behavior
Transaction (I guess the same thing happens with `Throwable`) is retrieved and filled only once when user on any of transaction details tabs.

## :camera: Screenshots
Cropped the gif due to privacy reasons:
chucker_live_data_issue_cropped

## :iphone: Tech info
 - Chucker version: 3.1.2

MiSikora:
## :page_facing_up: Context
There is an open issue. #244

## :pencil: Changes
DB queries which observe entities by their IDs now check beforehand is something changed before emitting data downstream.

This cannot be done on DB level as SQL emissions happen whenever table is changed (https://sqlite.org/lang_createtrigger.html).

One unpleasant change is custom `distinctUntilChanged()` `LiveData` extensions that accepts an executor on which it performs computation. It is needed as in our case data that is being compared can have long comparison computation (requests and responses), which would result in dropped frames.

Ideally `Flow` or Rx would be used to avoid this `LiveData` restriction.

## :paperclip: Related PR
Not blocking but there will be conflicts with https://github.com/ChuckerTeam/chucker/pull/233 due to `LiveData` extensions. I would suggest merging this one first becuase I introduced here a test wrapper for `LivedData`, that can be used in #233 as well.

## :no_entry_sign: Breaking
No.

Perform requests that are delayed so you can observe the effects of #244. Best way would be to either enqueue bunch of delayed requests or add a delay interceptor to the sample client. While requests are running open any transaction and observe it. Its tabs should not flash.

## :stopwatch: Next steps
No.

Closes #244.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
@@ -245,6 +245,6 @@ internal class HttpTransaction(
             (responseHeaders == other.responseHeaders) &&
             (responseBody == other.responseBody) &&
             (isResponseBodyPlainText == other.isResponseBodyPlainText) &&
-            responseImageData?.contentEquals(other.responseImageData ?: byteArrayOf()) == true
+            responseImageData?.contentEquals(other.responseImageData ?: byteArrayOf()) != false
     }
 }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
@@ -214,4 +214,43 @@ internal class HttpTransaction(
         scheme = url.scheme()
         return this
     }
+
+    // Not relying on 'equals' because comparison be long due to request and response sizes
+    // and it would be unwise to do this every time 'equals' is called.
+    @Suppress("ComplexMethod")
+    fun hasTheSameContent(other: HttpTransaction?): Boolean {
+        if (this === other) return true
+        if (other == null) return false
+
+        if (id != other.id) return false
+        if (requestDate != other.requestDate) return false
+        if (responseDate != other.responseDate) return false
+        if (tookMs != other.tookMs) return false
+        if (protocol != other.protocol) return false
+        if (method != other.method) return false
+        if (url != other.url) return false
+        if (host != other.host) return false
+        if (path != other.path) return false
+        if (scheme != other.scheme) return false
+        if (requestContentLength != other.requestContentLength) return false
+        if (requestContentType != other.requestContentType) return false
+        if (requestHeaders != other.requestHeaders) return false
+        if (requestBody != other.requestBody) return false
+        if (isRequestBodyPlainText != other.isRequestBodyPlainText) return false
+        if (responseCode != other.responseCode) return false
+        if (responseMessage != other.responseMessage) return false
+        if (error != other.error) return false
+        if (responseContentLength != other.responseContentLength) return false
+        if (responseContentType != other.responseContentType) return false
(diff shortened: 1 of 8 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -214,4 +214,37 @@ internal class HttpTransaction(
...
     }
+
+    // Not relying on 'equals' because comparison be long due to request and response sizes
+    // and it would be unwise to do this every time 'equals' is called.
+    @Suppress("ComplexMethod")
+    fun hasTheSameContent(other: HttpTransaction?): Boolean {
+        if (this === other) return true
+        if (other == null) return false
+
+        return id == other.id &&
```
vbuberen:
```suggestion
        return (id == other.id) &&
```

vbuberen: Could you, please, add braces for every comparison to make it more readable?

MiSikora: Sure.

vbuberen:
It seems that it doesn't fix the issue.
Here is how I tested:
1. Modified `HttpBinClient` in `sample` the following way
Screen Shot 2020-02-22 at 19 26 44
2. Opened one of already passed transactions.

Here is what I observe:
chucker_transaction_refresh

vbuberen:
I just cloned your repo and tried to run Chucker from there from `transaction-refresh` branch

MiSikora:
Oh, that's not good. I'll check it out. Thanks for the feedback.

MiSikora:
Yeah, ok. So I did a dumb nullable `Boolean` comparison check. Should be correct now.

vbuberen:
Yes, now works like a charm.

vbuberen:
LGTM 🚢 
Can I merge it? No other changes expected here?

MiSikora:
Yes, this one can be merged. I'll update encoding stuff with `LiveData` test style from here.

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
cortinico: Should I enable a Deketk/ktLint rule for this?

MiSikora: @cortinico I think it would be helpful if that's the way to go with conditions. One other thing I'd like to see is added `codeStyleConfig` to repository becuase sometimes I have to adjust stuff like line length or indentation manually as IDE does not format it for me properly and it is only caught by Detekt.

MiSikora: I can create a separate issue for this. Or not if you do not want `.idea` stuff in repository. :)

vbuberen: Good idea, actually. Also struggling with this styling issue all the time, since got used to use auto format feature a lot.

vbuberen: Creating an issue, so we could track it.

PR #252, opened 2020-02-26
Feature/tls info

### LINKED ISSUE #236: Add more info about TLS connection in transaction overview
Currently Chucker shows only `Yes` or `No` in SSL field of transaction overview. We could show more info, like TLS protocol version, cipher suites and even some certificates info.

At initial stage I would love to see TLS version and cipher suite in `Overview` tab. 

- [x] Yes
- [ ] No

vbuberen:
## :camera: Screenshots
Screen Shot 2020-02-26 at 12 11 54

## :page_facing_up: Context
This update closes #236 

## :pencil: Changes
Updated UI and DB model to show TLS connection info.
For non SSL connections newly added fields are hidden.

## :no_entry_sign: Breaking
Users will have to do a clean install of Chucker due to DB model update. Since it is a debug tool, don't see the reason in writing migrations.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionOverviewFragment.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionOverviewFragment.kt
@@ -7,6 +7,7 @@ import android.view.MenuInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.TextView
+import androidx.constraintlayout.widget.Group
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.Observer
 import androidx.lifecycle.ViewModelProvider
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionOverviewFragment.kt
```diff
@@ -80,6 +89,13 @@ internal class TransactionOverviewFragment : Fragment() {
                     status.text = transaction?.status?.toString()
                     response.text = transaction?.responseSummaryText
                     ssl.setText(if (transaction?.isSsl == true) R.string.chucker_yes else R.string.chucker_no)
```
cortinico: Can you move this also inside the `if` branch?

cortinico:
Other than this, the feature works fine 👍 Thanks for doing so!

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionOverviewFragment.kt
vbuberen: 👌

vbuberen:
> At least we should wipe the DB or provide some fallback. Eventually providing a migration is not a big deal (especially for this simple scenario).

We have ` fallbackToDestructiveMigration()` in DB declaration, so updating schema number should do the trick with clearing the DB (actually, already tested and it worked). Suggest to use this approach, since I really don't see reasons for storing such data as Chucker has for a long time.

cortinico:
> so updating schema number should do the trick with clearing the DB (actually, already tested and it worked)

Yes that should suffice.

vbuberen:
So, we are good to go? Or you have some additional feedback here?

PR #253, opened 2020-02-27
Switch views to ViewBinding

vbuberen:
## :page_facing_up: Context
Chucker relied on `findViewById` stuff for quite a long time. Since [ViewBinding](https://developer.android.com/topic/libraries/view-binding) is officially released, I believe we should switch to it to make library code more concise and robust.

## :pencil: Changes
- Enabled `ViewBinding` in `build.gradle`.
- Switched all Activities, Fragments and Adapters to use `ViewBinding`.
- Removed some boilerplate, like duplicate of `onTransactionItemClickListener()`.
- Fixed link to tutorial, which became invalid after recent resources renaming.

## :paperclip: Related PR
- [x] Would like to merge it after #252 so I could add new fields here as well.
- [ ] Need to merge #257 since there were some resources renaming.

Just run it.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/MainActivity.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/MainActivity.kt
@@ -5,16 +5,10 @@ import android.os.Bundle
 import androidx.lifecycle.ViewModelProvider
 import com.chuckerteam.chucker.api.Chucker
 import com.chuckerteam.chucker.databinding.ChuckerActivityMainBinding
-import com.chuckerteam.chucker.internal.ui.error.ErrorActivity
-import com.chuckerteam.chucker.internal.ui.error.ErrorAdapter
-import com.chuckerteam.chucker.internal.ui.transaction.TransactionActivity
-import com.chuckerteam.chucker.internal.ui.transaction.TransactionAdapter
 import com.google.android.material.tabs.TabLayout
 
 internal class MainActivity :
-    BaseChuckerActivity(),
-    TransactionAdapter.TransactionClickListListener,
-    ErrorAdapter.ErrorClickListListener {
+    BaseChuckerActivity() {
 
     private lateinit var viewModel: MainViewModel
     private lateinit var binding: ChuckerActivityMainBinding
@@ -66,14 +60,6 @@ internal class MainActivity :
         } else {
             HomePageAdapter.SCREEN_ERROR_INDEX
         }
-    }
-
-    override fun onErrorClick(throwableId: Long, position: Int) {
-        ErrorActivity.start(this, throwableId)
-    }
-
-    override fun onTransactionClick(transactionId: Long, position: Int) {
-        TransactionActivity.start(this, transactionId)
     }
 
     companion object {
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/MainActivity.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/MainActivity.kt
@@ -2,11 +2,9 @@ package com.chuckerteam.chucker.internal.ui
 
 import android.content.Intent
 import android.os.Bundle
-import androidx.appcompat.widget.Toolbar
 import androidx.lifecycle.ViewModelProvider
-import androidx.viewpager.widget.ViewPager
-import com.chuckerteam.chucker.R
 import com.chuckerteam.chucker.api.Chucker
+import com.chuckerteam.chucker.databinding.ChuckerActivityMainBinding
 import com.chuckerteam.chucker.internal.ui.error.ErrorActivity
 import com.chuckerteam.chucker.internal.ui.error.ErrorAdapter
 import com.chuckerteam.chucker.internal.ui.transaction.TransactionActivity
@@ -19,37 +17,36 @@ internal class MainActivity :
     ErrorAdapter.ErrorClickListListener {
 
     private lateinit var viewModel: MainViewModel
-    private lateinit var viewPager: ViewPager
+    private lateinit var binding: ChuckerActivityMainBinding
 
     private val applicationName: CharSequence
         get() = applicationInfo.loadLabel(packageManager)
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContentView(R.layout.chucker_activity_main)
-
-        val toolbar = findViewById<Toolbar>(R.id.chuckerTransactionToolbar)
-        setSupportActionBar(toolbar)
-        toolbar.subtitle = applicationName
-
+        binding = ChuckerActivityMainBinding.inflate(layoutInflater)
         viewModel = ViewModelProvider(this).get(MainViewModel::class.java)
 
-        viewPager = findViewById(R.id.chuckerMainViewPager)
-        viewPager.adapter = HomePageAdapter(this, supportFragmentManager)
-
-        val tabLayout = findViewById<TabLayout>(R.id.chuckerMainTabLayout)
-        tabLayout.setupWithViewPager(viewPager)
-
-        viewPager.addOnPageChangeListener(object : TabLayout.TabLayoutOnPageChangeListener(tabLayout) {
-            override fun onPageSelected(position: Int) {
-                super.onPageSelected(position)
-                if (position == 0) {
-                    Chucker.dismissTransactionsNotification(this@MainActivity)
-                } else {
-                    Chucker.dismissErrorsNotification(this@MainActivity)
+        with(binding) {
+            setContentView(root)
+            setSupportActionBar(chuckerTransactionToolbar)
+            chuckerTransactionToolbar.subtitle = applicationName
+            chuckerMainViewPager.adapter = HomePageAdapter(this@MainActivity, supportFragmentManager)
+            chuckerMainTabLayout.setupWithViewPager(chuckerMainViewPager)
+            chuckerMainViewPager.addOnPageChangeListener(
+                object : TabLayout.TabLayoutOnPageChangeListener(chuckerMainTabLayout) {
+                    override fun onPageSelected(position: Int) {
+                        super.onPageSelected(position)
+                        if (position == 0) {
+                            Chucker.dismissTransactionsNotification(this@MainActivity)
+                        } else {
+                            Chucker.dismissErrorsNotification(this@MainActivity)
+                        }
+                    }
                 }
-            }
-        })
+            )
+        }
+
         consumeIntent(intent)
     }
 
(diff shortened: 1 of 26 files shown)


vbuberen:
Renamed all views because can forget about possible ids clashing, like it happened with `findViewById` - now we are always sure that we use the right view thanks to `ViewBinding`.

vbuberen:
After resources renaming in #257 I need to merge that PR first and update this one afterwards.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionOverviewFragment.kt
```diff
@@ -83,28 +52,33 @@ internal class TransactionOverviewFragment : Fragment() {
...
     }
+
+    private fun populateUI(transaction: HttpTransaction?, encodeUrl: Boolean) {
+        with(overviewBinding) {
+            url.text = transaction?.getFormattedUrl(encodeUrl)
+            method.text = transaction?.method
+            protocol.text = transaction?.protocol
+            status.text = transaction?.status.toString()
+            response.text = transaction?.responseSummaryText
+            ssl.setText(if (transaction?.isSsl == true) R.string.chucker_yes else R.string.chucker_no)
```
MiSikora: Shouldn't this show an empty text or hide a field in case `isSsl` is null? Realistically users should never experience since it comes from the request which is processed rather immediately. Just something i noticed.

But maybe this suggests that the DB model is not proper. Maybe it should be split into requests and responses tables and transaction table would refer to those by their IDs. This is definitely not for this PR but maybe it should be discussed in some other ticket? I can create one and describe the issue I see with the current DB model.

vbuberen:
@MiSikora I don't agree with your point about memory leaks in fragments. Despite it is being said in Google`s tutorial, there is no case in Chucker where we would get a memory leak, because we don't use anything like `retain` fragments, etc. While getting the fragment from the backstack we will assign a new binding anyway. So there will be no leaks. 
If I am wrong, please provide a case and an evidence that there is a leak, because I don't see any at the moment.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainActivity.kt
```diff
@@ -2,54 +2,45 @@ package com.chuckerteam.chucker.internal.ui
...
+            tabLayout.setupWithViewPager(viewPager)
+            viewPager.addOnPageChangeListener(
+                object : TabLayout.TabLayoutOnPageChangeListener(tabLayout) {
+                    override fun onPageSelected(position: Int) {
+                        super.onPageSelected(position)
+                        if (position == 0) {
+                            Chucker.dismissTransactionsNotification(this@MainActivity)
+                        } else {
+                            Chucker.dismissErrorsNotification(this@MainActivity)
+                        }
```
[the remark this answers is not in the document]
vbuberen: 👍

vbuberen:
And I am against the solution from the article you suggested, because it is just redundant complexity.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionAdapter.kt
```diff
@@ -30,55 +29,63 @@ internal class TransactionAdapter internal constructor(
...
+        }
+
+        override fun onClick(v: View?) {
+            transactionId?.let {
+                listener?.onTransactionClick(it, adapterPosition)
             }
         }
 
-        private fun setStatusColor(holder: ViewHolder, transaction: HttpTransactionTuple) {
+        private fun setStatusColor(itemBinding: ChuckerListItemTransactionBinding, transaction: HttpTransactionTuple) {
```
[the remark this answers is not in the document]
vbuberen: 👍

library/src/main/java/com/chuckerteam/chucker/internal/ui/error/ErrorActivity.kt
```diff
@@ -94,13 +86,21 @@ internal class ErrorActivity : BaseChuckerActivity() {
     }
 
     private fun populateUI(throwable: RecordedThrowable) {
-        title.text = throwable.formattedDate
-        tag.text = throwable.tag
-        clazz.text = throwable.clazz
-        message.text = throwable.message
-        stacktrace.text = throwable.content
+        errorBinding.apply {
```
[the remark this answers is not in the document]
vbuberen: I don't agree that we should always stick to one particular scope function.
`with` or `apply` decision is based mainly on actions applied and on how you read it. 
In `onCreate()` we are doing something using `errorBinding`, that is why I used `with` there.
In `populateUI()` I used `apply`, because I explicitly mean that I want to apply some properties to items from this binding.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionOverviewFragment.kt
vbuberen: Yeah, I believe it should be hidden.

As to DB issues - sure, feel free to raise it. I wasn't involved in the current design, so would gladly join if needed.

library/src/main/java/com/chuckerteam/chucker/internal/ui/error/ErrorActivity.kt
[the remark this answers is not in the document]
MiSikora: It is still not clear to me to be honest. Should I split binding into two blocks of `with` and `apply` if I first do something with it and the apply some properties to it?

MiSikora:
> @MiSikora I don't agree with your point about memory leaks in fragments. Despite it is being said in Google`s tutorial, there is no case in Chucker where we would get a memory leak, because we don't use anything like `retain` fragments, etc. While getting the fragment from the backstack we will assign a new binding anyway. So there will be no leaks.

But there will be leaks in the meantime. Fragments will hold reference to memory which is and no longer will be used. I understand that it is something negligible and it will not cause crashes, especially in a QA tool with couple of screens, but those are still leaks. It might be fine to ignore them at this scale but those are still leaks, even if temporary.

> If I am wrong, please provide a case and an evidence that there is a leak, because I don't see any at the moment.

There was even a short discussion on Twitter about this. https://twitter.com/Piwai/status/1201686009349455872

> And I am against the solution from the article you suggested, because it is just redundant complexity.

I don't see how it is redundant but I agree that it introduces complexity. It really depends on if and how you want to deal with leaks. If there is no need to handle them then this solution (or nulling out views) is probably unnecessary. Otherwise there can be one well-defined mechanism that handles that or manual assigning of `null` to views.

vbuberen:
Thanks for pointing out to this discussion, but I am already aware of it.

This discussion is also the reason why I asked you to provide an evidence in particular Chucker case. We have LeakCanary here, so you could easily catch the issue if it existed. If you are able to do it - I would gladly fix it myself.

And this is also the reason for me to say `redundant complexity`. We don't have this problem at the moment, right? So why would we add more code?
I believe that Chucker isn't a showcase for all theoretically possible cases, but rather a tool which does its job and gets new code when the new feature request appears or the new bug reported. Otherwise, why solve the problem which doesn't even exist? (I might be wrong if you could catch a memory leak here).

library/src/main/java/com/chuckerteam/chucker/internal/ui/error/ErrorActivity.kt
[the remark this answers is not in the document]
vbuberen: No. I would stick to one function in case of using it in the same place. But in this particular case I use these extensions in different functions. 
So here it is just a matter of preference, because there are no clear conventions on these functions usage. Also, we don't care about the return type in both cases, since we are just removing multiple repetitions of `errorBinding` to improve readability. Finally, we will end up with the same Java code generated with additional variable generated, so there is no difference in preferring one or another in terms of some performance benefits.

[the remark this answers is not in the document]
MiSikora: Ok, I got it now.

vbuberen:
Talking about LeakCanary - see that we have an old version here. 😄 
Going to create a separate PR with update.

MiSikora:
Oh no, don't get me wrong. I'm not advocating that there are memory issues in Chucker. I'm only pointing out that views might be mishandled in fragments, even if leaks are not noticeable or even present right now.

Issues we are talking about would only occur if Chucker would add fragment transactions and if it is fine to postpone handling them for such a case that it is definitely ok.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionOverviewFragment.kt
MiSikora: I'm not sure if you forgot about it during CR fixes. Other than that LGTM.

vbuberen:
Ok, I got your point. I am not the only core contributor here, so my point isn't the single source of truth, but I am advocating to solve such problems one step at a time (when they really arise).

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionOverviewFragment.kt
vbuberen: Yeah, forgot it. Already fixed. Please approve it if other things are fine.

library/src/main/res/layout/chucker_transaction_item_image.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <ImageView xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/chuckerTransactionResponseBinaryData"
+    android:id="@+id/binaryData"
```
cortinico: Thanks for driving this forward @vbuberen @MiSikora 
Out of curiosity, why we renamed back all the IDs to don't have `chucker` in the name? Was this because we wanted to have a cleaner code (due to view binding)?

vbuberen: When we used `findViewById` we could get into trouble when user has the same id for some view as Chucker. In such cases Chucker would crash, since, for example, instead of `toolbar` from `activity_transaction` it would find a `toolbar` from some layout in user's project. Due to this issue I renamed all ids for views for these long `chucker.....` items.

Now, when we switched to `ViewBinding` we can be 100% sure that the generated binding class has references to proper views, so we can change their ids to more concise ones.

This was one of the reasons I decided to switch Chucker to `ViewBinding`.

cortinico: Awesome, thanks for the clarification

PR #257, opened 2020-02-29
Add missing ThrowableViewModel

vbuberen:
## :page_facing_up: Context
Recently Chucker switched to ViewModel approach, but one of its screens still relied on old approach with calling Repository methods from Activity. To fix it, I added an `ErrorViewModel`.
Currently it does nothing except retrieving the throwable, but it is better to have a uniform approach across the whole library, so we are ready for easier extensions and improvements.
As a side update I moved title retrieval in `TransactionActivity` from `onResume()` to `onCreate()`, since I don't see reasons for getting it on every screen resuming.

## :pencil: Changes
- Added `ErrorViewModel`
- Moved title retrieval from `onResume()` to `onCreate()` in `TransactionActivity`


library/src/main/java/com/chuckerteam/chucker/internal/ui/error/ErrorViewModel.kt
```diff
@@ -0,0 +1,25 @@
+package com.chuckerteam.chucker.internal.ui.error
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import com.chuckerteam.chucker.internal.data.entity.RecordedThrowable
+import com.chuckerteam.chucker.internal.data.repository.RepositoryProvider
+
+internal class ErrorViewModel(
```
cortinico: nit: I would rename this to `ThrowableViewModel`, let's try to be consistent with the naming. I noticed that we have a mixture of `Error` and `Throwable` in the codebase. I would stick to only one. Given that `Throwable` is what we use for the database, we should probably use it and apply it also to `ErrorActivity`, `ErrorViewModel`, etc. (so they should be `ThrowableActivity` and `ThrowableViewModel`).

I'm anyway not strongly opinionated on this front, so we can tackle this in the future as well.

vbuberen: Absolutely agree here. Also was regularly confused by mixing `error` and `throwable`. Will update the PR.

vbuberen:
Ok, updated naming everywhere.
Places without renaming left:
- Tab name on the main screen.
- `Chucker` class functions in the `api` package.
- `ChuckerCollector` class functions in the `api` package.
- This branch name 😄 

There were no renaming for Chucker classes in the `api` package to not introduce breaking changes. We might add `deprecate` for things, like `onError` and add `onThrowable`, for example to complete our switch to a single naming style. 
What do you think @cortinico ?

cortinico:
> There were no renaming for Chucker classes in the api package to not introduce breaking changes. We might add @Deprecated for things, like onError and add onThrowable, for example, to complete our switch to a single naming style.
> What do you think @cortinico ?

Great job!
I don't think it's worth to modify the API surface because of this change. At least, not at this point in time. It just adds overhead on the end user because of our internal renaming.

I will keep it in the backlog for now, and pull this deprecation in as soon as we touch that function again. Maybe we can leave a comment like (to be consistent we should call this method `onThrowable`). 
I anyway don't have a really strong opinion on this.

library/src/main/res/values/strings.xml
```diff
@@ -37,16 +37,16 @@
     <string name="chucker_tab_network">Network</string>
     <string name="chucker_tab_errors">Errors</string>
     <string name="chucker_network_tutorial">This tab displays network requests recorded with the ChuckerInterceptor. Add the interceptor to your OkHttp client to reap the benefits.</string>
-    <string name="chucker_errors_tutorial">This tab displays error recorded with the ChuckerCollector.onError method. Call this method when an error is thrown to check your stacktrace live.</string>
-    <string name="chucker_networks_notification_category">Chucker network requests</string>
-    <string name="chucker_errors_notification_category">Chucker throwables</string>
-    <string name="chucker_share_error_title">Share error details</string>
+    <string name="chucker_throwable_tutorial">This tab displays throwables recorded with the ChuckerCollector.onError method. Call this method when an error is thrown to check your stacktrace live.</string>
```
cortinico: nit: `Throwables` (uppercase) as it's a reference to subclasses of that class.

vbuberen: But I thought that it is a self-made word, since we don't have a word for multiple `Throwable` items.

cortinico: yeah makes sense :)

PR #258, opened 2020-03-01
Feature/truth

vbuberen:
## :page_facing_up: Context
As a part of test coverage improvement I would like to add [Truth](https://truth.dev/) for more concise assertions in tests.
While updating assertions found one test with missing `@Test` annotation and issue with one of tests.

## :pencil: Changes
- Added Truth.
- Added missing `@Test` annotation.
- Fixed `testGsonConfiguration_willParseDateTime()` since its success or failure was dependent on the timezone of the machine running it. Namely, `Date(0)` returns `Jan. 1 1970`, but the time is always different in different timezones, so the time specified in the expected result string wasn't the same for me.

### DIFF
--- a/library/src/test/java/com/chuckerteam/chucker/internal/support/LiveDataDistinctUntilChangedTest.kt
+++ b/library/src/test/java/com/chuckerteam/chucker/internal/support/LiveDataDistinctUntilChangedTest.kt
@@ -4,7 +4,7 @@ import androidx.arch.core.executor.testing.InstantTaskExecutorRule
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.distinctUntilChanged
 import com.chuckerteam.chucker.test
-import junit.framework.TestCase.assertEquals
+import com.google.common.truth.Truth.assertThat
 import org.junit.Rule
 import org.junit.Test
 
@@ -16,7 +16,7 @@ class LiveDataDistinctUntilChangedTest {
         val upstream = MutableLiveData<Any?>(null)
 
         upstream.distinctUntilChanged().test {
-            assertEquals(null, expectData())
+            assertThat(expectData()).isEqualTo(null)
         }
     }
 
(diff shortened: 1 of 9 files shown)


library/src/test/java/com/chuckerteam/chucker/internal/support/LiveDataDistinctUntilChangedTest.kt
```diff
@@ -54,13 +54,13 @@ class LiveDataDistinctUntilChangedTest {
 
         upstream.distinctUntilChanged().test {
             upstream.value = null
-            assertEquals(null, expectData())
+            assertThat(expectData()).isEqualTo(null)
```
cortinico: `isNull()`

library/src/test/java/com/chuckerteam/chucker/internal/support/LiveDataDistinctUntilChangedTest.kt
```diff
@@ -16,7 +16,7 @@ class LiveDataDistinctUntilChangedTest {
         val upstream = MutableLiveData<Any?>(null)
 
         upstream.distinctUntilChanged().test {
-            assertEquals(null, expectData())
+            assertThat(expectData()).isEqualTo(null)
```
cortinico: `isNull()`

library/src/test/java/com/chuckerteam/chucker/internal/support/LiveDataDistinctUntilChangedTest.kt
```diff
@@ -35,16 +35,16 @@ class LiveDataDistinctUntilChangedTest {
...
-            assertEquals(1, expectData())
+            assertThat(expectData()).isEqualTo(1)
 
             upstream.value = 2
-            assertEquals(2, expectData())
+            assertThat(expectData()).isEqualTo(2)
 
             upstream.value = null
-            assertEquals(null, expectData())
+            assertThat(expectData()).isEqualTo(null)
```
cortinico: `isNull()`

PR #265, opened 2020-03-05
Add stale builds canceller

vbuberen:
## :page_facing_up: Context
When first PR with GH Actions was created I saw that builds happen for every single action, be it push or pull request. So it seems that we are ending up with too much redundant work.
According to [this thread ](https://github.community/t5/GitHub-Actions/Github-actions-Cancel-redundant-builds/td-p/29549) Github has no native solution for this issue.
In that thread I also saw a few 3rd party actions and decided to test one of them.

Screenshot 2020-03-05 at 13 26 30

## :pencil: Changes
- Added https://github.com/marketplace/actions/gh-actions-stale-run-canceller to test auto cancelling

## :stopwatch: Next steps
If it works, I will add this step into the main workflow as well.

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -12,6 +12,7 @@ jobs:
       uses: yellowmegaman/gh-build-canceller@master
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
+        workflows_filter: "Validate Gradle Wrapper|Pre Merge Checks"
     - name: Validate Gradle Wrapper
       uses: gradle/wrapper-validation-action@v1
       
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -10,6 +10,8 @@ jobs:
       uses: actions/checkout@v2
     - name: Stale run canceller
       uses: yellowmegaman/gh-build-canceller@master
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
     - name: Validate Gradle Wrapper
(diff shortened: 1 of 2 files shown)


vbuberen:
Can't make it work. Probably, something wrong with the config.

cortinico:
> be it push or pull request. So it seems that we are ending up with too much redundant work.

I'm ok having it for `pull_request` only

vbuberen:
Yeah, good idea about leaving only `pull requests`. I will update the config and see how it behaves.

vbuberen:
@cortinico Updated to trigger workflows only for pull requests. Now we should be good to go.

.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -1,10 +1,13 @@
 name: Validate Gradle Wrapper
-on: [push, pull_request]
+on: pull_request
```
cortinico: Actually having also

```
on:
  push:
    branches:
      - develop
	  - release
```

Would be useful

cortinico: The idea is that we do have a run just after the merge on the `develop` and `release` branch so we're sure the HEAD is green.

vbuberen: That was my next PR)

cortinico: Great, so do you want to merge this?

vbuberen: No, I just needed to get to the laptop to update this PR.
Now it is ready

cortinico: Great, feel free to merge as soon as it's ✅

Issue #242, opened 2020-02-17
Chucker 3.1.2 re-writing the Server Response; throws IOException when used as a networkInterceptor


jannisveerkamp:
## :writing_hand: Describe the bug
After updating to 3.1.2 (from 3.0.1) every request fails with an `IOException`:
```
java.io.IOException: ID1ID2: actual 0x00007b22 != expected 0x00001f8b
System.err:     at okio.GzipSource.checkEqual(GzipSource.java:205)
System.err:     at okio.GzipSource.consumeHeader(GzipSource.java:120)
System.err:     at okio.GzipSource.read(GzipSource.java:73)
System.err:     at okio.RealBufferedSource.read(RealBufferedSource.java:51)
System.err:     at okio.ForwardingSource.read(ForwardingSource.java:35)
System.err:     at retrofit2.OkHttpCall$ExceptionCatchingResponseBody$1.read(OkHttpCall.java:288)
System.err:     at okio.RealBufferedSource.select(RealBufferedSource.java:100)
```

Chucker `3.0.1` returned the original Response. `3.1.x` seems to process the response and returns this "processed Response" (see [here](https://github.com/ChuckerTeam/chucker/blob/3.1.2/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt#L68)).

I see the need of processing the Response and doing stuff with it. But the Interceptor has to return the original response at the end, since Chucker is only some kind of debugging ouput, which should not change the Request/Response in any way.

## :bomb: Steps to reproduce
- Add `ChuckerInterceptor` as a `networkInterceptor`
- I think the Request has to be gzipped

## :wrench: Expected behavior
- Chucker should not alter the Server Response

## :iphone: Tech info
Chucker version: `3.1.2`
OkHttp version: `3.14.6`

## :page_facing_up: Additional context
This only happens when the `ChuckerInterceptor` is added as a `networkInterceptor`.
The Notification shows the correct Request with the Response.

vbuberen:
Thanks for your report.

@cortinico the more such reports we have the more I want to rework interceptor in a way where we just do a copy of response for processing, while returning the original response further to the chain. Something like I showed you when we discussed the issue presented in `3.1.0`.
Because it seems the most viable way since we could be sure that we aren't altering the original response in the chain.

cortinico:
[...]

PR #267, opened 2020-03-06
Multi cast upstream response for Chucker consumption.

### LINKED ISSUE #218: OutOfMemoryError in interceptor
This came to me from firebase crashlytics, i don't know the reproduction steps
```
Fatal Exception: java.lang.OutOfMemoryError: Failed to allocate a 13536178 byte allocation with 5074056 free bytes and 4MB until OOM
       at okio.Buffer.readByteArray(Buffer.java:1438)
       at okio.Buffer.readString(Buffer.java:301)
       at okio.Buffer.readString(Buffer.java:290)
       at com.chuckerteam.chucker.api.ChuckerInterceptor.processResponseBody(ChuckerInterceptor.java:156)
       at com.chuckerteam.chucker.api.ChuckerInterceptor.processResponse(ChuckerInterceptor.java:130)
       at com.chuckerteam.chucker.api.ChuckerInterceptor.intercept(ChuckerInterceptor.java:61)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:112)
       at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:87)
// rest of interceptors
```

## :bomb: To Reproduce
I don't know how to reproduce that

Chucker should handle any payload size gracefully

## :iphone: Smartphone
 - Device: Galaxy A5 (2016)
 - OS: 7.1.1
 - Lib version: 3.1.1

### LINKED ISSUE #263: ChuckerInterceptor fetches the whole body even if the application isn't doing so
Simply plugging in `ChuckerInterceptor` causes the whole response body to be read, even if the application isn't reading the whole of it. Not only this is a very unwanted side effect, this can also crash the app. This is an example crash while trying to (not) read a ~50MB file:

    2020-03-04 22:49:34.915 8061-8119 E/AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
        Process: process, PID: 8061
        java.lang.OutOfMemoryError: Failed to allocate a 102720552 byte allocation with 25165824 free bytes and 31MB until OOM, target footprint 260095632, growth limit 268435456
            at java.lang.StringFactory.newStringFromBytes(StringFactory.java:225)
            at java.lang.StringFactory.newStringFromBytes(StringFactory.java:249)
            at okio.Buffer.readString(Buffer.kt:306)
            at okio.Buffer.readString(Buffer.kt:295)
            at com.chuckerteam.chucker.api.ChuckerInterceptor.processResponseBody(ChuckerInterceptor.kt:159)
            at com.chuckerteam.chucker.api.ChuckerInterceptor.processResponse(ChuckerInterceptor.kt:133)
            at com.chuckerteam.chucker.api.ChuckerInterceptor.intercept(ChuckerInterceptor.kt:65)
            at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:100)
            at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:197)
            at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:502)
            at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
            at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
            at java.lang.Thread.run(Thread.java:919)

you can reproduce this by starting a fake web server and looking at how much it sends (and observing the crash):

    cat /dev/urandom > file
    { echo -ne "HTTP/1.0 200 OK\r\nContent-Length: $(wc -c <file)\r\n\r\n"; cat file } | pv | nc -l 8080

It is my impression that the interceptor should only observe the doings of the application and should report either what the application reads or what the device receives.

P.S. setting `maxContentLength` in `ChuckerInterceptor()` seems to have no effect

MiSikora:
##  :page_facing_up: Context

There are couple of issues. #218 #263 

## :pencil: Changes

I changed mostly `ChuckerInterceptor`. It uses now custom source for responses that multicasts to end consumer and to a temporary file from which Chucker can read once the consumption is done by the end consumer.

## :paperclip: Related PR

There will be conflicts with #250. I would suggest to merge this one first. It will be easier to account for changes in #250 than other way around.

## :no_entry_sign: Breaking

No. I changed the constructor but I made it internal and added another one that is public and compatible with the old one.

I don't know how to manually test it reliably. Make a request for a really big file and save it on a device you should get an OOM with the old code.

Theoretically user still can get errors if there is not enough disk space but I see it as an edge case with limiting the response size.

## :stopwatch: Next steps

No further steps expected. Maybe we should create an issue for doing the same with requests? Also, maybe Chucker should process these files on its custom `Executor`.

Closes #218
Closes #263


MiSikora:
Hmmm… I see now that for some reason `deflate` request is not getting processed. Will dig into it.

MiSikora:
I know that testing is lacking but for the things that I fixed in 4751cf0 and 03990f6 but it would be much easier for me to write tests for it in #250 after this one is merged. Would it be ok with you guys?

vbuberen:
Hey @MiSikora. Yeah, probably, it would be easier.
The only moment here is that I am not a big fan of this approach with having some temp files. I would like to stick to approach with in-memory handling of the request/response, since storage is slower than RAM and because I would prefer Chucker to leave as low footprint on the device/emulator as possible.

Another moment that disturbs me that currently all the changes in the core of the library (classes like ChuckerInterceptor) are done mainly without any distinct vision or plan, so each individual pushes his own. I am against it, because it makes project development unpredictable and the library itself being a box of different approaches instead of following some defined vision.

MiSikora:
Hi @vbuberen. This is a proposal implementation of a thing discussed here. https://github.com/ChuckerTeam/chucker/issues/242#issuecomment-591329250

MiSikora:
@vbuberen @cortinico Any further thoughts on this? I'll gladly discuss my reasoning behind this implementation.

File processing seems better to me because it has lower impact on a device. If responses were multicasted to a in-memory buffers, then it might still result in OOM for multiple request. This is not very likely, but plausible. What is more realistic scenario is increased request size to some amount that would not fit in the memory. This would result in the OOM for sure. With multiple parallel requests it is even worse.

I'm not sure I understand the argument about the impact about the footprint. Chucker already leaves a footprint due to the database. Files are temporary and are removed anyway. Manual calls to `delete()` will almost always remove the files. If not, they will be removed after some time due to being in the cache directory. And finally they will be removed after uninstalling the app.

Processing is also theoretically slower but disk IO is nothing compared to the network IO.

Additional benefit of using files (apart from more or less solving OOM issues) is that in the future Chucker could skip storing responses as blobs in the database directly and rely only on file handles.  Not sure if it is a direction that you would like to go, but in my opinion it would be beneficial.

What do you think?

vbuberen:
Ok, I agree with your point. I just didn't think about big payloads properly before and yeah, temp files are the best solution. Also, I should admit that we can always rework the solution at any moment if we ever need it.
So, green light from me. Going to review this PR.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -15,6 +19,7 @@ import okhttp3.Response
 import okhttp3.ResponseBody
 import okio.Buffer
 import okio.GzipSource
+import okio.Okio
 
 private const val MAX_BLOB_SIZE = 1000_000L
```
vbuberen: Seems that this const is unused, because the same is present in companion object.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -69,10 +83,13 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
 
-        val processedResponse = processResponse(response, transaction)
-        collector.onResponseReceived(transaction)
-
-        return processedResponse
+        processResponseMetadata(response, transaction)
+        return multiCastResponseBody(response) { cachedResponseFile ->
+            val buffer = cachedResponseFile?.let { readResponseBuffer(it, response.isGzipped) }
+            cachedResponseFile?.delete()
+            if (buffer != null) processResponseBody(response, buffer, transaction)
```
vbuberen: This whole block starting with `return` is too hard to read. Let's rewrite it in more simple way without putting multiple functions in `return`.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -140,50 +160,62 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
-        val buffer = Buffer().apply { responseSource.use { writeAll(it) } }
+        val teeSource = TeeSource(
+            responseBody.source(),
+            fileFactory.create(),
+            maxContentLength,
+            onResponseBodyReady
+        )
+
+        return response.newBuilder()
+            .body(ResponseBody.create(contentType, contentLength, Okio.buffer(teeSource)))
```
vbuberen: I insist on returning the original `Response` object further in chain without any builders, cloning, etc.
We had enough side effects and issues with providing processed or somehow changed objects. 
And while I see that here we should get the same object I still don't like that it is recreated. 
In some issues we already mentioned that Chucker should be as transparent as possible and this is possible only if we do copy only for Chucker processing without anything like that.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -140,50 +160,62 @@ class ChuckerInterceptor @JvmOverloads constructor(
...
             }
         } else {
             transaction.isResponseBodyPlainText = false
 
             val isImageContentType =
                 (contentType?.toString()?.contains(CONTENT_TYPE_IMAGE, ignoreCase = true) == true)
 
-            if (isImageContentType && buffer.size() < MAX_BLOB_SIZE) {
-                transaction.responseImageData = buffer.clone().readByteArray()
+            if (isImageContentType && responseBodyBuffer.size() < MAX_BLOB_SIZE) {
```
vbuberen:
```suggestion
            if (isImageContentType && (responseBodyBuffer.size() < MAX_BLOB_SIZE)) {
```

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -0,0 +1,77 @@
...
+            throw e
+        }
+
+        if (bytesRead == -1L) {
+            sideStream.close()
+            return -1L
+        }
+
+        totalBytesRead += bytesRead
+        if (!reachedLimit && totalBytesRead <= readBytesLimit) {
```
vbuberen:
```suggestion
        if (!reachedLimit && (totalBytesRead <= readBytesLimit)) {
```

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -0,0 +1,77 @@
...
+            sink.copyTo(sideStream.buffer(), sink.size() - bytesRead, bytesRead)
```
vbuberen: Why don't you move `sink.size() - bytesRead` into a variable?

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    override fun close() {
+        onSideChannelReady(sideChannel)
+        sideStream.close()
+        upstream.close()
+    }
+
+    override fun timeout(): Timeout {
+        return upstream.timeout()
```
vbuberen: What about removing `return`?

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
MiSikora: I don't see how it would be possible to stream original bytes to two sinks at the same time without replacing the response. This is a mechanism that is used under the hood by `OkHttp` for caching (just a simpler version). It is also recommended way by Square (https://github.com/square/okio/issues/186). Do you have any other approach in mind that would solve the problems addressed by this?

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
MiSikora: I didn't touch because it is removed in another PR but I'll remove it here after we find a solution to the main issue in this PR. :)

MiSikora:
I will make a CR fixes some time later to focus discussion on things that are not yet clear to me.

> On one side I also realised that the `maxContentLength` constructor parameters was used to truncate ONLY the request and not the response. This PR fixes this.
> I'm still wondering if this approach is a bit over engineered as truncating the responses will probably solve the OOM as well.

I don't think this is really viable. One approach might be to rely on `Content-Length` header but it is not always available and responses can be from bytes to megabytes. We could skip such responses but them we will have some false negatives.

Another approach might to try to do something like this:

```kotlin
val response = chain.proceed(chain.request())
val length = response.body()?.source()?.buffer?.size() ?: return response
if (length <= maxContentLength) {
  // Process response.
}
return response
```

This, however, also does not work. Buffer size is always 0 because it has not been read from at this point and will be some time in the future when the user consumes their network response and process it or whatever they do with it.

The only reason that we can check it right now is because of this piece of code.

https://github.com/ChuckerTeam/chucker/blob/c34300232fab219a7fa4b407243e4b03a3e47634/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt#L166

And this is the culprit here because it loads the whole response into the memory and potentially leads to OOM.

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -0,0 +1,79 @@
...
+         */
+        fun onSuccess(file: File)
+
+        /**
+         * Called when there was an issue while copying bytes to the [file].
+         *
+         * It might occur due to an exception thrown while reading bytes or due to exceeding capacity limit.
+         */
+        fun onFailure(exception: IOException, file: File)
+    }
```
MiSikora: I decided to go with a callback instead of magic headers. It gives type safety and is more extensible. In the future it is possible to add `onContentLengthAvailable(length: Long)` method in order to resolve #209.

MiSikora:
I applied most of the suggestions from the CR. One thing still stands - replacing the response. As I mentioned I don't know any other way that would solve the problems addressed by this PR. One thing I'd like to highlight is about the volatile nature of replacing the response and issues that occurred in the past. It is reasonable to be hesitant to do this again, however previous code did not have any tests for `ChuckerInterceptor`. Now that they are being gradually added they should give a confidence in doing such changes.

vbuberen:
Sorry for taking that long. I just want to play around in attempt to not replace response, because I don't know possible approach at the moment. Will try to do tomorrow.

MiSikora:
No worries. Take your time, I just want to provide as much information as possible. I know it is rather big change so it's best to consider it carefully.

vbuberen:
If no other commits planned here, I will merge it.

MiSikora:
Yes, it is ready to go. I'll update #250 soon

PR #272, opened 2020-03-11
Migration from Gson to Moshi

hitanshu-dhawan:
## :page_facing_up: Context
Moshi is a successor to Gson and often referred as "Gson 3.0".
There are many plus points for using Moshi over Gson listed [here](https://www.reddit.com/r/androiddev/comments/684flw/why_use_moshi_over_gson/dgx3gpm/).
Also, all the future developments will be done on Moshi instead of Gson.

## :pencil: Changes
- Refactored `JsonConverter.instance` to return `Moshi` instance instead of `Gson`.
- Created `MoshiUtils` for some utility methods like getting various JsonAdapters and adding convenience methods that serializes nulls and does pretty printing.
- Added a custom JsonAdapter called `DateJsonAdapter` for parsing `Date` objects.
- Refactored `HttpHeader` in accordance with Moshi.
- Refactored `HttpTransaction`, the `requestHeaders` and `responseHeaders` methods that used Gson where changed to use Moshi. 
- Refactored `FormatUtils.formatJson()` to use `JSONObject(json).toString(indentSpaces)` instead of relying on Gson utility methods.
- Refactored `JsonConverterTest` in accordance with Moshi. Removed `testGsonConfiguration_willUseLowerCaseWithUnderscores` test case as field naming policy is not recommended by Moshi.
- Added a new test case in `FormatUtilsTest` that checks pretty printing of json arrays, also made `FormatUtilsTest` to use robolectric as JSONObject is bundled up with the Android SDK.

## :no_entry_sign: Breaking
No. All test cases are passing.


library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
```diff
@@ -56,10 +57,13 @@ internal object FormatUtils {
 
     fun formatJson(json: String): String {
         return try {
-            val je = JsonParser.parseString(json)
-            JsonConverter.instance.toJson(je)
-        } catch (e: JsonParseException) {
-            json
+            JSONObject(json).toString(2)
```
vbuberen: Let's avoid `magic constants`.
For any other developer the first question here `What is 2 ?`. 
Provide some meaningful name here.

library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
```diff
@@ -56,10 +57,13 @@ internal object FormatUtils {
...
+        } catch (e: JSONException) {
+            try {
+                JSONArray(json).toString(2)
```
vbuberen: Same thing about `2` as above.

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
```diff
@@ -1,19 +1,14 @@
...
-        GsonBuilder()
-            .serializeNulls()
-            .setPrettyPrinting()
-            .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
-            .registerTypeAdapter(Date::class.java, DateTypeAdapter())
-            .create()
+    val instance: Moshi by lazy {
+        Moshi.Builder()
+            .add(DateJsonAdapter())
+            .add(KotlinJsonAdapterFactory())
```
vbuberen: With this line you are killing one of advantages mentioned in the Reddit thread you shared - size.
If you read the documentation for Moshi carefully, you could find such line:
`Note that the reflection adapter transitively depends on the kotlin-reflect library which is a 2.5 MiB .jar file.`
Info taken from here: https://github.com/square/moshi#reflection

library/src/test/java/com/chuckerteam/chucker/internal/support/FormatUtilsTest.kt
```diff
@@ -1,18 +1,17 @@
 package com.chuckerteam.chucker.internal.support
 
 import com.google.common.truth.Truth.assertThat
-import org.junit.jupiter.api.Test
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
 
+@RunWith(RobolectricTestRunner::class)
```
vbuberen: Why you decided to add Roboelectric? 
Do you know why it is used for?

vbuberen: Why would we add the whole abandoned Roboelectic to just have `JSONObject`?
Why don't you just add dependency to json directly instead of it?
Like `testImplementation 'org.json:json:20190722'`

vbuberen:
Also, it seems that my point about testing Moshi isn't that valid, because this PR drops `JsonParser` of Gson to replace it with just default JSON formatting. So it means that all those Moshi advantages aren't playing any role here, since Chucker mostly formats JSONs to show them, not converts into models or something like this.

vbuberen:
Another possible issue here is that the PR is missing config for R8, since Chucker users might rely on it even in minified builds:
https://github.com/square/moshi#r8--proguard

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
vbuberen: So the idea is to replace this approach to code generation. Only with code generation we will get the advantage of Moshi of avoiding reflection and won't have to add `kotlin-reflect` into the library.
Here is the part of documentation about code generation:
https://github.com/square/moshi#codegen

library/build.gradle
```diff
@@ -71,7 +77,8 @@ dependencies {
     implementation "androidx.room:room-runtime:$roomVersion"
     kapt "androidx.room:room-compiler:$roomVersion"
 
-    implementation "com.google.code.gson:gson:$gsonVersion"
+    implementation "com.squareup.moshi:moshi:$moshiVersion"
+    implementation "com.squareup.moshi:moshi-kotlin:$moshiVersion"
```
vbuberen: This dependency should be replaced with `kapt "com.squareup.moshi:moshi-kotlin-codegen:1.9.2"` to switch to code generation approach I mentioned on another comment.
Here is another duplicate of the documentation: https://github.com/square/moshi#codegen

library/build.gradle
```diff
@@ -81,6 +88,7 @@ dependencies {
     testImplementation "com.squareup.okhttp3:mockwebserver:$okhttp3Version"
     testImplementation "androidx.arch.core:core-testing:$androidXCoreVersion"
     testImplementation "com.google.truth:truth:$truthVersion"
+    testImplementation "org.robolectric:robolectric:$robolectricVersion"
```
vbuberen: As I mentioned in another comment we don't need Roboelectic due to the following reasons:
1. With your changes you only need JSON, which can be added on its own.
2. Roboelectic is too big. By adding it we are seriously increasing build time for the project on the CI.
3. Roboelectic isn't maintained well enough and looks quite abandoned by Google.

library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
hitanshu-dhawan: Fixed.

library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
hitanshu-dhawan: Fixed.

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
hitanshu-dhawan: Replaced `moshi-kotlin` with `moshi-kotlin-codegen`.

library/build.gradle
hitanshu-dhawan: Replaced `moshi-kotlin` with `moshi-kotlin-codegen`.

library/src/test/java/com/chuckerteam/chucker/internal/support/FormatUtilsTest.kt
hitanshu-dhawan: Removed robolectric dependency and used moshi for formatting json.

Also, found something interesting, there are API differences between android's `org.json` and the original `org.json:json:20190722`.

I tried adding `testImplementation 'org.json:json:20190722'` and found that the function `JSONObject(json).toString(indentSpaces)` works differently on these two APIs.

For example,

For json string
```
{ "field" : null }
```
android's `org.json` will return
```
{
  "field": null
}
```
but, the original `org.json:json:20190722` will return
```
{"field": null}
```

library/build.gradle
hitanshu-dhawan: Removed robolectric dependency and used moshi for formatting json.

hitanshu-dhawan:
Added proguard-rules for Moshi.

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
```diff
@@ -31,20 +32,8 @@ class JsonConverterTest {
...
-            }
-            """.trimIndent()
-        )
-    }
-
-    @Test
-    fun testGsonConfiguration_willSerializeNulls() {
-        val json = JsonConverter.instance.toJson(NullTestClass(null))
+    fun testMoshiConfiguration_willSerializeNulls() {
+        val json = JsonConverter.instance.adapter(NullTestClass::class.java).addConvenienceMethods().toJson(NullTestClass(null))
```
vbuberen: With these changes you just made this test useless.
For Gson there was `serializeNulls()` in `JsonConverter` and this test checked that such configuration worked.
Currently you are manually adding `serializeNulls()` with your `addConvenienceMethods()` right inside this test, so you are not testing `JsonConverter` configuration, but your `addConvenienceMethods()` function.

library/src/main/java/com/chuckerteam/chucker/internal/support/DateJsonAdapter.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.squareup.moshi.ToJson
+import java.text.DateFormat
+import java.util.Date
+import java.util.Locale
+
+class DateJsonAdapter {
+
+    @ToJson
+    fun toJson(date: Date): String {
+        val dateFormat = DateFormat.getDateTimeInstance(DateFormat.DEFAULT, DateFormat.DEFAULT, Locale.US)
```
vbuberen: Why don't you create one variable for date format instead of copying it multiple times?

library/src/main/java/com/chuckerteam/chucker/internal/support/DateJsonAdapter.kt
```diff
@@ -0,0 +1,22 @@
...
+        return dateFormat.format(date)
+    }
+
+    @FromJson
+    fun fromJson(date: String): Date {
+        val dateFormat = DateFormat.getDateTimeInstance(DateFormat.DEFAULT, DateFormat.DEFAULT, Locale.US)
+        return dateFormat.parse(date)!!
```
vbuberen: I would prefer to declare that this function can throw an exception and try to avoid `!!`.

library/src/main/java/com/chuckerteam/chucker/internal/support/MoshiUtils.kt
```diff
@@ -0,0 +1,20 @@
+@file:JvmName("MoshiUtils")
```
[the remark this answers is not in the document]
hitanshu-dhawan: `OkHttpUtils` is also annotated with `JvmName`.
I thought it's a coding convention used in this project.

library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
```diff
@@ -55,12 +55,11 @@ internal object FormatUtils {
...
     fun formatJson(json: String): String {
-        return try {
-            val je = JsonParser.parseString(json)
-            JsonConverter.instance.toJson(je)
-        } catch (e: JsonParseException) {
-            json
-        }
+        val buffer = Buffer().writeUtf8(json)
+        val jsonReader = JsonReader.of(buffer)
+        val jsonValue = jsonReader.readJsonValue()
```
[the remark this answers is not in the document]
hitanshu-dhawan: Fixed, also added a test case for the same.

library/src/main/java/com/chuckerteam/chucker/internal/support/DateJsonAdapter.kt
hitanshu-dhawan: Fixed.

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
hitanshu-dhawan: Removed this test case from here.
Added all `MoshiUtils` related test cases in `MoshiUtilsTest` which will test both the methods.
- `getHttpHeaderListJsonAdapter()`
- `addConvenienceMethods()`

library/src/main/java/com/chuckerteam/chucker/internal/support/DateJsonAdapter.kt
hitanshu-dhawan: Fixed, also added several test cases for it.

library/src/main/java/com/chuckerteam/chucker/internal/support/DateJsonAdapter.kt
```diff
@@ -11,8 +11,8 @@ class DateJsonAdapter {
     private val dateFormat = DateFormat.getDateTimeInstance(DateFormat.DEFAULT, DateFormat.DEFAULT, Locale.US)
 
     @ToJson
-    fun toJson(date: Date): String = dateFormat.format(date)
+    fun toJson(date: Date?): String? = date?.let { dateFormat.format(it) }
```
vbuberen: Can you explain why you decided to do nullable parameter and return value here nullable?

library/src/main/java/com/chuckerteam/chucker/internal/support/DateJsonAdapter.kt
```diff
@@ -11,8 +11,8 @@ class DateJsonAdapter {
...
 
     @FromJson
-    fun fromJson(date: String): Date = dateFormat.parse(date)!!
+    fun fromJson(dateString: String?): Date? = dateString?.let { dateFormat.parse(it) }
```
vbuberen: Can you explain why you decided to do nullable parameter and return value here nullable?
In my previous comment I talked about throwing an exception, not trying to hide possible errors by making nullables and hiding everything behind `?.let`

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
```diff
@@ -32,6 +33,27 @@ class JsonConverterTest {
...
+    @Test(expected = JsonDataException::class)
+    fun testMoshiConfiguration_willDeserialize_Invalid_DateTime() {
+        val dateTestClass = JsonConverter.instance.adapter(DateTestClass::class.java)
+            .fromJson("""{"date":"abcd"}""")
+        assertThat(dateTestClass).isEqualTo(DateTestClass(null))
+    }
+
     @JsonClass(generateAdapter = true)
-    internal data class DateTestClass(val date: Date)
+    internal data class DateTestClass(val date: Date?)
```
vbuberen: Can you explain how you decided that it should be nullable?

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
```diff
@@ -32,6 +33,27 @@ class JsonConverterTest {
...
+    fun testMoshiConfiguration_willDeserialize_null_DateTime() {
+        val dateTestClass = JsonConverter.instance.adapter(DateTestClass::class.java)
+            .fromJson("""{"date":null}""")
+        assertThat(dateTestClass).isEqualTo(DateTestClass(null))
+    }
+
+    @Test(expected = JsonDataException::class)
+    fun testMoshiConfiguration_willDeserialize_Invalid_DateTime() {
+        val dateTestClass = JsonConverter.instance.adapter(DateTestClass::class.java)
+            .fromJson("""{"date":"abcd"}""")
```
vbuberen: Why you created this test? 
It is an impossible case to have such string to be converted with DateAdapter except explicitly doing it here. 
What are you testing here?

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
```diff
@@ -32,6 +33,27 @@ class JsonConverterTest {
         assertThat(dateTestClass).isEqualTo(DateTestClass(Date(0)))
     }
 
+    @Test
+    fun testMoshiConfiguration_willSerialize_null_DateTime() {
+        val dateTestJson = JsonConverter.instance.adapter(DateTestClass::class.java)
+            .toJson(DateTestClass(null))
```
vbuberen: Why you decided to create this test? Please, explain.

library/src/main/java/com/chuckerteam/chucker/internal/support/MoshiUtils.kt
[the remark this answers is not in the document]
hitanshu-dhawan: Yes, I am aware about the usage of `JvmName`.
Removed the redundant `JvmName` from `MoshiUtils`.

library/src/main/java/com/chuckerteam/chucker/internal/support/MoshiUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+package com.chuckerteam.chucker.internal.support
+
+import com.chuckerteam.chucker.internal.data.entity.HttpHeader
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
+
+internal fun Moshi.getHttpHeaderListJsonAdapter(): JsonAdapter<List<HttpHeader>> {
+    val type = Types.newParameterizedType(List::class.java, HttpHeader::class.java)
+    val adapter = this.adapter<List<HttpHeader>>(type)
```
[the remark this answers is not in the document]
hitanshu-dhawan: Fixed.

library/src/main/java/com/chuckerteam/chucker/internal/support/MoshiUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+    return adapter.addConvenienceMethods()
+}
+
+internal fun <T> JsonAdapter<T>.addConvenienceMethods(): JsonAdapter<T> {
+    return this
```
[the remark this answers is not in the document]
hitanshu-dhawan: Done.

library/src/main/java/com/chuckerteam/chucker/internal/support/DateJsonAdapter.kt
hitanshu-dhawan: Removed code related to `Date` parsing as it is not used anywhere in the codebase.

library/src/main/java/com/chuckerteam/chucker/internal/support/DateJsonAdapter.kt
hitanshu-dhawan: Removed code related to `Date` parsing as it is not used anywhere in the codebase.

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
hitanshu-dhawan: Removed code related to `Date` parsing as it is not used anywhere in the codebase.

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
hitanshu-dhawan: Removed code related to `Date` parsing as it is not used anywhere in the codebase.

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
hitanshu-dhawan: Removed code related to `Date` parsing as it is not used anywhere in the codebase.

hitanshu-dhawan:
Resolved all conversations.

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
```diff
@@ -1,19 +1,11 @@
...
-    val instance: Gson by lazy {
-        GsonBuilder()
-            .serializeNulls()
-            .setPrettyPrinting()
-            .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
-            .registerTypeAdapter(Date::class.java, DateTypeAdapter())
-            .create()
+    val instance: Moshi by lazy {
+        Moshi.Builder()
+            .build()
```
vbuberen: nit: you could do a one line here

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
```diff
@@ -1,19 +1,11 @@
 package com.chuckerteam.chucker.internal.support
 
-import com.google.gson.FieldNamingPolicy
-import com.google.gson.Gson
-import com.google.gson.GsonBuilder
-import com.google.gson.internal.bind.DateTypeAdapter
-import java.util.Date
+import com.squareup.moshi.Moshi
 
 internal object JsonConverter {
```
cortinico: Imho we could nuke `JsonConverter` at all as it's exposing its implementation details. I would rather move this `Moshi` instance to the `MoshiUtil` file and have it as a top level field.

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
```diff
@@ -1,60 +1,14 @@
...
-import java.text.DateFormat
-import java.util.Date
-import java.util.Locale
-import org.junit.jupiter.api.Test
+import org.junit.Test
 
 class JsonConverterTest {
 
     @Test
     fun testInstance_sameInstance() {
```
cortinico: This test can also go as it's not really valuable (it's testing the `by lazy` implementation of Kotlin).

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
```diff
@@ -1,60 +1,14 @@
...
-            {
-              "date": "$expectedDateString"
-            }
-            """.trimIndent()
-        )
-    }
-
-    @Test
-    fun testGsonConfiguration_willUseLowerCaseWithUnderscores() {
-        val json = JsonConverter.instance.toJson(NamingTestClass("aCamelCaseString"))
```
cortinico: Can I ask you to save this test?

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
hitanshu-dhawan: Done.

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
hitanshu-dhawan: Yes, makes sense.

Should I change the name from `JsonConverter.instance` to `moshi` as it is a `Moshi` instance? 

something like this
```
internal val moshi: Moshi by lazy {
    Moshi.Builder().build()
}
```

and usage like this
```
moshi.getHttpHeaderListJsonAdapter().toJson(headers)
```

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
hitanshu-dhawan: Agreed. Deleted `JsonConverterTest`.

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
hitanshu-dhawan: Moshi doesn't support field naming policy.
It recommends us to name the field as it appears in the JSON.
More here: https://github.com/square/moshi#borrows-from-gson

Also, nowhere in the codebase, we were using this feature provided by Gson.
We are only serializing and deserializing `List<HttpHeader>` which is already annotated with `@Json`.

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
cortinico: Yes that is definitely better.

library/src/test/java/com/chuckerteam/chucker/internal/support/JsonConverterTest.kt
cortinico: ++ we can drop this comment then

library/src/main/java/com/chuckerteam/chucker/internal/support/JsonConverter.kt
hitanshu-dhawan: Done.

vbuberen:
So, as I mentioned, we need to run benchmarks before doing such migrations. I did some for Chucker particular case - we are mostly formatting JSONs which means that we parse them and turn into pretty string later. 
I used repo from this article https://tech.olx.com/android-jetpack-benchmark-json-parsers-performance-1e76031a296b and did the following:
1. Updated all dependencies to their latest versions: 
-  Gson:2.8.6
-  Kotlinx-serialization-runtime:0.20.0
-  Jackson:2.10.1
-  Moshi:1.9.2
2. Added classes for formatting JSONs with those libraries.
3. Did some testing.

Here is data for some low-end devices:
Sony Xperia M (flashed in LineageOS 14.1 based on Android 7.1.2):
```
Connected to process 11622 on device 'sony-nicki'.

benchmark:     5,402,106 ns JsonFormatersBenchmark.benchmarkJacksonWithSample1
benchmark:    13,703,648 ns JsonFormatersBenchmark.benchmarkJacksonWithSample2
benchmark:    33,328,247 ns JsonFormatersBenchmark.benchmarkJacksonWithSample3

benchmark:     5,371,585 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample1
benchmark:    21,028,537 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample2
benchmark:    57,164,662 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample3

benchmark:     5,493,667 ns JsonFormatersBenchmark.benchmarkKotlinWithSample1
benchmark:    21,059,060 ns JsonFormatersBenchmark.benchmarkKotlinWithSample2
benchmark:    39,890,132 ns JsonFormatersBenchmark.benchmarkKotlinWithSample3

benchmark:     3,754,006 ns JsonFormatersBenchmark.benchmarkGsonFormatSample1
benchmark:    14,527,698 ns JsonFormatersBenchmark.benchmarkGsonFormatSample2
benchmark:    43,003,210 ns JsonFormatersBenchmark.benchmarkGsonFormatSample3
```

Nexus 7 2012 (flashed with custom AOSP 7.1.2):
```
benchmark:     3,101,000 ns JsonFormatersBenchmark.benchmarkJacksonWithSample1
benchmark:     9,225,000 ns JsonFormatersBenchmark.benchmarkJacksonWithSample2
benchmark:    24,258,000 ns JsonFormatersBenchmark.benchmarkJacksonWithSample3

benchmark:     4,229,000 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample1
benchmark:    16,876,000 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample2
benchmark:    43,820,000 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample3

benchmark:     5,394,000 ns JsonFormatersBenchmark.benchmarkKotlinWithSample1
benchmark:    21,359,000 ns JsonFormatersBenchmark.benchmarkKotlinWithSample2
benchmark:    38,371,000 ns JsonFormatersBenchmark.benchmarkKotlinWithSample3

benchmark:     3,315,000 ns JsonFormatersBenchmark.benchmarkGsonFormatSample1
benchmark:    10,024,000 ns JsonFormatersBenchmark.benchmarkGsonFormatSample2
benchmark:    32,726,000 ns JsonFormatersBenchmark.benchmarkGsonFormatSample3
```
I can't run on other devices available (like Samsung S9+) due to this issue: https://issuetracker.google.com/issues/142273120

You can find my source code here: https://github.com/vbuberen/android-json-benchmark/tree/master/json/src/main/java/com/mootazltaief/json

Based on such results I would like to conclude that we aren't good to migrate Chucker, because it is not cool to make performance worse even thought it might not be noticeable to users.

hitanshu-dhawan:
Thanks for the benchmarking results.
Yes, seems like Moshi didn't perform here as expected.

I also did some benchmarking for mainly two uses cases for Chucker.
1. Serialization and Deserialization
2. JSON formatting

I also added plain old `org.json` to benchmark against Gson and Moshi.

Here are the results.

#### Serialization and Deserialization
OnePlus 5 (8GB RAM, Android 9)
```
benchmark:     2,73,229 ns JsonParsersBenchmark.benchmarkGsonFromJsonSample1
benchmark:    10,79,270 ns JsonParsersBenchmark.benchmarkGsonFromJsonSample2
benchmark:    26,79,687 ns JsonParsersBenchmark.benchmarkGsonFromJsonSample3

benchmark:     3,11,302 ns JsonParsersBenchmark.benchmarkMoshiFromJsonSample1
benchmark:    12,49,010 ns JsonParsersBenchmark.benchmarkMoshiFromJsonSample2
benchmark:    39,15,833 ns JsonParsersBenchmark.benchmarkMoshiFromJsonSample3

benchmark:     5,32,864 ns JsonParsersBenchmark.benchmarkGsonToJsonSample1
benchmark:    21,03,281 ns JsonParsersBenchmark.benchmarkGsonToJsonSample2
benchmark:    57,04,480 ns JsonParsersBenchmark.benchmarkGsonToJsonSample3

benchmark:     1,85,182 ns JsonParsersBenchmark.benchmarkMoshiToJsonSample1
benchmark:     7,46,667 ns JsonParsersBenchmark.benchmarkMoshiToJsonSample2
benchmark:    19,49,583 ns JsonParsersBenchmark.benchmarkMoshiToJsonSample3
```

#### JSON formatting
OnePlus 5 (8GB RAM, Android 9)
```
benchmark:     4,57,760 ns JsonFormatersBenchmark.benchmarkGsonFormatSample1
benchmark:    18,86,719 ns JsonFormatersBenchmark.benchmarkGsonFormatSample2
benchmark:    57,79,844 ns JsonFormatersBenchmark.benchmarkGsonFormatSample3

benchmark:     7,47,761 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample1
benchmark:    28,36,823 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample2
benchmark:    79,03,593 ns JsonFormatersBenchmark.benchmarkMoshiFormatSample3

benchmark:     6,65,573 ns JsonFormatersBenchmark.benchmarkOrgJsonFormatSample1
benchmark:    26,59,375 ns JsonFormatersBenchmark.benchmarkOrgJsonFormatSample2
benchmark:    64,75,729 ns JsonFormatersBenchmark.benchmarkOrgJsonFormatSample3
```

You can find the source code [here](https://bitbucket.org/hitanshu-dhawan/android-json-benchmark).

Conclusion
1. Serialization and Deserialization
Serialization is faster in case of Moshi and Deserialization is faster in case of Gson.
1. JSON formatting
JSON formatting is faster in case of Gson.

It seems like there are pros and cons for both Gson and Moshi, and it's difficult to decide a clear winner between them.
So, feel free to close this PR if you are not in favor of migration to Moshi.

PR #274, opened 2020-03-15
Remove redundant JvmName

vbuberen:
## :page_facing_up: Context
Removing outdated line with JVM Name.


cortinico:
Just adding it to https://github.com/ChuckerTeam/chucker/milestone/6 as this will show up if we run `japicmp` on the upcoming release. Nothing to worry about though

PR #275, opened 2020-03-17
UX improvements

vbuberen:
## :camera: Screenshots
| Before | After |
| --- | --- |
| Screenshot 2020-03-17 at 13 06 41 | Screenshot 2020-03-17 at 13 07 20 |
Screenshot 2020-03-17 at 12 43 30

## :page_facing_up: Context
Chucker had some issues with its `secondary` color (or is it was called at AppCompat times `accent`) - it was hard to read when it is on a white background due to contrast level. 
Check on some websites confirms the issue: https://contrastchecker.com/?c=69f0ae&b=ffffff
So I updated the color to be more readable while staying bright enough: https://contrastchecker.com/?c=009E09&b=ffffff
While it is still not passing every check it is much better than the previous one (check attached screenshots).
Also, I thought that for non https requests we shouldn't just hide the lock icon, but show red open lock icon instead to attract attention to such requests. Yeah, for devices with Pie and higher having http is harder, but I still see some projects with added `cleartextTrafficPermitted=true` line. It might be a good idea to indicate insecure requests instead of just showing nothing.

## :pencil: Changes
- Updated secondary color both for day and night themes to be more contrast.
- Added an open lock icon to show it when the request isn't https.


cortinico:
> Also, I thought that for non https requests we shouldn't just hide the lock icon, but show red open lock icon instead to attract attention to such requests. Yeah, for devices with Pie and higher having http is harder, but I still see some projects with added cleartextTrafficPermitted=true line. It might be a good idea to indicate insecure requests instead of just showing nothing.

++ here

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionAdapter.kt
```diff
@@ -60,7 +63,22 @@ internal class TransactionAdapter internal constructor(
...
+                val sslDrawableColorRes = if (transaction.isSsl) {
+                    R.color.chucker_color_primary
+                } else {
+                    R.color.chucker_color_error
+                }
+                ssl.setImageDrawable(AppCompatResources.getDrawable(itemView.context, sslDrawableRes))
+                ImageViewCompat.setImageTintList(
+                    ssl,
+                    ColorStateList.valueOf(ContextCompat.getColor(itemView.context, sslDrawableColorRes))
+                )
```
cortinico: nit: this could maybe be refactored to a sealed class to store the protocol logic (rather than having two similar if-then-else), something like?

```kotlin
sealed class Protocol(val icon : Int, val color: Int) {
        class Http() : Protocol(R.drawable.chucker_ic_http, R.color.chucker_color_primary)
        class Https() : Protocol(R.drawable.chucker_ic_https, R.color.chucker_color_primary)
        
        fun populateImage(imageDrawable: Drawable) {
            ...
        }
    }
```

vbuberen: I did it in slightly other way, so the sealed class has only resources without function to populate image to not mix responsibilities. Image populating is still up to adapter class.

vbuberen: I did it in slightly other way, so the sealed class has only resources without function to populate image to not mix responsibilities. Image populating is still up to adapter class.

Issue #273, opened 2020-03-12
Highlight time consuming API's?  time being configurable?


adb-shell:
Will it helpful to mark the API's(in a different color similar to failed API calls) that take more than x time(configurable) for responding?

NULLPointerGuy:
Opened a PR for the changes:https://github.com/ChuckerTeam/chucker/pull/282

vbuberen:
Closing due to #286

PR #282, opened 2020-03-20
identify time consuming api with time being configurable

NULLPointerGuy:
## :camera: Screenshots
darkmode
normalmode

## :page_facing_up: Context
Identify time-consuming API's in a different color with time being configurable.
issue: https://github.com/ChuckerTeam/chucker/issues/273

## :no_entry_sign: Breaking
Nothing breaking is expected

Just set `normalAPIDuration` any API taking more than normalAPIDuration time will be highlighted in purple color, as shown in the screenshot.

### DIFF
--- a/library-no-op/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library-no-op/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -12,7 +12,8 @@ class ChuckerInterceptor @JvmOverloads constructor(
     context: Context,
     collector: Any? = null,
     maxContentLength: Any? = null,
-    headersToRedact: Any? = null
+    headersToRedact: Any? = null,
+    normalAPIDuration: Any? = null
 ) : Interceptor {
 
     fun redactHeaders(vararg names: String): ChuckerInterceptor {
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -29,12 +29,15 @@ private const val MAX_BLOB_SIZE = 1000_000L
  * results.
  * @param headersToRedact a [Set] of headers you want to redact. They will be replaced
  * with a `**` in the Chucker UI.
+ * @param normalAPIDuration normal expected duration of api calls in millisecond, exceeding which it
+ * will be marked as slow api call.
  */
 class ChuckerInterceptor @JvmOverloads constructor(
     private val context: Context,
     private val collector: ChuckerCollector = ChuckerCollector(context),
     private val maxContentLength: Long = 250000L,
-    headersToRedact: Set<String> = emptySet()
+    headersToRedact: Set<String> = emptySet(),
+    private val normalAPIDuration: Long = -1L
 ) : Interceptor {
 
     private val io: IOUtils = IOUtils(context)
@@ -139,6 +142,12 @@ class ChuckerInterceptor @JvmOverloads constructor(
             responseContentLength = response.contentLength
 
             tookMs = (response.receivedResponseAtMillis() - response.sentRequestAtMillis())
+
+            tookMs?.let { tookMs ->
+                if (normalAPIDuration > -1L && tookMs > normalAPIDuration) {
+                    isSlowApiCall = true
+                }
+            }
         }
 
         return if (responseEncodingIsSupported) {
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
@@ -48,7 +48,8 @@ internal class HttpTransaction(
     @ColumnInfo(name = "responseHeaders") var responseHeaders: String?,
     @ColumnInfo(name = "responseBody") var responseBody: String?,
     @ColumnInfo(name = "isResponseBodyPlainText") var isResponseBodyPlainText: Boolean = true,
-    @ColumnInfo(name = "responseImageData") var responseImageData: ByteArray?
+    @ColumnInfo(name = "responseImageData") var responseImageData: ByteArray?,
+    @ColumnInfo(name = "isSlowApiCall") var isSlowApiCall: Boolean = false
 ) {
 
     @Ignore
@@ -262,6 +263,7 @@ internal class HttpTransaction(
             (responseHeaders == other.responseHeaders) &&
             (responseBody == other.responseBody) &&
             (isResponseBodyPlainText == other.isResponseBodyPlainText) &&
-            (responseImageData?.contentEquals(other.responseImageData ?: byteArrayOf()) != false)
+            (responseImageData?.contentEquals(other.responseImageData ?: byteArrayOf()) != false) &&
+            (isSlowApiCall == other.isSlowApiCall)
     }
 }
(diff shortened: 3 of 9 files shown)


library-no-op/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -12,6 +12,7 @@ class ChuckerInterceptor @JvmOverloads constructor(
     context: Context,
     collector: Any? = null,
     maxContentLength: Any? = null,
+    normalAPIDuration: Any? = null,
```
cortinico: This should be the last parameters

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -29,11 +29,14 @@ private const val MAX_BLOB_SIZE = 1000_000L
...
  * @param headersToRedact a [Set] of headers you want to redact. They will be replaced
  * with a `**` in the Chucker UI.
+ * @param normalAPIDuration normal expected duration of api calls in millisecond, exceeding which it
+ * will be marked as slow api call.
  */
 class ChuckerInterceptor @JvmOverloads constructor(
     private val context: Context,
     private val collector: ChuckerCollector = ChuckerCollector(context),
     private val maxContentLength: Long = 250000L,
+    private val normalAPIDuration: Long = -1L,
```
cortinico: This should be the last parameters

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -48,7 +48,8 @@ internal class HttpTransaction(
     @ColumnInfo(name = "responseHeaders") var responseHeaders: String?,
     @ColumnInfo(name = "responseBody") var responseBody: String?,
     @ColumnInfo(name = "isResponseBodyPlainText") var isResponseBodyPlainText: Boolean = true,
-    @ColumnInfo(name = "responseImageData") var responseImageData: ByteArray?
+    @ColumnInfo(name = "responseImageData") var responseImageData: ByteArray?,
+    @ColumnInfo(name = "isSlowApiCall") var isSlowApiCall: Boolean = false
```
cortinico: You don't need to store this boolean. This can easily be inferred from the `tookMs` field.

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransactionTuple.kt
```diff
@@ -21,7 +21,8 @@ internal class HttpTransactionTuple(
     @ColumnInfo(name = "responseCode") var responseCode: Int?,
     @ColumnInfo(name = "requestContentLength") var requestContentLength: Long?,
     @ColumnInfo(name = "responseContentLength") var responseContentLength: Long?,
-    @ColumnInfo(name = "error") var error: String?
+    @ColumnInfo(name = "error") var error: String?,
+    @ColumnInfo(name = "isSlowApiCall") var isSlowApiCall: Boolean = false
```
cortinico: Same as above

library/src/main/res/values/colors.xml
```diff
@@ -20,5 +20,7 @@
     <color name="chucker_status_300">#0D47A1</color>
 
     <color name="chucker_background_span_color">#ffffff00</color>
-    <color name="chucker_foreground_span_color">#ffff0000</color>
+    <color name="chucker_foreground_span_color">#FF5900</color>
```
cortinico: Was this intentional?

NULLPointerGuy: Sorry, that was not intentional.

library-no-op/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
NULLPointerGuy: updated

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
NULLPointerGuy: updated

library/src/main/res/values/colors.xml
NULLPointerGuy: sorry, that was not intentional hence reverted.

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
NULLPointerGuy: Since we are not storing the `normalAPIDuration` i believe we need this boolean field.
Or are you suggesting we save `normalAPIDuration` instead of the boolean?

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransactionTuple.kt
NULLPointerGuy: Since we are not storing the `normalAPIDuration` i believe we need this boolean field.
Or are you suggesting we save `normalAPIDuration` instead of the boolean?

cortinico:
Answering here:

> Since we are not storing the normalAPIDuration i believe we need this boolean field. Or are you suggesting we save normalAPIDuration instead of the boolean?

We should store neither of those two fields. I think the easiest approach would be to add a mutable property inside the `Chucker` object. You don't need this configuration inside the `ChuckerInterceptor`.

Then you can simply extend `HttpTransaction` with something like:

```kotlin
val isSlowApiCall : Boolean 
        get() = Chucker.normalApiDuration > -1L && tookMs > Chucker.normalApiDuration
```

And use the new field in the `TransactionAdapter` to properly color the text.

vbuberen:
I don't think that we need this feature at the moment and wouldn't like to merge it at this stage of library development.
It is not something that useful, since often this `slow` will depend on network quality.
The implementation with `isSlowAPI` field in `Transaction` model isn't good and should be avoided. I also don't like the idea of introducing something for this purpose inside Chucker object.

Additionally, it is not cool that we are introducing the new color for list items here, since it makes UX mush worse by being not obvious (e.g. for red color we now that it is a failure, but what is purple? It makes user do some additional search for color meaning).

Anyway, if there would be have more people asking this feature, I would be happy to get back. 
As for now - don't see much profit in this feature.

NULLPointerGuy:
@vbuberen we can always change the color that indicates the API is taking a long time than expected.
However, if this feature itself does not make sense at this point in time for the library, then I agree we should wait until more people request for the feature.

cortinico:
> Additionally, it is not cool that we are introducing the new color for list items here, since it makes UX mush worse by being not obvious (e.g. for red color we now that it is a failure, but what is purple? It makes user do some additional search for color meaning).

Agree on this point here.

As for the rest, I don't have a strong opinion on this feature to be honest.
I'm ok having the PR opened for now, and have people commenting if they're interested in having it merged 👍

vbuberen:
Yes, we need more opinions from users about it.
In general, after doing major updates to interceptor and DB entities I thought about creating some filters to easily sort transactions. Like some view with attributes like just GET items or taking more than specified time etc. (I would say something like online shops have in their filtering)
And it can be done mostly on UI, without requirements to modify entities with additional fields like it is done here.

NULLPointerGuy:
Interesting! that makes more sense. I would be very much interested in it.

vbuberen:
Closing due to planned #286

PR #281, opened 2020-03-19
Add tests to format utils

adammasyk:
## :page_facing_up: Context

This is my first PR. I want to help in Chucker development.
## :pencil: Changes

Add some tests to FomatUtils class to improve test coverage
## :no_entry_sign: Breaking

No
## :stopwatch: Next steps

Add more tests in next PRs! 😄

PR #285, opened 2020-03-23
format utils test refactor

adammasyk:
## :page_facing_up: Context
Tests

## :pencil: Changes

- refactor FromatUtilsSharedText: move common parts to method

- change name of two xml tests in FormatUtilsTest to be consistent with other tests in this class

## :paperclip: Related PR
https://github.com/ChuckerTeam/chucker/pull/281

## :no_entry_sign: Breaking
No

### DIFF
--- a/library/src/test/java/com/chuckerteam/chucker/internal/support/FormatUtilsSharedTextTest.kt
+++ b/library/src/test/java/com/chuckerteam/chucker/internal/support/FormatUtilsSharedTextTest.kt
@@ -31,17 +31,20 @@ class FormatUtilsSharedTextTest {
 
     @Test
     fun getShareTextForGetTransaction() {
-        Truth.assertThat(
-            FormatUtils.getShareText(contextMock, TestTransactionFactory.createTransaction("GET"), false)
-        )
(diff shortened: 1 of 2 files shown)


library/src/test/java/com/chuckerteam/chucker/internal/support/FormatUtilsSharedTextTest.kt
```diff
@@ -31,17 +31,20 @@ class FormatUtilsSharedTextTest {
...
+    }
+
+    private fun testSharedText(expected: String, method: String) {
+        val sharedText = FormatUtils.getShareText(
+            contextMock,
+            TestTransactionFactory.createTransaction(method),
+            false
         )
-            .isEqualTo(TestTransactionFactory.expectedHttpPostTransaction)
+        Truth.assertThat(sharedText).isEqualTo(expected)
```
vbuberen: Nice, but I could you move assertion back to `@Test` blocks? Because assertion is usually done inside fun with `@Test` annotation.

adammasyk: Done

PR #288, opened 2020-03-26
Properly handle unexhausted network responses

### LINKED ISSUE #287: Chucker does not recognise partially read responses

When only some portion of the upstream is read from the response it might fail when Chucker tries to read it back for processing. For example `GzipSource` expects that after it is done with the content there will be CRC32 checksum in the footer. If it is not present it will throw an exception and crash.

## :bomb: Steps to reproduce

Creating an example is doable but it is reported by @oakkitten in a follow-up discussion to #263.

## :wrench: Expected behavior

Chucker dismisses incomplete responses. This behaviour is consistent with maximum content length allowed.

MiSikora:
## :page_facing_up: Context

There is an issue #287

## :pencil: Changes

`TeeSource` now reports partially read responses as failures. This way `ChuckerInterceptor` does not try to process them. The bytes are still available in the side channel file if in the future we'd like to process these bytes.

I also wrapped reading from file in a `try/catch` expression for a good measure because of the volatile nature of a phone storage.

## :no_entry_sign: Breaking

No.

Unit tests are written. @oakkitten might confirm whether this fixes their issue using Chucker version from PR via JitPack.

## :stopwatch: Next steps

Closes #287.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
@@ -27,6 +27,7 @@ internal class TeeSource(
     private var totalBytesRead = 0L
     private var reachedLimit = false
     private var upstreamFailed = false
+    private var upstreamConsumed = false
 
     override fun read(sink: Buffer, byteCount: Long): Long {
         val bytesRead = try {
@@ -37,6 +38,7 @@ internal class TeeSource(
         }
 
         if (bytesRead == -1L) {
+            upstreamConsumed = true
             sideStream.close()
             return -1L
         }
@@ -60,8 +62,10 @@ internal class TeeSource(
     override fun close() {
         sideStream.close()
         upstream.close()
-        if (!upstreamFailed) {
+        if (upstreamConsumed) {
             callback.onSuccess(sideChannel)
+        } else {
+            callSideChannelFailure(IOException("Upstream was not fully consumed"))
         }
     }
 
@@ -83,7 +87,10 @@ internal class TeeSource(
         /**
          * Called when there was an issue while copying bytes to the [file].
          *
-         * It might occur due to an exception thrown while reading bytes or due to exceeding capacity limit.
+         * It might occur due to one of the following reasons:
+         * - an exception was thrown while reading bytes
+         * - capacity limit was exceeded
+         * - upstream was not fully consumed
          */
         fun onFailure(exception: IOException, file: File)
     }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
@@ -15,7 +15,8 @@ import okio.Timeout
  * to a [sideChannel] file. After the [upstream] is depleted or when a failure occurs
  * an appropriate [callback] method is called.
  *
- * Failure is considered any [IOException] during reading the bytes or exceeding [readBytesLimit] length.
+ * Failure is considered any [IOException] during reading the bytes,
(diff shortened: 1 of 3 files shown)


MiSikora:
@oakkitten Could check if this PR fixes the issue that you experience? You can try with a version from JitPack - https://jitpack.io/#MiSikora/chucker/corrupted-files-read-fix-SNAPSHOT.

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -27,6 +27,7 @@ internal class TeeSource(
     private var totalBytesRead = 0L
     private var reachedLimit = false
     private var upstreamFailed = false
+    private var upstreamConsumed = false
```
vbuberen: Can we change the name here to something more usual for boolean variables - `isUpstreamConsumed`?
Also, just saw that 2 variables above could also be renamed.

oakkitten:
the exception is gone, but there's another problem which is probably closely related to #209

it seems that in order for Chucker to display the transferred contents, it has to finish reading the input stream, and by finishing reading the stream i mean that `reader.read()` should return `-1`. if i close the stream when it has been only partially read, or if it was read fully but the trailing `read()` has not been called, the response body remains empty.

it would be nice—for clarity—to never have the body part of the response tab empty, or at least only have it empty only if there can't be a body, as with a response to `HEAD`. consider having “no body” or “waiting for the app to read the body...” there

p.s. i'm using Chucker as network interceptor

MiSikora:
But what would you expect to see if bytes are missing from the body?

oakkitten:
something along...
	
        <html>
	        <head>
	            <title>nice titl

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;*only a part of the document has been read*

MiSikora:
As I mentioned, this gets really ugly to handle very quickly when you consider things that I mentioned. I mean, you can create an issue and give it a try if there is a need for it.

oakkitten:
i plugged Chucker as an application level interceptor and it still doesn't seem to be displaying the body... application level interceptors aren't dealing with compression, so these should be able to display partial content just fine, shouldn't they? and even if it's compressed, isn't it easy unzipping on the fly?

even if all this is hard to accomplish, i still think that some message is necessary. empty body is just confusing

i will gladly open another issue when this is merged

p.s. thanks for all the the work! 🙇‍♀️

MiSikora:
Yes, it wouldn't work for application level interceptor as well. I mentioned network level interceptor because this is where most of the compression happens and it makes it that much harder.

Parsing plain text body wouldn't be too bad. We would just read it as as a string and be done with that. Formatting it gets much trickier because most of the formatters require properly formed bodies in order to format it. They are lenient to some point but we can't be really what will be handled by them. I might be wrong here and maybe there are solutions that can handle this already.

But I 100% agree about some feedback message when body was not fully read or exceeded capacity limit.

vbuberen:
@oakkitten 
Regarding empty state - you could check our list of issues first, since it is already there.
As to your suggestion with partially read body - you are welcome to implement it. 

@MiSikora 
Thanks for your PR. I am merging it and let's finish #250 to do a release before moving further with other stuff.

oakkitten:
@MiSikora i see, thanks for your input!

@vbuberen wow ok. so sorry i didn't realize that the issue of missing body in the response tab, which in my tests seemed to be directly tied to partially fetched data cases, is an existing issue. and i don't know kotlin so sadly i won't be able to contribute to Chucker's codebase in the nearest future.

vbuberen:
In such case I would like to ask you to fill a feature request with as much info as possible to speed up its consideration and possible implementation.

PR #294, opened 2020-03-27
adding back tick to make tests more readable?

NULLPointerGuy:
Would it make sense to add backticks to the test names to make it more readable and get rid of underscores?
One added advantage is that the test name itself can be a one-liner description of the test.

I would love to get your input on this.

### DIFF
--- a/library/src/test/java/com/chuckerteam/chucker/internal/support/FormatUtilsTest.kt
+++ b/library/src/test/java/com/chuckerteam/chucker/internal/support/FormatUtilsTest.kt
@@ -12,7 +12,7 @@ class FormatUtilsTest {
     )
 
     @Test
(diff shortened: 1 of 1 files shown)


cortinico:
I have to say I'm not a super fan of backticks, but I can live we it if we have strong arguments. I would just suggest that, if we end up with backticks, we stick to the following rule:

* All lowercase
* No `test` at the beginning of the name

like:

```kotlin
fun `format json with null values`() {
```

vbuberen:
I am not sure here too. Yeah, tests names look better this way, but having backticks confuses me as well.

NULLPointerGuy:
cool, I just wanted to get an opinion on this, hence I will be closing the PR based on the inputs received.

PR #300, opened 2020-04-03
Revert "Add selective interception. (#279)"

cortinico:
This reverts commit d14ed642b5bcff55e08b7976bf5701a0cc45f585.

## :page_facing_up: Context
A lot of the context for this revert is in those two comments: https://github.com/ChuckerTeam/chucker/pull/279#discussion_r395950647
https://github.com/ChuckerTeam/chucker/pull/279#discussion_r395950704

tl;dr: We really liked this feature developed by @MiSikora, but we need to find a better way to include it to make sure we don't bleed inside the `library-no-op`.

As of now we felt that the right thing to do is probably reverting the PR before it lands on a new release (v3.2.0 is coming soon) and discuss a bit more on how to re-apply the proposed change.

## :pencil: Changes
Reverting #279 + I had to resolve some merge conflicts so this probably needs a review.

## :paperclip: Related PR
#279

n/a

## :stopwatch: Next steps
I'll open a Issue where we can discuss further how to re-apply this change.


library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -81,7 +83,7 @@ class ChuckerInterceptorTest {
         val client = factory.create(chuckerInterceptor)
         val responseBody = client.newCall(request).execute().body()!!.source().readByteString()
 
-        assertThat(responseBody).isEqualTo(expectedBody)
+        assertEquals(expectedBody, responseBody)
```
MiSikora, line 86: This revert seems unnecessary.

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -98,8 +100,8 @@ class ChuckerInterceptorTest {
         client.newCall(request).execute().readByteStringBody()
         val transaction = chuckerInterceptor.expectTransaction()
 
-        assertThat(transaction.isResponseBodyPlainText).isTrue()
-        assertThat(transaction.responseBody).isEqualTo("Hello, world!")
+        assertTrue(transaction!!.isResponseBodyPlainText)
+        assertEquals("Hello, world!", transaction!!.responseBody)
```
MiSikora, line 104: Same as above.

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -115,94 +117,7 @@ class ChuckerInterceptorTest {
...
-        val request = Request.Builder().url(serverUrl)
-            .addHeader(Chucker.SKIP_INTERCEPTOR_HEADER_NAME, "false")
-            .build()
-
-        val client = factory.create(chuckerInterceptor)
-        var response = client.newCall(request).execute()
-        if (factory == ClientFactory.NETWORK) response = response.networkResponse()!!
-
-        assertThat(response.request().header(Chucker.SKIP_INTERCEPTOR_HEADER_NAME)).isNull()
+        assertEquals("Hello, world!", responseBody.utf8())
```
MiSikora, line 120: Same as above.

cortinico:
> I would just reopen #266 instead of creating a new issue if that is your intention. :)

Sure, let's do that 👍 

> This might be the case for that bot I suggested in Slack:
> [github.com/marketplace/actions/automatic-revert](https://github.com/marketplace/actions/automatic-revert) 😄
> 
> The review made by @MiSikora is totally legit. Please update to Truth assertions.

Yeah the problem is that this revert had conflicts that I had to manually resolve. I don't know how much a bot could help in such scenario.

PR #152, opened 2019-10-30
Add ability to export the list of transactions

matthewcmckenna:
## :camera: Screeshots
*Show us what you've changed, we love images.*

[Demo Video](https://youtu.be/FepIzHvDRN8)

## :page_facing_up: Context
*Why did you changed something? Is there an [issue](https://github.com/ChuckerTeam/chucker/issues) to link here? Or an external link?*

I am looking to provide a potential fix to  #91 

## :pencil: Changes
*Which code did you change? How?*

- Added some strings to `strings.xml`
- Added a menu option to `chucker_transactions_list.xml`
- Added a share function to `TransactionListFragment.kt`
- Added logic to show the `AlertDialog` in `TransactionListFragment.kt`
- Added functions to get all transactions in a `List`

## :paperclip: Related PR
*PR that blocks this one, or the ones blocked by this PR*

None

## :warning: Breaking
*Is there something breaking in the API? Any class or method signature changed?*

No

## :pencil: How to test
*Is there a special case to test your changes?*

No special cases. Just try to share!

## :crystal_ball: Next steps
*Do we have to plan something else after the merge?*

No

PR #305, opened 2020-04-04
Ability to export the list of transactions

### LINKED ISSUE #91: Provide a way to query / get logs
In Chuck the only way to get a log of the requests was through reflection for the Db open helper.
It would be very useful to add a way to obtain the records to export a trace

Add to Chucker a method to query the logs and get a list of http or error logs.

adb-shell:
Branched from PR:https://github.com/ChuckerTeam/chucker/pull/152
Since the above PR has not been updated for 5 months, I took the liberty to branch out raise new PR.

There are a few changes when compared to PR:https://github.com/ChuckerTeam/chucker/pull/152
1. instead of converting all transactions to one `string` and sharing it via intents, I have created a cache file(where I dump all the transactions to be exported), sharing it via FileProvider.
to potentially avoid `transaction too large exception`

2. added test for `ShareUtils`

Closes #91


library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -79,20 +97,52 @@ internal class TransactionListFragment :
...
+                    ) { _, _ ->
+                        viewModel.clearTransactions()
+                    }
+                    .setNegativeButton(R.string.chucker_cancel, null)
+                    .show()
+                true
+            }
+            R.id.export -> {
+                if (::transactions.isInitialized && transactions.isNotEmpty()) {
+                    AlertDialog.Builder(requireContext())
```
vbuberen, line 116: Please move this dialog into its own function as the above one.

library/src/main/res/values/strings.xml
```diff
@@ -47,8 +49,13 @@
     <string name="chucker_share_throwable_content"><![CDATA[Date: %1$s\nException: %2$s\nTag: %3$s\nMessage: %4$s\n\n%5$s]]></string>
     <string name="chucker_clear_http_confirmation">Do you want to clear complete network calls history?</string>
     <string name="chucker_clear_throwable_confirmation">Do you want to clear complete throwables history?</string>
+    <string name="chucker_export_http_confirmation">Export all http calls</string>
```
vbuberen, line 52: Screenshot 2020-04-05 at 16 53 45
This dialog seems not that user friendly and needs some texts updates.

```suggestion
    <string name="chucker_export_http_confirmation">Do you want to export all network transactions?</string>
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -79,20 +97,52 @@ internal class TransactionListFragment :
...
+                        .setPositiveButton(R.string.chucker_export) { _, _ ->
+                            val file = cacheFileFactory.createExportFile()
+                            file.writeText(ShareUtils.getStringFromTransactions(transactions, requireContext()))
+
+                            val uri = FileProvider.getUriForFile(
+                                requireContext(),
+                                getString(R.string.chucker_provider_authority), file
+                            )
+
+                            val intent = Intent().apply {
```
vbuberen, line 128: Please switch to the `ShareCompat.IntentBuilder` with chooser like it is done in `TransactionActivity` or `ThrowableActivity`.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -63,6 +75,12 @@ internal class TransactionListFragment :
                     if (transactionTuples.isEmpty()) View.VISIBLE else View.GONE
             }
         )
+        RepositoryProvider.transaction().getAllTransactions().observe(
+            viewLifecycleOwner,
+            Observer {
```
vbuberen, line 80: For this particular case we can just return a list of items once without `LiveData` and observers.
Like when user clicks on `Export` we just call `viewModel.getAllTransactions()` and that is it. There is no need in maintaining LiveData object, since user fetches items only once for his export action.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -79,20 +97,52 @@ internal class TransactionListFragment :
...
+                        R.string.chucker_clear
+                    ) { _, _ ->
+                        viewModel.clearTransactions()
+                    }
+                    .setNegativeButton(R.string.chucker_cancel, null)
+                    .show()
+                true
+            }
+            R.id.export -> {
+                if (::transactions.isInitialized && transactions.isNotEmpty()) {
```
vbuberen, line 115: I would remove `lateinit var` for transactions and just stick with the logic I suggested in the comment above where I mentioned that we can just fetch items once when user wants to export for sure.

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -51,6 +51,10 @@ class ChuckerInterceptorTest {
             override fun create(): File {
                 return File(tempDir, "testFile")
             }
+
+            override fun createExportFile(): File {
```
vbuberen, line 55: Nit: it is not a big issue, since Chucker is a debug tool and the file is small, but in the scenario you presented we are creating the file and not removing it after sharing. We can leave it as it is for now, but it is some place for improvement.

library/src/main/java/com/chuckerteam/chucker/internal/support/AndroidCacheFileFactory.kt
```diff
@@ -10,7 +10,14 @@ internal class AndroidCacheFileFactory(
     private val fileDir = context.cacheDir
     private val uniqueIdGenerator = AtomicLong()
 
-    override fun create(): File {
-        return File(fileDir, "chucker-${uniqueIdGenerator.getAndIncrement()}")
+    override fun create(): File = File(fileDir, "chucker-${uniqueIdGenerator.getAndIncrement()}")
```
vbuberen, line 13: If you are adding `createExportFile` please provide some meaningful name for this fun as well.

library/src/main/res/xml/provider_paths.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<paths xmlns:android="http://schemas.android.com/apk/res/android">
```
vbuberen, line 2: `android` is a redundant namespace declaration here.

library/src/test/java/com/chuckerteam/chucker/internal/support/ShareUtilsTest.kt
```diff
@@ -0,0 +1,56 @@
...
+        every { getString(R.string.chucker_export_separator) } returns "=================="
+        every { getString(R.string.chucker_export_prefix) } returns "/* Export Start */"
+        every { getString(R.string.chucker_export_postfix) } returns "/*  Export End  */"
+    }
+
+    @Test
+    fun isStringFromTransactionsListValid() {
+        val transactionList = ArrayList<HttpTransaction>()
+
+        for (i in 0 until 10)
```
cortinico, line 40: nit: 
```suggestion
        repeat(10) {
```

library/src/test/java/com/chuckerteam/chucker/internal/support/ShareUtilsTest.kt
```diff
@@ -0,0 +1,56 @@
...
+            transactionList.add(TestTransactionFactory.createTransaction(getRandomHttpMethod()))
+
+        val actualStringFromTransactions = ShareUtils.getStringFromTransactions(transactionList, contextMock)
+        val expectedStringFromTransactions = transactionList.joinToString(
+            separator = "\n${contextMock.getString(R.string.chucker_export_separator)}\n",
+            prefix = "${contextMock.getString(R.string.chucker_export_prefix)}\n",
+            postfix = "\n${contextMock.getString(R.string.chucker_export_postfix)}\n"
```
cortinico, line 47: Here you're constructing the expected string, you don't need to call `contextMock.getString` but just copy over the string.

library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.chuckerteam.chucker.R
+import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
+
+internal object ShareUtils {
+    fun getStringFromTransactions(transactions: List<HttpTransaction>, context: Context): String {
+        return transactions.joinToString(
+            separator = "\n${context.getString(R.string.chucker_export_separator)}\n",
+            prefix = "${context.getString(R.string.chucker_export_prefix)}\n",
+            postfix = "\n${context.getString(R.string.chucker_export_postfix)}\n"
+        ) { "\n${FormatUtils.getShareText(context, it, false)}\n" }
```
cortinico, line 13:
```suggestion
        ) { FormatUtils.getShareText(context, it, false) }
```

I'm not sure you need all those `\n`

library/src/main/java/com/chuckerteam/chucker/internal/support/AndroidCacheFileFactory.kt
```diff
@@ -10,7 +10,14 @@ internal class AndroidCacheFileFactory(
...
+
+    override fun createExportFile(): File {
+        val file = File(fileDir, "transactions")
+        if (file.exists()) {
+            file.delete()
+        }
+        file.createNewFile()
+        return file
```
cortinico, line 21: You can be a bit more idiomatic like:

```kotlin
override fun createExportFile() = File(fileDir, "transactions").apply { 
        if (exists()) {
            delete()
        }
        createNewFile()
    }
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -63,6 +75,12 @@ internal class TransactionListFragment :
                     if (transactionTuples.isEmpty()) View.VISIBLE else View.GONE
             }
         )
+        RepositoryProvider.transaction().getAllTransactions().observe(
```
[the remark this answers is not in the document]
adb-shell: updated

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -79,20 +97,52 @@ internal class TransactionListFragment :
...
-            AlertDialog.Builder(requireContext())
-                .setTitle(R.string.chucker_clear)
-                .setMessage(R.string.chucker_clear_http_confirmation)
-                .setPositiveButton(
-                    R.string.chucker_clear
-                ) { _, _ ->
-                    viewModel.clearTransactions()
+        return when (item.itemId) {
+            R.id.clear -> {
+                AlertDialog.Builder(requireContext())
```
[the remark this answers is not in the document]
adb-shell: updated

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -79,20 +97,52 @@ internal class TransactionListFragment :
...
+                true
+            }
+            R.id.export -> {
+                if (::transactions.isInitialized && transactions.isNotEmpty()) {
+                    AlertDialog.Builder(requireContext())
+                        .setTitle(R.string.chucker_export)
+                        .setMessage(R.string.chucker_export_http_confirmation)
+                        .setPositiveButton(R.string.chucker_export) { _, _ ->
+                            val file = cacheFileFactory.createExportFile()
+                            file.writeText(ShareUtils.getStringFromTransactions(transactions, requireContext()))
```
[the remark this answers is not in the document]
adb-shell: agreed, have updated.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -79,20 +97,52 @@ internal class TransactionListFragment :
...
+                                putExtra(Intent.EXTRA_STREAM, uri)
+                                addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+                            }
+                            startActivity(intent)
+                        }
+                        .setNegativeButton(R.string.chucker_cancel, null)
+                        .show()
+                } else {
+                    view?.let {
+                        Snackbar.make(it, R.string.chucker_export_empty_text, Snackbar.LENGTH_SHORT).show()
```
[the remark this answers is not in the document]
adb-shell: sure, updated.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
adb-shell: moved all dialogs to common utils called `DialogUtils`

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
adb-shell: agreed hence updated.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
adb-shell: updated

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
adb-shell: yes will keep in mind for further improvements.

library/src/main/java/com/chuckerteam/chucker/internal/support/AndroidCacheFileFactory.kt
adb-shell: updated

library/src/main/res/xml/provider_paths.xml
adb-shell: updated

library/src/test/java/com/chuckerteam/chucker/internal/support/ShareUtilsTest.kt
adb-shell: updated

library/src/test/java/com/chuckerteam/chucker/internal/support/ShareUtilsTest.kt
adb-shell: updated

library/src/main/java/com/chuckerteam/chucker/internal/support/AndroidCacheFileFactory.kt
adb-shell: updated

library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
adb-shell: agreed, hence updated.

library/src/main/java/com/chuckerteam/chucker/internal/support/FileFactory.kt
```diff
@@ -3,5 +3,6 @@ package com.chuckerteam.chucker.internal.support
 import java.io.File
 
 internal interface FileFactory {
-    fun create(): File
+    fun createFileForResponseBody(): File
+    fun createFileForExport(): File
```
MiSikora, line 7: I would be against such methods. They are too specific and couple responsibilities of separate functionalities. What if there is a need for more file types? Add next method? I think it would be better to provide another implementation of the previous interface and use it where it is needed.

cortinico: Agree. Maybe @adb-shell you can reuse the `create()` method and accept a `fileName : String` as input?

library/src/main/java/com/chuckerteam/chucker/internal/support/FileFactory.kt
```diff
@@ -3,5 +3,5 @@ package com.chuckerteam.chucker.internal.support
 import java.io.File
 
 internal interface FileFactory {
-    fun create(): File
+    fun create(filename: String = ""): File
```
cortinico, line 6:
```suggestion
    fun create(filename: String): File
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
```diff
@@ -6,15 +6,22 @@ import androidx.lifecycle.MutableLiveData
...
+import java.io.File
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
 internal class MainViewModel : ViewModel() {
 
     private val currentFilter = MutableLiveData<String>("")
+    private val exportFileName = "transactions.txt"
```
cortinico, line 24: Please move this inside `AndroidCacheFileFactory` as a top level `const val` as it's not really relate to the `MainViewModel`

library/src/main/java/com/chuckerteam/chucker/internal/support/FileFactory.kt
adb-shell: @cortinico any reason why default parameter should be removed from the `FileFactory` interface? the reason why I added was API can be used as it was earlier `fileFactory.create()` , in addition to that people can pass filename if required(filename being optional) instead of mandatorily passing an empty string as `fileFactory.create("")`

library/src/main/java/com/chuckerteam/chucker/internal/support/DialogUtils.kt
```diff
@@ -0,0 +1,25 @@
+package com.chuckerteam.chucker.internal.support
+
+import android.content.Context
+import com.chuckerteam.chucker.internal.data.model.DialogData
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+
+internal object DialogUtils {
+    fun showDialog(
```
vbuberen, line 8: We might change this to be an extension function for Context.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
```diff
@@ -107,7 +131,51 @@ internal class TransactionListFragment :
         TransactionActivity.start(requireActivity(), transactionId)
     }
 
+    private fun exportTransactions() {
+        viewModel.getAllTransactions { transactions ->
```
vbuberen, line 135: This whole function is hard to read, especially when we get to part after `else`. 
Let's change it to plain simple functions, like `viewModel.getAllTransactions()` just returns a `List` and `createExportFile` just returns file.  
We need to try keep library code as simple as possible, so everyone is able to understand it fast.

library/src/main/java/com/chuckerteam/chucker/internal/data/model/DialogData.kt
```diff
@@ -0,0 +1,8 @@
+package com.chuckerteam.chucker.internal.data.model
+
+internal data class DialogData(
```
vbuberen, line 3: I don't see much value in adding this class.

library/src/main/java/com/chuckerteam/chucker/internal/support/FileFactory.kt
cortinico: I generally feel that having a method `fun create(filename: String = ""): File` in an interface is a bad API design practice. What you're saying here is that you offer a _create_ method that will return a File given a filename that you could not provide, if you don't provide the filename is going to be `""`.

What is happening is that in your implementation you're actually ignoring the parameter if is `""` and you're there defining the real _default value_. So I suggest you either:

* Convert `FileFactory` to an `abstract class`. Move up the `AtomicLong` and have a single method like: `abstract fun create(filename: String = "chucker-${uniqueIdGenerator.getAndIncrement()}"): File`. There you actually be backward compatible and the intent would be clearer.

* Offer two `create` method in the interface: a `create()` (for backward compatibility and a `create(filename: String)` with no default parameter.

library/src/main/java/com/chuckerteam/chucker/internal/data/model/DialogData.kt
adb-shell: if I remove this data class [showDialog](https://github.com/ChuckerTeam/chucker/pull/305/files/5e1275ab62a02055be52d8a50156637cc27dedf8#diff-98f2123ddac8875c70a7ec16e42d5a1dR8) will accepting 6 parameters.
1. title
2. message
3. positive button text
4. negative button text
5. onPositiveClick
6. onNegativeClick

which causes detekt to fail the build with ` LongParameterList` message. however, I can remove 4 and 6 since its usage in this project is less, but in order to keep the method more generic, I have added those params.

I was not able to think of any other solution to avoid these parameters, But I am open to any suggestions you guys might have here.

library/src/main/java/com/chuckerteam/chucker/internal/support/DialogUtils.kt
```diff
@@ -0,0 +1,25 @@
...
+        context: Context,
+        dialogData: DialogData,
+        positiveButtonHandler: (() -> Unit)?,
```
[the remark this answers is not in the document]
adb-shell: updated.

library/src/main/java/com/chuckerteam/chucker/internal/support/DialogUtils.kt
```diff
@@ -0,0 +1,25 @@
...
+        negativeButtonHandler: (() -> Unit)?
```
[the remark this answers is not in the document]
adb-shell: updated.

library/src/main/java/com/chuckerteam/chucker/internal/support/DialogUtils.kt
adb-shell: updated as an extension  `ContextExt`

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionListFragment.kt
adb-shell: agreed, my bad. I have refactored to make it simpler please do take another look at this function.

library/src/main/res/values/strings.xml
adb-shell: updated

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
```diff
@@ -35,6 +40,20 @@ internal class MainViewModel : ViewModel() {
...
+    }
+
+    fun createExportFile(content: String, fileFactory: FileFactory): File = runBlocking {
+        val file = async {
+            val file = fileFactory.create(fileFactory.exportFileName)
+            file.writeText(content)
+            return@async file
+        }
+        file.await()
+    }
```
cortinico, line 55:
```suggestion
    fun createExportFile(content: String, fileFactory: FileFactory): File = runBlocking {
        val file = viewModelScope.async(Dispatchers.IO) {
            val file = fileFactory.create(EXPORT_FILENAME)
            file.writeText(content)
            return@async file
        }
        file.await()
    }
```

Same as above. Moreover you can maybe specify `Dispatchers.IO` here as you're doing IO work.

library/src/main/java/com/chuckerteam/chucker/internal/support/FileFactory.kt
```diff
@@ -3,5 +3,7 @@ package com.chuckerteam.chucker.internal.support
 import java.io.File
 
 internal interface FileFactory {
+    val exportFileName: String get() = "transactions.txt"
```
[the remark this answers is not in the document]
vbuberen: And it should be turned into `private`, I guess.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
```diff
@@ -35,6 +40,20 @@ internal class MainViewModel : ViewModel() {
     val throwables: LiveData<List<RecordedThrowableTuple>> = RepositoryProvider.throwable()
         .getSortedThrowablesTuples()
 
+    fun getAllTransactions(): List<HttpTransaction>? = runBlocking {
+        val transactions = async { RepositoryProvider.transaction().getAllTransactions() }
+        transactions.await()
+    }
```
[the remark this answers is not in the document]
vbuberen: It was there before latest changes from the author.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
adb-shell: updated.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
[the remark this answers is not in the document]
adb-shell: sorry was missed, hence updated.

library/src/main/java/com/chuckerteam/chucker/internal/support/FileFactory.kt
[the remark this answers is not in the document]
adb-shell: it is accessed in `MainViewModel` hence cannot be private but have limited access internal.

library/src/main/java/com/chuckerteam/chucker/internal/data/model/DialogData.kt
vbuberen: Ok, let's leave it.

vbuberen:
There is an error in logcat when you are trying to export files:
```
2020-04-17 14:04:46.590 17878-17901/com.chuckerteam.chucker.sample E/DatabaseUtils: Writing exception to parcel
    java.lang.SecurityException: Permission Denial: reading androidx.core.content.FileProvider uri content://com.chuckerteam.chucker.provider/export/transactions.txt from pid=18005, uid=1000 requires the provider be exported, or grantUriPermission()
        at android.content.ContentProvider.enforceReadPermissionInner(ContentProvider.java:729)
        at android.content.ContentProvider$Transport.enforceReadPermission(ContentProvider.java:602)
        at android.content.ContentProvider$Transport.query(ContentProvider.java:231)
        at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:104)
        at android.os.Binder.execTransactInternal(Binder.java:1021)
        at android.os.Binder.execTransact(Binder.java:994)
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
```diff
@@ -35,6 +42,22 @@ internal class MainViewModel : ViewModel() {
     val throwables: LiveData<List<RecordedThrowableTuple>> = RepositoryProvider.throwable()
         .getSortedThrowablesTuples()
 
+    fun getAllTransactions(): List<HttpTransaction>? = runBlocking {
```
vbuberen, line 45: With `runBlocking` you just killing the whole point of using coroutines. No matter what you switched to other dispatcher below - you already blocked the main thread till this code finishes.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
```diff
@@ -35,6 +42,22 @@ internal class MainViewModel : ViewModel() {
...
+        val transactions = viewModelScope.async(Dispatchers.Default) {
+            RepositoryProvider.transaction().getAllTransactions()
+        }
+        transactions.await()
+    }
+
+    fun createExportFile(content: String, fileFactory: FileFactory): File = runBlocking {
```
vbuberen, line 52: Same as above about `runBlocking`

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
adb-shell: I am kinda new to the coroutines concept, thanks for pointing it out.

so I tried a couple of approaches here
1. passing a function as a callback argument, but as you mentioned it seems to make the function hard to read.
2. keep Livedata list of httptransaction and update it after your fetch, but it seemed overkill to observe for exporting transactions.

Hence the final approach for which I settled here was adding `lifecycle-runtime-ktx` so that I would be able to launch coroutine in lifecycle scope, and it is canceled when the Lifecycle is destroyed. 

Let me know if you have any alternate solution or think adding `lifecycle-runtime-ktx` dependency is unnecessary.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
adb-shell: updated, with the same approach as above.

adb-shell:
> There is an error in logcat when you are trying to export files:
> 
> ```
> 2020-04-17 14:04:46.590 17878-17901/com.chuckerteam.chucker.sample E/DatabaseUtils: Writing exception to parcel
>     java.lang.SecurityException: Permission Denial: reading androidx.core.content.FileProvider uri content://com.chuckerteam.chucker.provider/export/transactions.txt from pid=18005, uid=1000 requires the provider be exported, or grantUriPermission()
>         at android.content.ContentProvider.enforceReadPermissionInner(ContentProvider.java:729)
>         at android.content.ContentProvider$Transport.enforceReadPermission(ContentProvider.java:602)
>         at android.content.ContentProvider$Transport.query(ContentProvider.java:231)
>         at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:104)
>         at android.os.Binder.execTransactInternal(Binder.java:1021)
>         at android.os.Binder.execTransact(Binder.java:994)
> ```

sure will look into the error, however, I was able to export file in the android 10 device.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
vbuberen: I am not a fan of adding more dependencies. Take a look at `TransactionPayloadFragment` to see how another contributor did pretty similar stuff.

vbuberen:
> sure will look into the error, however, I was able to export file in the android 10 device.
Yes, it seems to work, but it is better to resolve the error, since we don't know where it might not work due it.

library/build.gradle
```diff
@@ -73,6 +73,7 @@ dependencies {
     implementation "androidx.palette:palette-ktx:$paletteKtxVersion"
 
     implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:$lifecycleVersion"
+    implementation "androidx.lifecycle:lifecycle-runtime-ktx:$lifecycleVersion"
```
vbuberen, line 76: No need to add this one. I already mentioned a solution in our conversation under `runBlocking`

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
adb-shell: cool thanks for the pointer, updated.

library/build.gradle
adb-shell: updated.

adb-shell:
Resolved the log cat error, error was because of [FLAG_GRANT_READ_URI_PERMISSION](https://developer.android.com/reference/android/content/Intent.html#FLAG_GRANT_READ_URI_PERMISSION). The permission grants read operations on the URI in the `Intent's data` and any URIs specified in its `ClipData`. had missed it earlier hence added the URI in clip-data.

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
```diff
@@ -35,6 +41,16 @@ internal class MainViewModel : ViewModel() {
     val throwables: LiveData<List<RecordedThrowableTuple>> = RepositoryProvider.throwable()
         .getSortedThrowablesTuples()
 
+    suspend fun getAllTransactions(): List<HttpTransaction>? = withContext(Dispatchers.Default) {
```
vbuberen, line 44: There is no need in specifying `Dispatcher`, since Room handles switching from UI thread for DAOs.
You can find more info here: https://medium.com/androiddevelopers/room-coroutines-422b786dc4c5

library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
```diff
@@ -0,0 +1,15 @@
+package com.chuckerteam.chucker.internal.support
+
+import android.content.Context
+import com.chuckerteam.chucker.R
+import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
+
+internal object ShareUtils {
+    fun getStringFromTransactions(transactions: List<HttpTransaction>, context: Context): String {
```
vbuberen, line 8: I think that this fun might have to deal with pretty big lists of transactions. Let's turn it into suspend fun as well to something like this:
```suggestion
     suspend fun getStringFromTransactions(transactions: List<HttpTransaction>, context: Context): String =
        withContext(Dispatchers.Default) {
            return@withContext transactions.joinToString(
                separator = "\n${context.getString(R.string.chucker_export_separator)}\n",
                prefix = "${context.getString(R.string.chucker_export_prefix)}\n",
                postfix = "\n${context.getString(R.string.chucker_export_postfix)}\n"
            ) { FormatUtils.getShareText(context, it, false) }
        }
```

library/src/test/java/com/chuckerteam/chucker/internal/support/ShareUtilsTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    @Test
+    fun isStringFromTransactionsListValid() {
+        val transactionList = ArrayList<HttpTransaction>()
+
+        repeat(10) {
+            transactionList.add(TestTransactionFactory.createTransaction(getRandomHttpMethod()))
+        }
+
+        val actualStringFromTransactions = ShareUtils.getStringFromTransactions(transactionList, contextMock)
```
vbuberen, line 45: In case you apply my suggestion to make `getStringFromTransactions()` a suspend function you will have to update this line as well. And this is where your previous approach with `runBlocking` should be applied:
```suggestion
        val actualStringFromTransactions = runBlocking {
            ShareUtils.getStringFromTransactions(transactionList, contextMock)
        }
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/MainViewModel.kt
adb-shell: thanks for the pointer, updated.

library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
adb-shell: updated.

library/src/test/java/com/chuckerteam/chucker/internal/support/ShareUtilsTest.kt
adb-shell: updated.

cortinico:
Looks great to me as well 👍 Thanks for the contribution @adb-shell

PR #308, opened 2020-04-05
Sample UI updates

vbuberen:
## :camera: Screenshots
| Day | Night |
| --- | --- |
| Screenshot 2020-04-05 at 14 07 57 | Screenshot 2020-04-05 at 14 08 10 |

## :page_facing_up: Context
While the library itself had some theme updates the sample app was quite abandoned. So I created a minimalistic theme (without that many colors as the library has) to match both sample and library.

## :pencil: Changes
- Switched from `AppCompat` to `MaterialComponents` theme.
- Added `night` colors to match the theme of Chucker library when dark mode enabled.
- Made buttons slightly bigger and added margins between them to improve UX.
- Removed some redundant resources like `mdpi` images and ic_launcher_web` image.

## :stopwatch: Next steps
- Create another PR with updated AlertDialogs.
- After getting feedback about current PR and merging it I will create updated GIF assets for the readme file.


sample/src/main/res/values/colors.xml
```diff
@@ -1,15 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources xmlns:tools="http://schemas.android.com/tools">
-    <!--
-    Palette from
-    https://material.io/tools/color/#!/?view.left=0&view.right=0&primary.color=01579B&secondary.color=00E676
```
cortinico, line 5: Would be cool if we still have some reference to the Chucker palette somewhere. It doesn't necessary has to be in the `colors.xml`.

vbuberen: Ok, I will write it down into the backlog)

PR #320, opened 2020-04-07
Image preview background contrast

### LINKED ISSUE #319: Detect contrast colour for downloaded images

Downloading images that are fully tinted with a colour in a range of the background colour used by Chucker makes them virtually invisible for the user. Here is an example of such an icon which is white. It is visible in night mode but not in day mode.

Use `Palette` to determine what should be a contrasting background for the image. It could look like something along the lines of this.

https://github.com/MiSikora/Squash-It/blob/93812d1a13d8fbb62c2290659d0785f250f94507/squash-it/library/src/main/java/io/mehow/squashit/screenshot/ScreenshotActivity.kt#L55-L60

The only difference would be to use `Palette` output to get colour instead of day or night mode.

Screenshot_20200407-172130

Screenshot_20200407-172209

- [x] Yes
- [ ] No

MiSikora:
## :camera: Screenshots

Screenshot_1586783170
Screenshot_1586783151
Screenshot_1586783159
Screenshot_1586783154
Screenshot_1586783146
Screenshot_1586783173

## :page_facing_up: Context

There is an issue. #319

## :pencil: Changes

Showing downloaded image preview triggers `Palette` to determine contrast colour based on the luminescence of the image. You can see in the screenshots that light image has dark background and vice versa.

One unfortunate thing is that `Palette` does not recognise alpha channel so we cannot filter out the the transparent background and `Palette` treats it as a black colour. What I do is I create a new bitmap that uses magenta in place of transparent pixels and then filter them out in the `Palette`. I chose magenta as it is rather a rare colour.

## :no_entry_sign: Breaking

No.

I added image requests to the sample app. Unfortunately I didn't find any image generator that would operate on images with alpha and starting a local server just for that purpose seems too much.

## :stopwatch: Next steps

Closes #319.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/ChessboardDrawable.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/ChessboardDrawable.kt
@@ -1,0 +1,75 @@
+package com.chuckerteam.chucker.internal.support
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.Bitmap.Config.ARGB_8888
+import android.graphics.BitmapShader
+import android.graphics.Canvas
+import android.graphics.ColorFilter
+import android.graphics.Paint
+import android.graphics.Paint.Style.FILL
+import android.graphics.PixelFormat
+import android.graphics.Rect
+import android.graphics.Shader.TileMode.REPEAT
+import android.graphics.drawable.Drawable
+import androidx.annotation.ColorInt
+import androidx.annotation.ColorRes
+import androidx.annotation.DimenRes
+import androidx.annotation.Px
+import androidx.core.content.ContextCompat
+
+internal class ChessboardDrawable(
+    @ColorInt evenColor: Int,
+    @ColorInt oddColor: Int,
+    @Px squareSize: Int
+) : Drawable() {
+    private val chessboardPaint = Paint().apply {
+        val patternBitmap = Bitmap.createBitmap(squareSize * 2, squareSize * 2, ARGB_8888)
+        patternBitmap.eraseColor(evenColor)
+
+        color = oddColor
+        style = FILL
+        val patternCanvas = Canvas(patternBitmap)
+        val squareRect = Rect(squareSize, 0, 2 * squareSize, squareSize)
+        patternCanvas.drawRect(squareRect, this)
+        squareRect.offsetTo(0, squareSize)
+        patternCanvas.drawRect(squareRect, this)
+
+        reset()
+        shader = BitmapShader(patternBitmap, REPEAT, REPEAT)
+    }
+
+    override fun draw(canvas: Canvas) {
+        canvas.drawPaint(chessboardPaint)
+    }
+
+    override fun setAlpha(alpha: Int) {
+        chessboardPaint.alpha = alpha
+    }
+
+    override fun getOpacity(): Int {
+        return if (chessboardPaint.colorFilter == null) {
+            PixelFormat.OPAQUE
+        } else {
+            PixelFormat.TRANSLUCENT
+        }
+    }
+
+    override fun setColorFilter(colorFilter: ColorFilter?) {
+        chessboardPaint.colorFilter = colorFilter
+    }
+
+    companion object {
+        fun createPattern(
+            context: Context,
+            @ColorRes evenColorId: Int,
+            @ColorRes oddColorId: Int,
+            @DimenRes sizeId: Int
+        ): ChessboardDrawable {
+            val evenColor = ContextCompat.getColor(context, evenColorId)
+            val oddColor = ContextCompat.getColor(context, oddColorId)
+            val size = context.resources.getDimensionPixelSize(sizeId)
+            return ChessboardDrawable(evenColor, oddColor, size)
+        }
+    }
+}
--- a/build.gradle
+++ b/build.gradle
@@ -11,6 +11,7 @@ buildscript {
         roomVersion = '2.2.5'
         lifecycleVersion = '2.2.0'
         androidXCoreVersion = '2.1.0'
+        paletteKtxVersion = '1.0.0'
 
         // Publishing
         androidMavenGradleVersion = '2.1'
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -70,6 +70,7 @@ dependencies {
 
     implementation "com.google.android.material:material:$materialComponentsVersion"
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutVersion"
+    implementation "androidx.palette:palette-ktx:$paletteKtxVersion"
 
     implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:$lifecycleVersion"
     implementation "androidx.lifecycle:lifecycle-livedata-ktx:$lifecycleVersion"
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/BitmapUtils.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/BitmapUtils.kt
@@ -1,0 +1,47 @@
+package com.chuckerteam.chucker.internal.support
+
+import android.graphics.Bitmap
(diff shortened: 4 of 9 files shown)


vbuberen:
I haven't worked with Palette before, but is it that demanding task that it requires something to offload it from the main thread?

MiSikora:
> 1. I don't like that we have that color defining logic inside adapter. While added code relates to displaying I would try to avoid doing such things there and leave only things which adapter usually does there (I am also struggling with such issue now with my payload search improvements feature)
>    Since we already doing our own items creation for this adapter in PayloadFragment, let's do this thing there as well and just pass the background color into `TransactionPayloadItem`. We already have a coroutine there and you shouldn't worry about creating new async things.

Sure, I'll change it.

> 2. Users might not be aware whenever this contrast border is a part of their downloaded image or Chucker's background. This is solely a UX problem, but quite serious from my point of view.  I will think about ideas here, but we need to clearly indicate our background and user content to not confuse anyone.

I was thinking about changing the whole theme for a fragment but it would make a little bit inconsistent with other ones. If it is not an issue for you I'd go with that.

> I haven't worked with Palette before, but is it that demanding task that it requires something to offload it from the main thread?

It iterates over all pixels of an image and analyses them in a histogram-like fashion. It depends on the size of the image and the max area of the image that you allow for. By default it limits it to `112 x 112`. Personally, I wouldn't bother with doing it asynchronously but didn't know what's your approach so I took safer one.

vbuberen:
As to background - you can leave it for now. I might come up with some ideas here, so to not make you do something twice let's skip for now.

Thanks for explanations about Palette stuff.

MiSikora:
> Thanks for explanations about Palette stuff.

Just do be sure. You'd like to change it to synchronous call? I can always make it asynchronous and write a suspending extension since it will be moved out from the `ViewHolder` where there is access to some UI scope.

vbuberen:
No-no, under `background` I meant fragment background change that your mentioned when told about theme 😄

vbuberen:
That code with palette can would perfectly fit the coroutine in `processPayload`, so let it be there.

cortinico:
> Users might not be aware whenever this contrast border is a part of their downloaded image or Chucker's background. This is solely a UX problem, but quite serious from my point of view. I will think about ideas here, but we need to clearly indicate our background and user content to not confuse anyone.

I'm also a bit concerned here. How about having an icon on the bottom right (or somewhere else) that will toggle the background of the `ImageView` from black to white and vice versa?

I really like the approach of using `Palette` but I'm afraid it could create some confusion in the user that don't really understand what is going on (plus is yet another dependency that we're adding on our users load).

MiSikora:
> I'm also a bit concerned here. How about having an icon on the bottom right (or somewhere else) that will toggle the background of the `ImageView` from black to white and vice versa?

The only issue is that it requires one step from the user which might not be at time confusing is when they do not see the image. Maybe a switch (with i.e. moon and sun icons) that triggers background of the image view and that has the initial state set by `Palette`?

> I really like the approach of using `Palette` but I'm afraid it could create some confusion in the user that don't really understand what is going on (plus is yet another dependency that we're adding on our users load).

 I wouldn't be too concerned about the size of `Palette`. It has like 160 methods and 23 KB. Unless you also mean additional dependency that devs have to be aware of. But then, I guess, automatic recognition is out of the scope and we should go only with the switch.

cortinico:
> Unless you also mean additional dependency that devs have to be aware of.

I'm just afraid that pulling in dependencies could create problems in debug builds for our consumers. Say a customer is on a older version of `Palette`, and applying `Chucker` that it's on a newer version of `Palette` will cause the resulting version to be ours. This could have unintended side effects as well as complicating debugging.

Anyway I feel that I'm maybe worrying too much for this specific case as the library is pretty small.

> I wouldn't be too concerned about the size of `Palette`. It has like 160 methods and 23 KB. 

Ok it feels like is something we can live with. Let's go for that.

> Maybe a switch (with i.e. moon and sun icons) that triggers background of the image view and that has the initial state set by Palette?

This would be my preferred approach if possible 👍

vbuberen:
@cortinico 
Regarding concerns about versions difference between Chucker and user's app - I am not sure a lot of people use `Palette` in their dependencies. Also, it seems to update not that often, so we should be fine here.
As to additional button - I am not sure where we should place it at the moment.

What comes to my mind is a variation of idea suggested initially by @MiSikora on background change, but only for image view item - to make such item match parent view width and height and make it is background grey (for example) so it will be clear that it is our background. 
Here is a quick prototype of such thing:
Screenshot 2020-04-13 at 14 22 34
It is not ideal, but might be better for user perception. We might need to increase height as well.

What do you think, guys?

vbuberen:
Or we might take it further and in that `Response` tab add some image data, like its resolution, etc. and after a few sort of headers add something like `Image preview:` item and place an image below. Might be a good enhancement. 
It is not a request for this PR or author of this PR 😄Just came to my mind.

MiSikora:
> As to additional button - I am not sure where we should place it at the moment.

Yes, I'm also having trouble picturing it. I tried to code it but then I kind of hit a brick wall when it came to the design. Mostly because images can take up the whole width of the screen. I thought of two solutions.

1. There is a button below the image that changes the background of the whole item.
2. There is a switch to the end of an item. But then item should be probably a constraint layout that limits the size of an image to i.e. 60% of width so it still can be centred and fit the switch at the same time.

> What comes to my mind is a variation of idea suggested initially by @MiSikora on background change, but only for image view item - to make such item match parent view width and height and make it is background grey (for example) so it will be clear that it is our background.
> 
> What do you think, guys?

Do you mean to have it always grey regardless of the luminescence? I think it might be ok but maybe it would be better just to have two values of grey? One that is good for light images and the other that is good for dark images. Otherwise it will fall into the same trap as the original issue if an image has a transparent background and is tinted with a colour similar to the background one.

> Or we might take it further and in that `Response` tab add some image data, like its resolution, etc. and after a few sort of headers add something like `Image preview:` item and place an image below. Might be a good enhancement.

It would be good for now but I think this should belong to an item in a view holder if multipart support is something to be implemented in the future.

vbuberen:
> Mostly because images can take up the whole width of the screen.

I would like to address it in some future PRs by adding sort of scaling for images.

> Do you mean to have it always grey regardless of the luminescence? I think it might be ok but maybe it would be better just to have two values of grey?

No, we would have to use 2 shades of grey anyway to match both day and night theme or image luminescence. 
And another idea just came to me after you mentioned images with alpha channel - what if we use patterned background for images, like Android Studio does (attaching a screenshot)? This background would be hard to confuse with image content, I guess.

Screenshot 2020-04-13 at 14 57 49

MiSikora:
> And another idea just came to me after you mentioned images with alpha channel - what if we use patterned background for images, like Android Studio does (attaching a screenshot)? This background would be hard to mix with image content, I guess.

Yes, I like. I will make another PR with this approach to see how it looks like.

MiSikora:
I added the chessboard background. I think the result is good. The colours for chess fields are based on the primary colour so they are not only in shades of grey. Here are the screenshots.

Screenshot_1586783170
Screenshot_1586783151
Screenshot_1586783159
Screenshot_1586783154
Screenshot_1586783146
Screenshot_1586783173

library/src/main/java/com/chuckerteam/chucker/internal/support/ChessboardDrawable.kt
```diff
@@ -0,0 +1,75 @@
+package com.chuckerteam.chucker.internal.support
```
MiSikora, line 1: I wasn't sure about the package for this class. I added it to the `support` but maybe `ui` is more appropriate.

vbuberen: From my point of view it is fine to place in `support`.

cortinico:
> The result with that chessboard pattern is awesome 👍

Really great job 💪 Thanks for the addition @MiSikora

PR #323, opened 2020-04-11
Format URL encoded forms

gm-vm:
## :camera: Screenshots
|Current|Formatted|
|-|-|
|current|formatted|

## :page_facing_up: Context
Reading encoded forms is nearly impossible, especially when dealing values have symbols in them. Unlike for XML and JSON requests, this does not simply pretty print the body and that may be undesired.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
@@ -192,10 +192,11 @@ internal class HttpTransaction(
 
     private fun formatBody(body: String, contentType: String?): String {
         return when {
-            contentType != null && contentType.contains("json", ignoreCase = true) ->
-                FormatUtils.formatJson(body)
-            contentType != null && contentType.contains("xml", ignoreCase = true) ->
-                FormatUtils.formatXml(body)
+            contentType.isNullOrBlank() -> body
+            contentType.contains("json", ignoreCase = true) -> FormatUtils.formatJson(body)
+            contentType.contains("xml", ignoreCase = true) -> FormatUtils.formatXml(body)
+            contentType.contains("x-www-form-urlencoded", ignoreCase = true) ->
+                FormatUtils.formatUrlEncodedForm(body)
             else -> body
         }
     }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
@@ -10,6 +10,8 @@ import java.io.ByteArrayInputStream
 import java.io.IOException
 import java.io.PrintWriter
 import java.io.StringWriter
+import java.io.UnsupportedEncodingException
+import java.net.URLDecoder
 import java.nio.charset.Charset
 import java.util.Locale
 import javax.xml.XMLConstants
@@ -91,6 +93,24 @@ internal object FormatUtils {
             xml
         } catch (t: TransformerException) {
             xml
+        }
+    }
+
+    fun formatUrlEncodedForm(form: String): String {
+        return try {
+            if (form.isBlank()) {
+                return form
+            }
+            form.split("&").joinToString(separator = "\n") { entry ->
+                val keyValue = entry.split("=")
+                val key = keyValue[0]
+                val value = if (keyValue.size > 1) URLDecoder.decode(keyValue[1], "UTF-8") else ""
+                "$key: $value"
+            }
+        } catch (e: IllegalArgumentException) {
+            form
+        } catch (e: UnsupportedEncodingException) {
+            form
         }
     }
 
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
@@ -26,6 +26,7 @@ import com.chuckerteam.chucker.R
 import com.chuckerteam.chucker.databinding.ChuckerFragmentTransactionPayloadBinding
 import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
 import com.chuckerteam.chucker.internal.support.calculateLuminance
+import com.chuckerteam.chucker.internal.support.combineLatest
 import java.io.FileNotFoundException
 import java.io.FileOutputStream
 import java.io.IOException
@@ -79,27 +80,29 @@ internal class TransactionPayloadFragment :
             adapter = payloadAdapter
         }
 
-        viewModel.transaction.observe(
-            viewLifecycleOwner,
-            Observer { transaction ->
-                if (transaction == null) return@Observer
-                uiScope.launch {
-                    payloadBinding.loadingProgress.visibility = View.VISIBLE
+        viewModel.transaction
+            .combineLatest(viewModel.formatRequestBody)
+            .observe(
+                viewLifecycleOwner,
+                Observer { (transaction, formatRequestBody) ->
+                    if (transaction == null) return@Observer
+                    uiScope.launch {
+                        payloadBinding.loadingProgress.visibility = View.VISIBLE
 
-                    val result = processPayload(type, transaction)
-                    if (result.isEmpty()) {
-                        showEmptyState()
-                    } else {
-                        payloadAdapter.setItems(result)
-                        showPayloadState()
+                        val result = processPayload(type, transaction, formatRequestBody)
+                        if (result.isEmpty()) {
+                            showEmptyState()
+                        } else {
+                            payloadAdapter.setItems(result)
+                            showPayloadState()
+                        }
+                        // Invalidating menu, because we need to hide menu items for empty payloads
+                        requireActivity().invalidateOptionsMenu()
+
+                        payloadBinding.loadingProgress.visibility = View.GONE
                     }
-                    // Invalidating menu, because we need to hide menu items for empty payloads
-                    requireActivity().invalidateOptionsMenu()
-
-                    payloadBinding.loadingProgress.visibility = View.GONE
                 }
-            }
-        )
+            )
     }
 
     private fun showEmptyState() {
@@ -148,7 +151,14 @@ internal class TransactionPayloadFragment :
             }
         }
 
-        menu.findItem(R.id.encode_url).isVisible = false
+        if (type == TYPE_REQUEST) {
+            viewModel.doesRequestBodyRequireEncoding.observe(
+                viewLifecycleOwner,
+                Observer { menu.findItem(R.id.encode_url).isVisible = it }
+            )
+        } else {
+            menu.findItem(R.id.encode_url).isVisible = false
+        }
 
         super.onCreateOptionsMenu(menu, inflater)
     }
@@ -228,7 +238,8 @@ internal class TransactionPayloadFragment :
 
     private suspend fun processPayload(
         type: Int,
-        transaction: HttpTransaction
+        transaction: HttpTransaction,
+        formatRequestBody: Boolean
     ): MutableList<TransactionPayloadItem> {
         return withContext(Dispatchers.Default) {
             val result = mutableListOf<TransactionPayloadItem>()
@@ -240,7 +251,11 @@ internal class TransactionPayloadFragment :
             if (type == TYPE_REQUEST) {
                 headersString = transaction.getRequestHeadersString(true)
                 isBodyPlainText = transaction.isRequestBodyPlainText
-                bodyString = transaction.getFormattedRequestBody()
+                bodyString = if (formatRequestBody) {
+                    transaction.getFormattedRequestBody()
+                } else {
+                    transaction.requestBody ?: ""
+                }
             } else {
                 headersString = transaction.getResponseHeadersString(true)
                 isBodyPlainText = transaction.isResponseBodyPlainText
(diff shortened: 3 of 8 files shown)


gm-vm:
I thought about connecting the existing button, but I had some doubts and, honestly, I don't know why I didn't mention them.

What I've just come up with is the following: I'd add a new button that allows to switch between two display modes: "raw" and "formatted". This button could be either the existing button or a new one, not sure what's better. I simply don't want to add request-specific parameters to `HttpTransaction` and this is the only way to do that. I'd also include responses, just for consistency.

But how should the share button behave? Would it follow the current display mode? Should it include both the formatted request and the raw request? Just the one currently displayed?

library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
```diff
@@ -94,6 +96,19 @@ internal object FormatUtils {
         }
     }
 
+    fun formatUrlEncodedForm(form: String): String {
```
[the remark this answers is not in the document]
gm-vm: Sure, I'll also rewrite the method not to make any assumption on the input.

cortinico:
> This button could be either the existing button or a new one, not sure what's better. I simply don't want to add request-specific parameters to `HttpTransaction` and this is the only way to do that. I'd also include responses, just for consistency.

Agree. The stored information should be unchanged. What we're addressing here is just on the presentation layer.

If possible I would hook to the same button we already have on the URL side. The semantic is pretty clear as it's an encoded/decoded toggle.

gm-vm:
> Agree. The stored information should be unchanged. What we're addressing here is just on the presentation layer.

My bad, I didn't notice you were also exposing the original request body, which makes sense, and thought I had to do some invasive changes.

Anyway, I hooked the existing button. I should probably implement something like `doesUrlRequireEncoding` and maybe rename some variables, but first let me know if something like this is OK.

vbuberen:
Joining the conversation here - agree with @cortinico that we should attach this feature to an existing button. However, we might need to reevaluate UI in next PRs (I suggest to do it in first of 4.x releases).

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -71,21 +71,50 @@ internal class TransactionPayloadFragment :
...
         viewModel.transaction.observe(
             viewLifecycleOwner,
             Observer { transaction ->
                 if (transaction == null) return@Observer
                 uiScope.launch {
                     showProgress()
-                    val result = processPayload(type, transaction)
-                    payloadBinding.responseRecyclerView.adapter = TransactionBodyAdapter(result)
-                    payloadBinding.responseRecyclerView.setHasFixedSize(true)
+                    val encode = viewModel.encodeUrl.value ?: true
```
vbuberen, line 99: Why don't you call this variable the same way as you named parameter `updateAdapter` fun?

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -71,21 +71,50 @@ internal class TransactionPayloadFragment :
...
+
+        val result = mutableListOf<TransactionPayloadItem>()
+        val adapter = TransactionBodyAdapter(result)
+        payloadBinding.responseRecyclerView.adapter = adapter
+        payloadBinding.responseRecyclerView.setHasFixedSize(true)
+
+        viewModel.encodeUrl.observe(
+            viewLifecycleOwner,
+            Observer { encode ->
+                if (adapter.itemCount == 0) {
```
vbuberen, line 83: Why do you need this check if you also do the check for transaction value below?

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -113,7 +142,13 @@ internal class TransactionPayloadFragment :
             }
         }
 
-        menu.findItem(R.id.encode_url).isVisible = false
+        menu.findItem(R.id.encode_url).apply {
+            isVisible = true
+            setOnMenuItemClickListener {
```
vbuberen, line 147: This options item already had click listener in `TransactionActivity` and called exactly the same fun from ViewModel.

library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
```diff
@@ -94,6 +96,24 @@ internal object FormatUtils {
...
+        return try {
+            if (form.isBlank()) {
+                return form
+            }
+            form.split("&").joinToString(separator = "\n") { entry ->
+                val split = entry.split("=")
```
vbuberen, line 105: Rename this `val` please for more clarity.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -71,21 +71,50 @@ internal class TransactionPayloadFragment :
...
         )
     }
 
+    private suspend fun updateUpdater(
+        data: MutableList<TransactionPayloadItem>,
+        adapter: TransactionBodyAdapter,
+        transaction: HttpTransaction,
+        formatted: Boolean
+    ) {
+        data.clear()
```
vbuberen, line 113: In general this function shouldn't be in this fragment. It should be inside the adapter itself named something like `setItems()`.

vbuberen: Also, in case of having such thing as `setItems()` function you can remove adapter initialisation with an empty list and make its declaration lazy for cases when we have empty payload.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -71,21 +71,50 @@ internal class TransactionPayloadFragment :
...
-                    payloadBinding.responseRecyclerView.setHasFixedSize(true)
+                    val encode = viewModel.encodeUrl.value ?: true
+                    updateUpdater(result, adapter, transaction, !encode)
                     hideProgress()
                 }
             }
         )
     }
 
+    private suspend fun updateUpdater(
```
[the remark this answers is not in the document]
gm-vm: Definitely, I had a typo there and instead of fixing it my brain decided to make it worse.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
gm-vm: As I mentioned, I need to rename some stuff. I started with just the formatting of url encoded forms and ended up with more. I was wondering if I should stick with just the formatting of url encoded forms or continue with this more generic solution.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
gm-vm: Definitely something I should have commented. There are cases in which `transaction` is not null here (e.g., configuration change), so you end up notifying the adapter twice: once here and once when observing `transaction`. Not a big deal, but it's unnecessary work.

I tried to think of a way not to have this here and the only thing I could think of is having a separate `transaction` LiveData that under the hood is just a MediatorLiveData that combines `transaction` and `encodeUrl`, taking care of notifying observers only once the first time it's observed. It didn't look pretty and it was more convoluted than this simple check, which is just a micro-optimization and not something required. With or without this check everything would work just fine. But I don't really like the fact that this trick only works if you start observing this LiveData before the other.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
gm-vm: Definitely, I didn't like this to begin with and I guess I just didn't want to change too many things. I'll fix this. But I'm not sure I get exactly what you meant with your proposed solution. Are you suggesting the use of a lazy property or something else?

Here I can only see a nullable list as marginally better, given that the moment you attach the adapter to the RecyclerView its `getItemCount()` gets called, so anything that creates the list lazily would do it immediately.

vbuberen: Yes, I suggest to have a lazy property and create it only when we have some result to show.
However, now I think it might be not the best idea, since it will cause logcat show `No adapter attached`, like it does now.

gm-vm:
I rebased my changes and applied your suggestions.

I also changed the way things work. The button now only encodes/decodes url encoded requests.

I don't really like that `formatRequestBody` boolean there. I gave it a generic name, but we know that's a lie. I'd rather have the fragment not do anything and receive already formatted bodies. And maybe not as simple Strings, but something like a `Spannable`, which I see is what we end up with anyway. This would allow to format nicely certain type of requests.

For example, `application/x-www-form-urlencoded` requests are generally represented as I'm doing here, but keys and values are generally shown either with different colors or font weights. Then there are also `multipart/form-data`, where things get a bit more complicated.

vbuberen:
> I don't really like that `formatRequestBody` boolean there. I gave it a generic name, but we know that's a lie. I'd rather have the fragment not do anything and receive already formatted bodies.

Agree with you. It is something I also suggested recently. Look at `Next steps` here https://github.com/ChuckerTeam/chucker/pull/325#issue-403678329 . TL:DR - it left since times we had no view models. But it is going to change soon.

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -191,12 +191,16 @@ internal class HttpTransaction(
...
+        return if (contentType.isNullOrBlank()) {
+            body
+        } else {
+            when {
+                contentType.contains("json", ignoreCase = true) -> FormatUtils.formatJson(body)
+                contentType.contains("xml", ignoreCase = true) -> FormatUtils.formatXml(body)
+                contentType.contains("x-www-form-urlencoded", ignoreCase = true) ->
+                    FormatUtils.formatUrlEncodedForm(body)
+                else -> body
+            }
```
cortinico, line 203:
```suggestion
        return when {
            contentType.isNullOrBlank() -> body
            contentType.contains("json", ignoreCase = true) -> FormatUtils.formatJson(body)
            contentType.contains("xml", ignoreCase = true) -> FormatUtils.formatXml(body)
            contentType.contains("x-www-form-urlencoded", ignoreCase = true) ->
                FormatUtils.formatUrlEncodedForm(body)
            else -> body
        }
```

Please keep the single `return when` style as it's easier to read and follow.

library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
```diff
@@ -94,6 +96,24 @@ internal object FormatUtils {
...
+    fun formatUrlEncodedForm(form: String): String {
+        return try {
+            if (form.isBlank()) {
+                return form
+            }
+            form.split("&").joinToString(separator = "\n") { entry ->
+                val keyValue = entry.split("=")
+                val key = keyValue[0]
+                val value = if (keyValue.size > 1) URLDecoder.decode(keyValue[1], "UTF-8") else ""
+                "$key: $value"
```
cortinico, line 108:
```suggestion
entry.split("=").let { 
    (key, value) -> "$key: ${URLDecoder.decode(value, "UTF-8")}" 
}
```

You don't need extra variables as you can use a destructuring declaration.
Moreover you have to extend the catch below to include the `IndexOutOfBoundsException`

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
```diff
@@ -33,8 +33,23 @@ internal class TransactionViewModel(
...
         }
 
+    val doesRequestBodyRequireEncoding: LiveData<Boolean> = RepositoryProvider.transaction()
+        .getTransaction(transactionId)
+        .map { transaction ->
+            if (transaction == null) {
+                false
+            } else {
+                transaction.requestContentType?.contains("x-www-form-urlencoded", ignoreCase = true) ?: false
+            }
```
cortinico, line 43:
```suggestion
            transaction?.requestContentType?.contains("x-www-form-urlencoded", ignoreCase = true) ?: false
```

library/src/main/java/com/chuckerteam/chucker/internal/support/FormatUtils.kt
gm-vm: My initial change had a destructuring declaration, but dropped it in favor of this to support what you can see in `testFormatUrlEncodedForm_noValues` (i.e., requests that need to send empty strings).

The current code should never throw an `IndexOutOfBoundsException`, since `split` always returns at least one string.

cortinico: Makes sense, let's drop my comment. Thanks for clarifying.

vbuberen:
@cortinico So, we are good to merge now? I see that PR is updated already.

cortinico:
Yes 👍

PR #325, opened 2020-04-15
Empty state for payloads

### LINKED ISSUE #276: Add state for empty screens
Currently Chucker shows no difference between a request with empty body and a request which somehow failed to fetch info to display. States should be distinct, so user could understand what is going on.

In cases where the request/response is really empty we should show some state telling user that there is an empty request/response. 

Here is how empty state looks now
Screenshot 2020-03-17 at 22 00 16

- [x] Yes
- [ ] No

vbuberen:
## :camera: Screenshots
| Request | Response |
| --- | --- |
|Screenshot 2020-04-15 at 13 21 35 | Screenshot 2020-04-15 at 13 21 58|

## :page_facing_up: Context
This PR closes #276 by checking request or response payload and showing corresponding state in case any of them is empty.

## :pencil: Changes
- Added empty state view into layout for requests and responses.
- Slightly changed the way `TransactionPayloadAdater` is declared and gets its items.
- Moved adapter attachment to `RecyclerView` to `onViewCreated`, so we could get rid of logcat messages `No adapter attached`.

## :paperclip: Related PR
This PR would require some updates in #305 and #323, since there were changes in same classes.

## :stopwatch: Next steps
While working on this PR I came to conclusion that we should move `processPayload` method to `TransactionViewModel`, because it is left from times when we had no `ViewModel`s at all and because transformation of items into view items looks like a job for `ViewModel`.

However, I would postpone it till #305, #323 and my planned PR with search improvement gets merged to avoid too much updates requirements for those PRs.


library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -72,21 +73,54 @@ internal class TransactionPayloadFragment :
...
+                    if (result.isEmpty()) {
+                        showEmptyState()
+                    } else {
+                        payloadAdapter.setItems(result)
+                        showPayloadState()
+                    }
+                    // Invalidating menu, because we need to hide menu items for empty payloads
+                    requireActivity().invalidateOptionsMenu()
+
+                    payloadBinding.loadingProgress.visibility = View.GONE
```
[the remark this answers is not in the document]
vbuberen: It is a ProgressBar. We show it when trying to fetch the payload and hide when it is fetched.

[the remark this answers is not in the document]
MiSikora: But is it suspended or something? Wouldn't it result always in `GONE` since the code is executed in one function?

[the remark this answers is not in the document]
vbuberen: No, it works fine with showing progress on screen opening and hiding it properly.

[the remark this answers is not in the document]
MiSikora: Ahh… I check out the code in the IDE. Process payload suspends so it allows for changing progress visibility. Too bad it is not that obvious via GH. :(

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
```diff
@@ -20,12 +20,18 @@ import com.chuckerteam.chucker.internal.support.highlightWithDefinedColors
...
  * performances when loading big payloads.
  */
-internal class TransactionBodyAdapter(
-    private val bodyItems: List<TransactionPayloadItem>
-) : RecyclerView.Adapter<TransactionPayloadViewHolder>() {
+internal class TransactionBodyAdapter : RecyclerView.Adapter<TransactionPayloadViewHolder>() {
+
+    private val items = arrayListOf<TransactionPayloadItem>()
+
+    fun setItems(bodyItems: List<TransactionPayloadItem>) {
```
cortinico, line 27:
```suggestion
    fun setItems(items: List<TransactionPayloadItem>) {
```

cortinico: That's actually a nit, but still will keep the code consistent.

library/src/main/res/layout/chucker_fragment_transaction_payload.xml
```diff
@@ -1,31 +1,72 @@
...
+        app:layout_constraintTop_toTopOf="parent" />
+
+    <ImageView
+        android:id="@+id/emptyPayloadImage"
+        android:layout_width="@dimen/chucker_octa_grid"
+        android:layout_height="@dimen/chucker_octa_grid"
+        android:tint="@color/chucker_color_primary"
+        android:layout_marginBottom="@dimen/chucker_quad_grid"
+        android:contentDescription="@string/chucker_body_empty"
+        android:src="@drawable/chucker_empty_payload"
```
cortinico, line 30: Nit: can you please reformat this XML file. The order of the attributes is wrong. It should be:

```suggestion
    <ImageView
        android:id="@+id/emptyPayloadImage"
        android:layout_width="@dimen/chucker_octa_grid"
        android:layout_height="@dimen/chucker_octa_grid"
        android:layout_marginBottom="@dimen/chucker_quad_grid"
        android:contentDescription="@string/chucker_body_empty"
        android:src="@drawable/chucker_empty_payload"
        android:tint="@color/chucker_color_primary"
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -249,8 +267,10 @@ internal class TransactionPayloadFragment :
                     result.add(TransactionPayloadItem.BodyLineItem(SpannableStringBuilder.valueOf(it)))
                 }
             } else {
-                bodyString.lines().forEach {
-                    result.add(TransactionPayloadItem.BodyLineItem(SpannableStringBuilder.valueOf(it)))
+                if (!bodyString.isBlank()) {
```
cortinico, line 270:
```suggestion
                if (bodyString.isNotBlank()) {
```
Also why is this needed? `.lines()` should be empty if the bodyString is blank.

vbuberen: It is not empty. It has 1 item with just an empty string. I added this check so I could decide if the payload is empty.

library/src/main/res/layout/chucker_fragment_transaction_payload.xml
vbuberen: Ok, will do. Just skipped that because it also changes spacing.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
vbuberen: I made it like that to avoid calling `this.items`. Just for readability.

vbuberen:
What about the idea of moving `processPayload` into a ViewModel?

cortinico:
> What about the idea of moving `processPayload` into a ViewModel?

Yeah definitely 👍 Are you going to take care of this?

vbuberen:
Yes

PR #326, opened 2020-04-15
Kotlin to 1.3.72

cortinico:
## :page_facing_up: Context
Just bumping Kotlin to the latest patch version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
     ext {
-        kotlinVersion = '1.3.71'
+        kotlinVersion = '1.3.72'
         androidGradleVersion = '3.6.1'
         coroutineVersion = '1.3.5'
 


cortinico:
> Potentially, we could start testing 1.4, since by the time we release `3.3.0` JetBrains should already have it released.

I actually wanted to setup a Github Actions matrix to run our test suite both on `1.3.x` and `1.4.x` just to see how it behaves.

As for bumping directly to `1.4` I'll wait for M2 or M3 as it's still the first release and I'm afraid this might add uncertainty to the SNAPSHOT builds.

vbuberen:
> I actually wanted to setup a Github Actions matrix to run our test suite both on `1.3.x` and `1.4.x` just to see how it behaves.
 
That would be awesome!

PR #327, opened 2020-04-16
Split the CI steps between build and check

cortinico:
## :page_facing_up: Context
Currently we run our build in only on Gradle run and it makes hard to understand where the failure actually happened. Let's split it into separate steps.

Let's check if we have a major impact on the build time.
I expect some improvement due to the caching (not on the first run though). On the other side, running several separate gradle commands will lead to running configuration phase multiple times, as well as some duplicate work.

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -14,5 +14,36 @@ jobs:
     steps:
     - name: Checkout Repo
       uses: actions/checkout@v2
-    - name: Run Gradle tasks
-      run: ./gradlew clean ktlintCheck lint detekt build test
+
+
+    - name: Cache Gradle Caches
+      uses: actions/cache@v1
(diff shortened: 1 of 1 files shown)


cortinico:
Great so the last build https://github.com/ChuckerTeam/chucker/runs/593782902 took 2m55sec. While on the avg. a successful build takes 4m+.

vbuberen:
It is not exactly the same that I suggested when we discussed it recently, but it is my fault that I wasn't clear enough.
I thought about moving ktlint, detekt stuff to a separate workflow, so we will have clearly 3 check for every PR - Gradle wrapper validation, tests, code checks.

However, that might cause longer build times and you approach is actually good.

vbuberen:
It seem that you achieved faster times because steps can run in parallel. It is much getter at the moment.

PR #330, opened 2020-04-24
Setup a Kotlin version matrix on GH Actions

cortinico:
## :page_facing_up: Context

I'm setting up a matrix on GH Actions for Kotlin version. This will allow us to have two parallel builds. One for Kotlin 1.3.x and another for the preview of 1.4

Relevant context here https://github.com/ChuckerTeam/chucker/pull/326#issuecomment-614462958

## :pencil: Changes
The Kotlin version is not set from an environment variable and has a fallback at the 1.3.x version so local builds are not impacted.

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -10,11 +10,22 @@ on:
 
 jobs:
   gradle:
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-latest]
+    strategy:
+      fail-fast: false
+      matrix:
+        kotlin-version:
+          - 1.3.72
+          - 1.4-M1
+
     steps:
     - name: Checkout Repo
       uses: actions/checkout@v2
 
+    - name: Set environment variables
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-merge.yaml
```diff
@@ -10,11 +10,22 @@ on:
 
 jobs:
   gradle:
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-latest]
+    strategy:
+      fail-fast: false
```
vbuberen, line 15: This line meant to not fail build if we have issues on 1.4? 
Since we shouldn't have issues on 1.3.72 obviously.

If so, maybe change it to `true` or remove? To enforce preparations for 1.4.

cortinico: > This line meant to not fail build if we have issues on 1.4?

No it means that if one of the two builds fail, the other will be aborted.
I generally prefer to see the result of both builds to easily understand if the issue is on the Kotlin preview or affecting both builds.

PR #335, opened 2020-05-02
Update payload type definition

vbuberen:
## :page_facing_up: Context
This PR starts a process of sort of big rework how payload data is fetched to remove any business logic related things from fragments. To avoid bloated PR which is hard to review I will do a series of small when it is possible.
This time I changed approach to definition of required payload type for `TransactionPayloadFragment` from constants to a sealed class, which should help have less checks in place as well as be used by other classes in future.

## :pencil: Changes
- Switched from constants to sealed class in `TransactionPayloadFragment`.

## :stopwatch: Next steps
To bring more context - I would like to move all payload processing from fragment and have everything in `ViewModel`, so fragment will just render data provided to it without any decisions.
With current state of things in the library it will be quite hard, but can be managed in a step by step manner.
I am still not sure if we should try to fit all the logic into the shared ViewModel as we currently have or split them in separate ones. Having separate ones would require additional sync for common things like `encodeURL`, etc., but should ease the process of testing. At the same time having everything in one ViewModel would benefit from not having to sync data, but will be hard to test, not to mention issues with mixed responsibility. 
Due to mentioned facts I would like to also hear some feedback/thoughts about preferred ways to ease future development.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/PayloadType.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/PayloadType.kt
@@ -1,0 +1,8 @@
+package com.chuckerteam.chucker.internal.ui.transaction
+
+import java.io.Serializable
+
+sealed class PayloadType : Serializable {
+    object Request : PayloadType()
+    object Response : PayloadType()
+}
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/PayloadType.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/PayloadType.kt
@@ -1,8 +1,6 @@
 package com.chuckerteam.chucker.internal.ui.transaction
 
-import java.io.Serializable
-
-sealed class PayloadType : Serializable {
-    object Request : PayloadType()
-    object Response : PayloadType()
+internal enum class PayloadType {
+    REQUEST,
+    RESPONSE
 }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
@@ -47,7 +47,7 @@ internal class TransactionPayloadFragment :
     private var backgroundSpanColor: Int = Color.YELLOW
     private var foregroundSpanColor: Int = Color.RED
 
-    private val payloadType: PayloadType by lazy {
+    private val payloadType: PayloadType by lazy(LazyThreadSafetyMode.NONE) {
         arguments?.getSerializable(ARG_TYPE) as PayloadType
     }
 
@@ -108,7 +108,7 @@ internal class TransactionPayloadFragment :
 
     private fun showEmptyState() {
         payloadBinding.apply {
-            emptyPayloadTextView.text = if (payloadType is PayloadType.Response) {
+            emptyPayloadTextView.text = if (payloadType == PayloadType.RESPONSE) {
                 getString(R.string.chucker_response_is_empty)
             } else {
                 getString(R.string.chucker_request_is_empty)
@@ -152,7 +152,7 @@ internal class TransactionPayloadFragment :
             }
         }
 
-        if (payloadType is PayloadType.Request) {
+        if (payloadType == PayloadType.REQUEST) {
             viewModel.doesRequestBodyRequireEncoding.observe(
                 viewLifecycleOwner,
                 Observer { menu.findItem(R.id.encode_url).isVisible = it }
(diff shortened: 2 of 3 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/PayloadType.kt
```diff
@@ -0,0 +1,8 @@
+package com.chuckerteam.chucker.internal.ui.transaction
+
+import java.io.Serializable
+
+sealed class PayloadType : Serializable {
```
cortinico, line 5: Two things here:

- Visibility should be `internal`
- This use case seems more appropriate for an `enum`, unless you had some future development in mind?

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -47,15 +47,16 @@ internal class TransactionPayloadFragment :
     private var backgroundSpanColor: Int = Color.YELLOW
     private var foregroundSpanColor: Int = Color.RED
 
-    private var type: Int = 0
+    private val payloadType: PayloadType by lazy {
```
cortinico, line 50:
```suggestion
    private val payloadType: PayloadType by lazy(LazyThreadSafetyMode.NONE) {
```
Given that the `payloadType` is accessed only by the main thread.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -293,26 +293,23 @@ internal class TransactionPayloadFragment :
...
-                                if (transaction.responseImageData != null) {
-                                    fos.write(transaction.responseImageData)
-                                } else {
-                                    throw IOException(TRANSACTION_EXCEPTION)
-                                }
+                                    ?: if (transaction.responseImageData != null) {
+                                        fos.write(transaction.responseImageData)
+                                    } else {
+                                        throw IOException(TRANSACTION_EXCEPTION)
+                                    }
```
cortinico, line 312: This should be rewritten in just:

```
transaction.responseBody?.byteInputStream()?.copyTo(fos)
    ?: throw IOException(TRANSACTION_EXCEPTION)	                                   
```

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -293,26 +293,23 @@ internal class TransactionPayloadFragment :
     }
 
     @Suppress("ThrowsCount")
```
cortinico, line 295: With your change this `@Suppress("ThrowsCount")` can also go

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/PayloadType.kt
vbuberen: Yeah, agree. I was in that mindset with other changes I have locally at the moment where some sealed classes have some data. 
Will update.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
vbuberen: Interesting. Didn't know about this thing. Thanks.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
vbuberen: Note that with this change we are going to get errors every time user tries to save response with image. Now it saves it, despite not opening it due no file extension.

I wanted to address it in a separate issue to have a clear file type for saved transactions, like `txt` for text transactions, so users could open this saved file easily right on their devices.

vbuberen: Changing as you are suggesting, since it is useless anyway to save an image without being able to open it with ease later.

vbuberen:
@cortinico Please take a look at the `Next steps` part of PR.

cortinico:
> @cortinico Please take a look at the `Next steps` part of PR.

Sorry my bad I wanted to answer it and I just left a review:

> To bring more context - I would like to move all payload processing from fragment and have everything in ViewModel, so fragment will just render data provided to it without any decisions.
With current state of things in the library it will be quite hard, but can be managed in a step by step manner.

++ for this.

> I am still not sure if we should try to fit all the logic into the shared ViewModel as we currently have or split them in separate ones

On this side, I'd try to first cleanup all the logic from the Fragments and move it inside a single `ViewModel`. Splitting it could be done in a separate round. Having a clear separation of concern between Fragments and VM sounds like already as a great achievement to me. How we organise the VM is something that we can still change and discuss later.

PR #336, opened 2020-05-03
Move share related functions into a proper file

vbuberen:
## :page_facing_up: Context
Saw that our share related functions are in `FormatUtils` file, while we already have `ShareUtils`.
To avoid confusion moved such functions into `ShareUtils` and updated corresponding tests.


library/src/test/java/com/chuckerteam/chucker/internal/support/ShareUtilsTest.kt
```diff
@@ -35,6 +37,10 @@ class ShareUtilsTest {
         every { getString(R.string.chucker_body_empty) } returns "(body is empty)"
     }
 
+    private val requestMethods = arrayOf("GET", "POST", "PUT", "DELETE")
+
+    private fun getRandomHttpMethod(): String = requestMethods[Random.nextInt(0, 4)]
```
cortinico, line 42: I know you're just moving the code so feel free to drop, but:

There is a `.random()` extension function on `Array` that we can use here:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/random.html

vbuberen: I will update. Thanks.

Issue #324, opened 2020-04-14
JSON Request/Response Body Syntax Color Highlighting


seanghay:
## :warning: Is your feature request related to a problem? Please describe

![image](https://user-images.githubusercontent.com/15277233/79184062-5c9b2e00-7e3d-11ea-9e19-a3bed9c525a1.png)

## :raising_hand: Do you want to develop this feature yourself?

- [x] Yes
- [ ] No

seanghay:
Would it be nice if we can display JSON like this?
![image](https://user-images.githubusercontent.com/15277233/79192878-bb6ba200-7e53-11ea-9e72-f844a9d4be93.png)

cortinico:
Sounds like a valuable feature @seanghay. Just a couple of things to keep in mind:

- I'd love to make sure that the color scheme is configurable has good contrast ratio both in the light and the dark background.
- In this page there is also a "search" feature that will highlight the found occurrences of a string you search. Please make sure that it still works correctly.
- Performance is really important in this page. We already worked on the body page and refactored to use a `RecyclerView` with one entry per line. Please make sure that adding a syntax highlighting won't impact the performance of this page significantly.

seanghay:
@cortinico Is is slow to use set everything into a single TextView rather than using RecyclerView to render each line?

vbuberen:
Regarding search - there will be my PR this week changing it a little bit. Just be aware.

@seanghay As to putting everything into one TextView - yes, it is slow. On some big responses it caused ANR errors. This is why we switched to RecyclerView.

seanghay:
Thanks @vbuberen. I will start looking into it and hopefully will open a pull request soon.

vbuberen:
Would like to to give a heads up on this issue - I started to play around with [MergeAdapter](https://medium.com/androiddevelopers/merge-adapters-sequentially-with-mergeadapter-294d2942127a), which looks a great thing for lots of improvements, including this feature request. While it is still in alpha we can't let it go into the library, but I will open a draft PR to have it visible and maybe inspire someone.

mertceyhan:
Hi @seanghay, how is it going your working? It's a really good feature request, if you want I can help you for delivering the feature.

seanghay:
> Hi @seanghay, how is it going your working? It's a really good feature request, if you want I can help you for delivering the feature.

Hi @mertceyhan! I haven't worked on it yet, if you can help that would be great.

mertceyhan:
[...]

PR #337, opened 2020-05-04
Payload processing refactoring

vbuberen:
## :page_facing_up: Context
[ConcatAdapter](https://medium.com/androiddevelopers/merge-adapters-sequentially-with-mergeadapter-294d2942127a) seems like a very suitable component for Chucker and its payload RecyclerView. Having 3 separate adapters for each part of payload (headers, text body or image body) can help us to incorporate lots of improvements, like #324 or #195 much easier.

For example, we could get rid of having all headers formatted as a single String with HTML markup as it is done now (this code is already in PR) and use just raw names and values, which gives us more flexibility like being able to set spans however we need it.

I came to this idea after doing some work on search improvements for #195. Moreover, as we already discussed in some PRs as a side topic we need to move payload processing for output into a ViewModel.
My intention here is to have 3 lazily initialised adapters which we will add into parent only if there is a corresponding content (like add `ImageAdapter` only if there is an image in a payload).
Also, I am going to switch to something similar to MVI approach to have states which view observes and renders (loading, data, empty).

This PR is in early preview, so no need for review. However, feedback is welcome.
I made it visible to everyone, since it might help some contributors to see where this part of the library is heading and to not produce too much conflicts.

## :pencil: Changes
- Changed `PayloadAdapter` to be a new `MergeAdapter`
- Added 3 separate adapters for headers, text body, image body.
to be continued...

## :paperclip: Related PR
Will add later


library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionHeadersAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.RecyclerView
+import com.chuckerteam.chucker.databinding.ChuckerTransactionItemHeadersBinding
+import com.chuckerteam.chucker.internal.data.entity.HttpHeader
+import com.chuckerteam.chucker.internal.support.applyBoldSpan
+
+internal class TransactionHeadersAdapter : RecyclerView.Adapter<TransactionHeadersAdapter.HeaderViewHolder>() {
+
+    private val headers = arrayListOf<HttpHeader>()
```
olivierperez, line 13: Generaly speaking, I don't get why we should use `arrayListOf` instead of `mutableListOf`.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -42,7 +43,19 @@ internal class TransactionPayloadFragment :
     Fragment(), SearchView.OnQueryTextListener {
 
     private lateinit var payloadBinding: ChuckerFragmentTransactionPayloadBinding
-    private val payloadAdapter = TransactionBodyAdapter()
+
+    private val payloadAdapter = TransactionPayloadAdapter()
+
+    private val headerAdapter by lazy { TransactionHeadersAdapter() }
+    private val bodyAdapter by lazy { TransactionBodyAdapter() }
+    private val imageAdapter by lazy { TransactionImageAdapter() }
```
olivierperez, line 51: These 3 adapter are not yet called, is it normal?

vbuberen: It is a draft PR, it is not finalized as well as not meant for review.

library/src/main/java/com/chuckerteam/chucker/internal/support/SpanUtils.kt
```diff
@@ -60,3 +63,7 @@ private fun applyColoredSpannable(
             builder
         }
 }
+
+fun CharSequence.applyBoldSpan() = SpannableString(this).apply {
+    setSpan(StyleSpan(Typeface.BOLD), 0, this.length, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+}
```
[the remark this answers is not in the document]
vbuberen: Sounds good.

vbuberen:
Looks like RecyclerView update isn't a priority for Google now, because 1.2.0 is still in alpha.
I doubt that we get a stable version any time soon, so this PR will have to live quite a long time gathering some conflicts etc.

Closing till we have a stable `RecyclerView` version with `ConcatAdapter`.

Issue #321, opened 2020-04-08
Poll about `Errors` tab future


vbuberen:
Hey everyone.

Chucker evolves pretty fast and becomes better every day thanks to great feedback from the community and contributors hard work. However, there are some features that don't get much feedback on possible improvements. One such feature is ability to capture `Throwable` errors and show them in `Errors` tab. Moreover there are even some suggestions asking to disable that tab or remove it completely (like in #88 or [here](https://www.reddit.com/r/androiddev/comments/fvbmw8/chucker_v320_is_out_with_a_lot_of_uiperf/fmifj3n/)). 

While it looks reasonable that Chucker could remove this feature completely and focus just on network part we don't know for sure if this feature is used a lot or everyone just ignores it. Since there is no analytics available we would like to create sort of a poll in this issue and see your opinion.

To vote just put a reaction to this message with emoji. 

Available options are:

👍 - Leave `Errors` tab and errors collection feature.
👎 - Remove `Errors` tab and errors collection feature.

This poll is valid for 1 month till `May 8th` so we could gather enough feedback.

We would also love to hear your feedback on whenever you or your team used/use `Errors` tab and if it was/is helpful, so don't hesitate to leave comments here, in this issue.

cortinico:
We also got this question on Reddit, I'm dropping the link here so we have it as a reference: https://www.reddit.com/r/androiddev/comments/fvbmw8/chucker_v320_is_out_with_a_lot_of_uiperf/fmifj3n

ColtonIdle:
It just seems to me like chucker should stick to doing what it does well and what it was created to do. Networking.

I've got other libraries for crashes.

cortinico:
Quick reminder as this is closing in 2 weeks.

tpakis:
Just to add to the discussion, I recently wrote a small article about the 2 use cases I found for Chucker, one is of course to log network traffic but the other is to log Uncaught exceptions with a custom crash handler. It's really useful mainly for testers, having both features in only one tool is something that helps a lot, so should you choose to focus on network logs please leave a way (maybe in settings) to enable it for people that find this useful: https://itnext.io/log-uncaught-exceptions-and-network-traffic-with-chucker-for-android-ps-967fe3e0488c?source=friends_link&sk=71105759bd2316d8f5cc420c4dc7d5a7

vbuberen:
Hi @tpakis.
[...]

PR #338, opened 2020-05-10
Deprecate Throwable related parts

vbuberen:
## :page_facing_up: Context
Since[ there is a plan](https://github.com/ChuckerTeam/chucker/issues/321#issuecomment-626138370) to remove Throwable catching functionality in `4.x` we need to prepare users to not introduce breaking changes right away. This PR adds proper deprecation messages, which should both inform and help to switch to valid functions.

## :pencil: Changes
- Added deprecation messages to public API.

Nothing special for testing. IDE should show suggestions if there is a replacement for deprecated function.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
@@ -20,6 +20,8 @@ internal class NotificationHelper(val context: Context) {
 
     companion object {
         private const val TRANSACTIONS_CHANNEL_ID = "chucker_transactions"
+
+        @Deprecated("This variable will be removed in 4.x release")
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
```diff
@@ -163,4 +163,9 @@ internal class NotificationHelper(val context: Context) {
     fun dismissErrorsNotification() {
         notificationManager.cancel(ERROR_NOTIFICATION_ID)
     }
+
+    fun dismissNotifications() {
```
vbuberen, line 167: This fun is a subject to change after removing Throwable stuff.

vbuberen:
Good feedback. Thanks. I will address then in this PR.

PR #348, opened 2020-05-22
Update rebase workflow to use new checkout

vbuberen:
## :page_facing_up: Context
It seems that Rebase workflow used old version of Checkout Action

## :pencil: Changes
Switched to `checkout@v2`

## :stopwatch: Next steps
If this one works fine I will also update the same Action in Revert workflow.


vbuberen:
Let's check how it works
/rebase

PR #350, opened 2020-05-22
Parallelise the pre-merge workflow

cortinico:
## :page_facing_up: Context
Let's split the `pre-merge` workflow into smaller jobs that can run in parallel.

This should significantly increase the speed of the CI and will provide immediate feedback on the PR on what failed.

If the CI is ✅, we're good to go.


vbuberen:
It is a nice boost!

Issue #56, opened 2019-06-26
Is there any plan to host the artifact on https://mvnrepository.com/.


TonyTangAndroid:
If the artifact is hosted on https://jitpack.io/, we could have security concerns. I know that it could be painful to get it work for the first time. We could leverage the plugin here https://github.com/vanniktech/gradle-maven-publish-plugin to minimize our effort.

cortinico:
Hey @TonyTangAndroid 

> we could have security concerns

Which security concerns are you talking about? If you refer to XBI attacks and something similar to what was described here https://blog.autsoft.hu/a-confusing-dependency/ ...those kind of attacks are unrelated to JitPack and more related to your build.gradle configuration.

On the other hand, I was wondering if you're raising the issue as you're having problem accessing JitPack or not?

koral--:
If chucker is published on maven central then another projects published there can depend on it.

cortinico:
> If chucker is published on maven central then another projects published there can depend on it.

Is this a real blocker?
According to this page: https://maven.apache.org/repository/guide-central-repository-upload.html
Is not a real blocker. 

Ideally we could host Chucker on Maven Central as soon as we release 3.0.0

koral--:
Thanks for pointing this out!
Last time I've checked there was a requirement that each dependency must be also available on central.

vbuberen:
@cortinico 
Going to refresh the state of this issue. Are we going to redeploy to Maven?
Or we will do something like multiple sources to have some sort of backup?

cortinico:
> Going to refresh the state of this issue. Are we going to redeploy to Maven?

Sure, I'm happy to explain to everyone.
Initially, @olivierperez hosted the first versions of Chucker on Bintray/JCenter. We then moved to JitPack as this simplifies the whole deployment process. Specifically:

* We don't have to deal with a GPG key, signing, key-sharing, etc.
* We get `-SNAPSHOTS` versions for every branch out of the box
* The release process is extremely streamlined as we need to only push a git tag to get a new release.

> Or we will do something like multiple sources to have some sort of backup?

I think we can definitely publish to Maven Central if the community is requesting it. This will have the benefit that other libraries on Central can easily depend on us. Moreover will overcome some of the potential security issues mentioned by @TonyTangAndroid. We just need to agree internally on the whole GPG key mgmt @vbuberen

Sloy:
Hi! Have you discussed this topic any further?

[...]

PR #352, opened 2020-05-22
Update readme mentioning deployment to Maven Central

cortinico:
## :page_facing_up: Context
I'm updating the README to reflect #56 
I've also added a note in the Snapshot section mentioning that artifacts are still available both on Sonatype or JitPack.

The link of Maven Central:
https://search.maven.org/artifact/com.github.chuckerteam.chucker/library
might take some hours before showing meaningful results.


cortinico:
> So we are dropping JitPack for releases?

Not really, they will still be there. I think we should anyway suggest a "preferred" source and Maven Central is definitely more reliable, artifacts are signed, and not extra setup is needed on the `repositories{}` block.

Should we update the wording in the REAMDE?

vbuberen:
No, for now we can leave it as it is. We will have some changes pretty soon anyway.

PR #384, opened 2020-07-04
Allow Chucker to display partially read application responses

### LINKED ISSUE #370: Response body is not shown when stream is not fully received
The TeeSource class that is used for wrapping around a response body Source and capture the response, requires the input stream to become exhausted (e.g. read() to return -1), in order to capture the response. However, it is possible that an entity reads the exact number of bytes to fully decode a resource, and then close the stream. In this case, the response is not captured (only the response headers are displayed).

This was working up until 3.1.2, and in version 3.2.0.

## :bomb: Steps to reproduce

I have created a sample app which demonstrates the problem. You can find it [here](https://github.com/gmetal/chucker-response-app). If you use the [retrofit-logansquare](https://github.com/mannodermaus/retrofit-logansquare) converter, no responses are shown. However, if you use either the kotlinx-serialization or moshi retrofit adapters, responses are properly shown.

## :wrench: Expected behavior

The responses bodies are included in the response.

## :camera: Screenshots
N/A

## :iphone: Tech info
 - Device: 
 - OS: 
 - Chucker version: 3.2.0

MiSikora:
## :page_facing_up: Context

There is an open issue - #370.

## :pencil: Changes

I mostly worked on `TeeSource` and implemented `Callback`. Callers are now informed of finished byte processing whenever an upstream source is closed and now they are responsible for handling any data corruption that might have occurred.

Exceeded byte count limit is no longer reported as a failure. Instead, bytes are no longer written to a file. Also, I added some guarding against failures that might happen when writing to a file (due to low-end devices and so on).

This change also enables displaying not fully read responses if Chucker is applied as an application-level `Interceptor`. Network-level is not always doable due to compression (this might be further improved in the future but right now I don't want to mess with it).

## :no_entry_sign: Breaking

No.

Some units test are written. You can also run the sample from #370 created by @gmetal with Chucker from my branch `com.github.MiSikora.chucker:library:response-read-SNAPSHOT` available via JitPack.

## :stopwatch: Next steps

Closes #370.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -235,30 +235,27 @@ class ChuckerInterceptor internal constructor(
         private val response: Response,
         private val transaction: HttpTransaction
     ) : TeeSource.Callback {
-        override fun onSuccess(file: File) {
-            val buffer = readResponseBuffer(file, response.isGzipped)
-            file.delete()
+        override fun onClosed(file: File) {
+            val buffer = try {
+                readResponseBuffer(file, response.isGzipped)
+            } catch (e: IOException) {
+                null
+            }
             if (buffer != null) processResponseBody(response, buffer, transaction)
             collector.onResponseReceived(transaction)
-        }
-
-        override fun onFailure(exception: IOException, file: File) {
             file.delete()
-            collector.onResponseReceived(transaction)
         }
 
-        private fun readResponseBuffer(responseBody: File, isGzipped: Boolean): Buffer? {
+        override fun onFailure(exception: IOException, file: File) = Unit
+
+        private fun readResponseBuffer(responseBody: File, isGzipped: Boolean): Buffer {
             val bufferedSource = Okio.buffer(Okio.source(responseBody))
             val source = if (isGzipped) {
                 GzipSource(bufferedSource)
             } else {
                 bufferedSource
             }
-            return try {
-                Buffer().apply { source.use { writeAll(it) } }
-            } catch (_: IOException) {
-                null
-            }
+            return Buffer().apply { source.use { writeAll(it) } }
         }
     }
 
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
@@ -46,7 +46,9 @@ internal class TeeSource(
             return bytesRead
         }
 
-        copyBytesToFile(sink, bytesRead)
+        if (!isFailure) {
+            copyBytesToFile(sink, bytesRead)
+        }
 
         return bytesRead
     }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
@@ -14,8 +14,6 @@ import java.io.IOException
  * like a regular [Source]. While bytes are read from the [upstream] the are also copied
  * to a [sideChannel] file. After the [upstream] is depleted or when a failure occurs
  * an appropriate [callback] method is called.
- *
- * Failure is considered any [IOException] during reading the bytes.
  */
 internal class TeeSource(
     private val upstream: Source,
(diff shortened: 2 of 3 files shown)


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -235,30 +235,27 @@ class ChuckerInterceptor internal constructor(
...
             collector.onResponseReceived(transaction)
-        }
-
-        override fun onFailure(exception: IOException, file: File) {
             file.delete()
-            collector.onResponseReceived(transaction)
         }
 
-        private fun readResponseBuffer(responseBody: File, isGzipped: Boolean): Buffer? {
+        override fun onFailure(exception: IOException, file: File) = Unit
```
cortinico, line 249: Should you delete the file here?

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -26,9 +23,8 @@ internal class TeeSource(
 ) : Source {
     private val sideStream = Okio.buffer(Okio.sink(sideChannel))
     private var totalBytesRead = 0L
-    private var isReadLimitExceeded = false
-    private var isUpstreamExhausted = false
     private var isFailure = false
+    private var isClosed = false
```
cortinico, line 27: Do we really need those two? Is there a risk `.onClosed` or `.onFailure` gets called twice?

cortinico:
> Network-level is not always doable due to compression (this might be further improved in the future but right now I don't want to mess with it).

Agree

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
MiSikora: `isClosed` flag should be here. Otherwise closing the source multiple times would trigger callback that many times as well and closing should generally be an idempotent operation.

`isFalure` can be removed. Personally I don't think it makes sense to report all failures and multiple can occur while writing to a file. If a single write fails (due to some file corruption on a low-end device or whatever) there is a high chance it will fail again. This way we can skip writing If a failure occurred and report only a first failure (either network IO or disk IO).

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
MiSikora: Not anymore. Now, `onClosed` is called whenever upstream is closed. Even when there was a failure in the meantime. Deleting file here would make it unavailable in the `onClosed` callback.

MiSikora: Unless you want to `onClosed` be called only when failure did not occur. I changed the name from `onSuccess` to `onClosed` because now it is hard to know what success really is since we no longer rely on the upstream being exhausted.

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -78,23 +80,22 @@ internal class TeeSource(
...
 
         /**
-         * Called when there was an issue while copying bytes to the [file].
-         *
-         * It might occur due to one of the following reasons:
-         * - an exception was thrown while reading bytes
-         * - capacity limit was exceeded
-         * - upstream was not fully consumed
+         * Called when an exception was thrown while reading bytes from the upstream
+         * or when writing to a side channel file occurs. Any read byte are available in a [file].
```
vbuberen, line 98: I am not sure if there is a typo in `or when writing to a side channel file occurs`. For me it seems like `called when writing occurs` and it might be not the case meant. Or I didn't understand something here?

MiSikora: Yes, the wording is misleading. I meant when an exception is thrown while writing to a side-channel file. I will update it.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
cortinico: Yeah I think now with `onClosed` is clearer 👍

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
cortinico: Great thanks for clarifying

vbuberen:
I would like to do some testing later today. I already feel that there is no need to, but anyway. If it works as expected I will merge.

vbuberen:
Wasn't able to find time yesterday. 
Tested and works fine. 
Thanks again for dealing with such tricky issue.

Issue #128, opened 2019-10-03
Add Cookie support


cortinico:
**Is your feature request related to a problem? Please describe.**
Chucker should provide a way to show the Cookies that are sent for every request.

**Describe the solution you'd like**
Ideally having a separated tab for cookies would be helpful. As a MVP having them listed below the headers would be a good starting point.

**Do you want to develop this feature yourself?**
nop :) It's open for development

psh:
Before I drop a PR for cookies support, I'd like to get a little feedback on the UI.

The "cookies" tab is displayed if there are cookies present in either the request or the response
![01-request-with-cookies-overview](https://user-images.githubusercontent.com/407647/85891154-43efb280-b7b4-11ea-93bf-ae8543cf706a.png)

The "cookies" tab reuses the same UI style of headers that we have on request / response
![02-request-with-cookies-cookies-tab](https://user-images.githubusercontent.com/407647/85891405-a3e65900-b7b4-11ea-9661-6dc80d966452.png)

Here's the question - the cookies are present as headers in both the request and response.  I opted to leave the headers (see screenshot) even though they appear on the cookies tab, as they *are* valid headers.  Would you prefer them to be hidden?
![03-request-with-cookies-request](https://user-images.githubusercontent.com/407647/85891481-cbd5bc80-b7b4-11ea-872a-34cae5313599.png)

Cookies going in both directions
![04-cookies-sent-and-received](https://user-images.githubusercontent.com/407647/85891656-1820fc80-b7b5-11ea-9332-c4cb690e9ef6.png)

Request/response without any cookies - tab is hidden
![05-traffic-without-cookies](https://user-images.githubusercontent.com/407647/85891695-2d962680-b7b5-11ea-8c89-e665e83315b7.png)

vbuberen:
Hey @psh 
Thanks for taking initiative here.

From my point of view adding a new tab with cookies will make UX worse due to creating additional confusion for users.
We have `Request` and `Response` tabs for all corresponding info. Putting part of request or response into a separate tab isn't something that looks good for me. 

So I am for having cookies in the proper place as headers - if these are cookies for request lets put them in `Request`, if these are cookies for response we should put the in `Response`. This way library user will always know that all connected with request/response is 100% in the corresponding tab.

psh:
[...]

PR #387, opened 2020-07-11
Add cookies support (Issue #128)

psh:
## :page_facing_up: Context
This PR adds cookie support, as discussed in issue #128 

To capture cookies, Chucker needs to be configured as a `NetworkInterceptor` - *OkHttp* doesn't provide the headers in the request / response if you just call `addInterceptor()`.  This should be reflected in documentation for this feature.

## :pencil: Changes
Note: the first commit starts down the road of adding an extra tab, which was rolled back in the second commit in favor of keeping the cookies on existing _Request_ and _Response_ tabs per feedback in the issue.

## :camera: Screenshots
**Without cookies, the UI remains unchanged**

request-no-cookies

**Request and response tabs include cookies (if present)**

request-with-cookies response-with-cookies

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpCookie.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpCookie.kt
@@ -1,0 +1,26 @@
+package com.chuckerteam.chucker.internal.data.entity
+
+internal data class HttpCookie(val name: String, val value: String) {
+    companion object {
+        fun fromRequestHeader(cookieHeader: String?): List<HttpCookie> =
+            mutableListOf<HttpCookie>().apply {
+                if (!cookieHeader.isNullOrEmpty()) {
+                    cookieHeader.split(";").forEach { c ->
+                        val (name, value) = c.split("=")
+                        add(HttpCookie(name.trim(), value.trim()))
+                    }
+                }
+            }
+
+        fun fromResponseHeaders(headers: List<HttpHeader>?): List<HttpCookie> =
+            mutableListOf<HttpCookie>().apply {
+                headers?.forEach { h ->
+                    h.value.split(",").forEach { c ->
+                        val cookieParts = c.split(";")
+                        val (name, value) = cookieParts[0].split("=")
+                        add(HttpCookie(name.trim(), value.trim()))
+                    }
+                }
+            }
+    }
+}
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
@@ -182,6 +182,10 @@ internal class HttpTransaction(
         return FormatUtils.formatHeaders(getParsedRequestHeaders(), withMarkup)
     }
 
+    fun getRequestCookiesString(): String {
+        return FormatUtils.formatCookies(requestCookies)
+    }
+
     fun setResponseHeaders(headers: Headers) {
         setResponseHeaders(toHttpHeaderList(headers))
     }
@@ -192,6 +196,10 @@ internal class HttpTransaction(
 
     fun getResponseHeadersString(withMarkup: Boolean): String {
         return FormatUtils.formatHeaders(getParsedResponseHeaders(), withMarkup)
+    }
+
+    fun getResponseCookiesString(): String {
+        return FormatUtils.formatCookies(responseCookies)
     }
 
     private fun toHttpHeaderList(headers: Headers): List<HttpHeader> {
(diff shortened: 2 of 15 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpCookie.kt
```diff
@@ -0,0 +1,26 @@
+package com.chuckerteam.chucker.internal.data.entity
+
+internal data class HttpCookie(val name: String, val value: String) {
+    companion object {
+        fun fromRequestHeader(cookieHeader: String?): List<HttpCookie> =
```
vbuberen, line 5: There is some inconsistent naming between these 2 functions.

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpCookie.kt
```diff
@@ -0,0 +1,26 @@
...
+            mutableListOf<HttpCookie>().apply {
+                if (!cookieHeader.isNullOrEmpty()) {
+                    cookieHeader.split(";").forEach { c ->
+                        val (name, value) = c.split("=")
```
vbuberen, line 9: I believe these functions shouldn't be here. 
Also, the code could be simplified, since current one is cool from usage of Kotlin features point of view, but hard to read.

library/src/main/java/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -142,6 +142,18 @@ internal class HttpTransaction(
...
+    val requestCookies: List<HttpCookie> by lazy {
+        HttpCookie.fromRequestHeader(getParsedRequestHeaders()?.firstOrNull { it.name == "Cookie" }?.value)
+    }
+
+    @delegate:Ignore
+    val responseCookies: List<HttpCookie> by lazy {
+        HttpCookie.fromResponseHeaders(getParsedResponseHeaders()?.filter { it.name == "set-cookie" })
+    }
+
+    val cookiesPresent: Boolean get() = requestCookies.isNotEmpty() || responseCookies.isNotEmpty()
```
vbuberen, line 155: I don't think entity should have such functions.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -275,15 +272,47 @@ internal class TransactionPayloadFragment :
...
+    }
+
+    private fun MutableList<TransactionPayloadItem>.addCookies(cookiesString: String) {
+        if (cookiesString.isNotBlank()) {
+            add(TransactionPayloadItem.HeaderLabelItem(getString(R.string.chucker_cookies_label)))
+            add(TransactionPayloadItem.HeaderItem(cookiesString.htmlFormatted()))
+        }
+    }
+
+    private suspend fun saveToFile(
```
vbuberen, line 311: Why you decided to reformat things you didn't touch?

library/src/test/java/com/chuckerteam/chucker/internal/data/entity/HttpTransactionTest.kt
```diff
@@ -0,0 +1,174 @@
...
+        assertThat(cookies[3].name).isEqualTo("blue")
+        assertThat(cookies[3].value).isEqualTo("fish")
+    }
+
+    @Test
+    fun formatMultipleResponseCookieHeaderWithMultipleValues() {
+        val testObject = HttpTransaction().apply {
+            setResponseHeaders(
+                listOf(
+                    HttpHeader("set-cookie", "One=fish; Path=/, two=fish; Path=/"),
```
vbuberen, line 164: What is `Path=/` ? I see it everywhere in UI. I believe it shouldn't be there.

sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
```diff
@@ -120,4 +139,27 @@ class HttpBinClient(
             }
         })
     }
+
+    private class HttpBinSampleCookieJar : CookieJar {
```
vbuberen, line 143: I don't think that we need this thing at all with some specifically set data.
 
If you want to set some custom cookies just use that provided url from `httpbin.org` and put data into the query there.

psh:
Thank you all for the excellent feedback - its good to see such passionate people involved with the project and I think we've learned a lot in the process here.

I think splitting the issue has a lot of merit.  What you call out as step _i_ would directly benefit end users looking to see the difference between data collected as a _NetworkInterceptor_ or just an _Interceptor_ serving as a nice means to test other features coming along afterward.  I added #389 to cover this step.

Steps _ii_ and _iii_ definitely feel like a unit that stands together, so I quoted your feedback when I created issue #390. 

I am fine if you want to close this PR for now, and we concentrate on the 2 new issues that were created, and the ongoing discussion of what a user-interface for cookies might look like within the library itself.

cortinico:
> I am fine if you want to close this PR for now, and we concentrate on the 2 new issues that were created, and the ongoing discussion of what a user-interface for cookies might look like within the library itself.

Sure, let's close this and we can still reuse part of the implementation in this PR

PR #391, opened 2020-07-19
Do not rely on "Content-Length" for payload size.

### LINKED ISSUE #209: Payload size isn't correct for some transactions

MiSikora:
## :camera: Screenshots

| Before | After |
| --- | --- |
| Screenshot_20200719-064434 | Screenshot_20200719-064034 |

## :page_facing_up: Context

There is an open issue - #209. Also, recently the community asked for it.

## :pencil: Changes

Two main changes. I redesigned a little bit DB entities to reflect that they have payload size and not content length. This required a destructive migration. Since this is changed, the field is set after the response read is done by `ChuckerInterceptor` and it does not rely on the `Content-Length` header.

## :no_entry_sign: Breaking

No.

Run the sample app. Responses without the `Content-Length` headers should display now payload size correctly. Also, some tests were written.

## :stopwatch: Next steps

Closes #209.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
@@ -73,7 +73,7 @@ internal class TeeSource(
         upstream.close()
         if (!isClosed) {
             isClosed = true
-            callback.onClosed(sideChannel)
+            callback.onClosed(sideChannel, totalBytesRead)
         }
     }
 
@@ -90,10 +90,11 @@ internal class TeeSource(
     interface Callback {
         /**
          * Called when the upstream was closed. All read bytes are copied to the [file].
-         * This does not mean that the content of a [file] is valid. Only that the user
-         * is done with the reading process.
+         * This does not mean that the content of the [file] is valid. Only that the user
+         * is done with the reading process. [totalBytesRead] is the exact amount of data
+         * that the client downloaded even if the [file] is corrupted.
          */
-        fun onClosed(file: File)
+        fun onClosed(file: File, totalBytesRead: Long)
 
         /**
          * Called when an exception was thrown while reading bytes from the upstream
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
@@ -83,7 +83,7 @@ internal class TeeSource(
         if (!isFailure) {
             isFailure = true
             sideStream.close()
-            callback.onFailure(exception, sideChannel)
+            callback.onFailure(sideChannel, exception)
         }
     }
 
@@ -100,6 +100,6 @@ internal class TeeSource(
          * Called when an exception was thrown while reading bytes from the upstream
          * or when writing to a side channel file fails. Any read bytes are available in a [file].
          */
-        fun onFailure(exception: IOException, file: File)
+        fun onFailure(file: File, exception: IOException)
     }
 }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
@@ -90,7 +90,7 @@ internal class TeeSource(
     interface Callback {
         /**
          * Called when the upstream was closed. All read bytes are copied to the [file].
-         * This does not mean that the content of the [file] is valid. Only that the user
+         * This does not mean that the content of the [file] is valid. Only that the client
          * is done with the reading process. [totalBytesRead] is the exact amount of data
          * that the client downloaded even if the [file] is corrupted.
          */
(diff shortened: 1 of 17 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -90,10 +90,11 @@ internal class TeeSource(
...
         /**
          * Called when the upstream was closed. All read bytes are copied to the [file].
-         * This does not mean that the content of a [file] is valid. Only that the user
-         * is done with the reading process.
+         * This does not mean that the content of the [file] is valid. Only that the user
+         * is done with the reading process. [totalBytesRead] is the exact amount of data
+         * that the client downloaded even if the [file] is corrupted.
          */
-        fun onClosed(file: File)
+        fun onClosed(file: File, totalBytesRead: Long)
```
cortinico, line 97: Let's invert the parameter order either here or in the onFailure:

```kotlin
fun onClosed(file: File, totalBytesRead: Long)
fun onFailure(file: File, exception: IOException)
```

File should always be either the first or the last param.

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -198,4 +198,86 @@ class ChuckerInterceptorTest {
...
+        source.use { it.readByteString(segmentSize.toLong()) }
+
+        val transaction = chuckerInterceptor.expectTransaction()
+        assertThat(transaction.responseHeaders).contains(
+            """
+            |  {
+            |    "name": "Content-Length",
+            |    "value": "${body.size()}"
+            |  }
+            """.trimMargin()
```
cortinico, line 246: You should be able to remove the `|` and change this to `.trimIndent()`

MiSikora: It won't work, unfortunately. JSON has this structure.

```json
[
  {
    "name": "Content-Length",
    "value": "x"
  }
]
```

`.trimIndent()` would remove first indentation before `{` and `}`. And I didn't want to compare the whole content as it would make assumptions on no other headers being there.

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -83,22 +83,23 @@ internal class TeeSource(
...
+            callback.onFailure(sideChannel, exception)
         }
     }
 
     interface Callback {
         /**
          * Called when the upstream was closed. All read bytes are copied to the [file].
-         * This does not mean that the content of a [file] is valid. Only that the user
-         * is done with the reading process.
+         * This does not mean that the content of the [file] is valid. Only that the user
```
vbuberen, line 93: Minor observation - I guess `user` on this line and `client` on 95 mean the same thing. Suggest to pick one word for naming. `client` seems more suitable for this case.

library/src/test/java/com/chuckerteam/chucker/ChuckerInterceptorDelegate.kt
```diff
@@ -12,10 +12,10 @@ import okhttp3.Response
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.atomic.AtomicLong
 
-internal class ChuckerInterceptorDelegate(
-    fileFactory: FileFactory,
-    maxContentLength: Long = 250000L,
-    headersToRedact: Set<String> = emptySet()
+internal data class ChuckerInterceptorDelegate(
```
vbuberen, line 15: Not sure I understood what this change gives us. Could explain, please?

MiSikora: It comes from this test.

https://github.com/ChuckerTeam/chucker/blob/e3e259a46d057d06ca33387236a93e0ae7221d25/library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt#L269-L282

Since the interceptor is created in `setUp` there is no way of changing its input parameters. I wanted to have more granular control over that and an explicit test.

MiSikora: Now that I'm thinking `copy` is not needed. I could just create a new interceptor. What do you think?

vbuberen: It might be a question of preference. I think creation of a new interceptor would be sort of more understandable for a new reader of the code.

PR #392, opened 2020-07-19
Add sample for partial body download.

MiSikora:
## :camera: Screenshots

Screenshot_20200719-064819
Screenshot_20200719-064832

## :page_facing_up: Context

I wanted to have a sample for a partial download of the content. It is somewhat a follow-up to #384.

## :pencil: Changes

I added a new call. It uses Postman instead of httpbin. The reason is that httpbin formats the echoed response in a really strange way as it escapes newline and quotation characters. Also, the cool thing about Postman is that it does not have the `Content-Length` header.

Additional things that I changed is the order of interceptors in the sample. Logging interceptor downloads all of the data before the Chucker can do its thing so this made partial reads impossible on the Chucker level. For this reason, I had to change all responses from `Void` to `Any?` in order to make `Retrofit` automatically read whole bodies of responses and not shortcircuit them.

## :paperclip: Related PR

#384.

## :no_entry_sign: Breaking

No.

Just run the sample and find the request in the list.

### DIFF
--- a/sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
+++ b/sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
@@ -6,8 +6,10 @@ import com.chuckerteam.chucker.api.ChuckerCollector
 import com.chuckerteam.chucker.api.ChuckerInterceptor
 import com.chuckerteam.chucker.api.RetentionManager
 import com.chuckerteam.chucker.sample.HttpBinApi.Data
+import okhttp3.MediaType
 import okhttp3.OkHttpClient
 import okhttp3.Request
+import okhttp3.RequestBody
 import okhttp3.logging.HttpLoggingInterceptor
 import retrofit2.Call
 import retrofit2.Callback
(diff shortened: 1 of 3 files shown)


sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
```diff
@@ -120,4 +124,19 @@ class HttpBinClient(
             }
         })
     }
+
+    private fun getResponsePartially() {
```
cortinico, line 128: nit: I think we should consider extracting this to a separate class.
We have this class called `HttpBinClient` but starts to have a lot of sample HTTP calls to other endpoints.

MiSikora: Yes, you're right. It makes sense to not evolve this into a ball of mud. I, or any other volunteer, can make it in this PR or a follow-up one. I agree that it is not crucial to do it right now.

vbuberen: I support the idea of doing such change in another PR.

PR #396, opened 2020-07-26
Do not fail requests when side channel file cannot be used

### LINKED ISSUE #394: FileNotFoundException drops the request
After performing one of the requests in my app I receive that kind of error which results in failed request:

```
java.io.FileNotFoundException: /data/user/0/app.name.dev/cache/chucker-21 (No such file or directory)
        at java.io.FileOutputStream.open0(Native Method)
        at java.io.FileOutputStream.open(FileOutputStream.java:308)
        at java.io.FileOutputStream.<init>(FileOutputStream.java:238)
        at okio.Okio__JvmOkioKt.sink(JvmOkio.kt:170)
        at okio.Okio.sink(Unknown Source:1)
        at okio.Okio__JvmOkioKt.sink$default(JvmOkio.kt:170)
        at okio.Okio.sink$default(Unknown Source:1)
        at okio.Okio.sink(Unknown Source:3)
        at com.chuckerteam.chucker.internal.support.TeeSource.<init>(TeeSource.kt:27)
        at com.chuckerteam.chucker.api.ChuckerInterceptor.multiCastResponseBody(ChuckerInterceptor.kt:184)
        at com.chuckerteam.chucker.api.ChuckerInterceptor.intercept(ChuckerInterceptor.kt:93)
```

Repeating the same request doesn't solve this problem, but performing a different request does.

## :bomb: Steps to reproduce

It happens after giving the control off app to another [third-party library](https://github.com/idnow/de.idnow.android). I cannot tell you more concrete info because I really don't know what happens inside this library. Maybe it clears all caches of app which then results in Chucker error.

## :wrench: Expected behavior

Anyway it seems that dropping the request with internal error is not the best option. A warning should be enough even though the request wouldn't be logged.

Tech info - different devices, different OS versions

MiSikora:
## :page_facing_up: Context

There is an open issue - #394.

## :pencil: Changes

`TeeSource` guards against failures to use files. When `FileNotFoundException` occurs when opening a stream it is reported back to the caller and no payload side streaming happens. Also, any IO failures during interception are now logged.

## :no_entry_sign: Breaking

No.

Not really sure how to reproduce this manually without a convoluted setup. Maybe @CodeBreak524 can confirm that their issue is fixed with a dependency from my branch on JitPack (`com.github.MiSikora.chucker:library:side-channel-file-failure-SNAPSHOT`). I don't really feel that it is necessary as the description in #394 is quite clear and it is now covered with tests.

## :stopwatch: Next steps

Closes #394.

To be honest I wonder now about the necessity to write to a file instead of to the memory. Responses need to be loaded to a memory at some point anyway.

https://github.com/ChuckerTeam/chucker/blob/ba2c21b4f5cf7d04363626f49a411bb7f0dfddb0/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt#L251-L259

I pushed for a file-based solution before because of parallel downloads but I didn't consider that the content of files needs to be read to the memory before it is saved to the DB. However, there is still an advantage of files that could be leveraged. Files could be saved to some non-volatile directory instead of cache and DB could just store handles to these files (processing still would be required as we need to decompress network streams). I see the pros and cons of both solutions and I'm open to suggestions.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -246,7 +246,7 @@ class ChuckerInterceptor internal constructor(
             file.delete()
         }
 
-        override fun onFailure(file: File, exception: IOException) = Unit
+        override fun onFailure(file: File, exception: IOException) = exception.printStackTrace()
 
         private fun readResponseBuffer(responseBody: File, isGzipped: Boolean): Buffer {
             val bufferedSource = Okio.buffer(Okio.source(responseBody))
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
@@ -5,6 +5,7 @@ import okio.Okio
 import okio.Source
 import okio.Timeout
 import java.io.File
+import java.io.FileNotFoundException
 import java.io.IOException
 
 /**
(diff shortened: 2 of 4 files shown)


cortinico:
> However, there is still an advantage of files that could be leveraged. Files could be saved to some non-volatile directory instead of cache and DB could just store handles to these files (processing still would be required as we need to decompress network streams).

As you said, responses would have to be loaded in memory anyway (e.g. you have to show an image, parse the body of a JSON to split it in strings, etc). I feel that the file approach we have right now is way more robust than before, where we just loaded everything in memory with subsequent OOM errors.

I'll suggest that we stick to it and try to refine it following issue reports (as we're doing) 👍

vbuberen:
Agree with @cortinico here that current approach looks pretty robust for now.

Issue #248, opened 2020-02-22
Add pre-commit hooks to run code checks


vbuberen:
## :warning: Is your feature request related to a problem? Please describe
It seems that contributors often have to do multiple commits, since their contributions have issues with styling etc.

## :bulb: Describe the solution you'd like
It might be more convenient to run `ktlintCheck` and `detekt` tasks as a pre-commit hook. With these changes contributors could be sure that their code passes Chucker checks and also make our library build process a little more eco-friendly, since we won't have to spin up Travis CI that often 🌲

## :page_facing_up: Additional context
We might also include `test` step as well.

## :raising_hand: Do you want to develop this feature yourself?

- [x] Yes
- [ ] No

cortinico:
++ for pre-commit hooks. Are you planning to use some tool to distribute them @vbuberen or how do we architect this?

vbuberen:
I wanted to do something simple to not introduce any additional tool. 
Something like [standard hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) should work fine.

cortinico:
> I wanted to do something simple to not introduce any additional tool.
> Something like [standard hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) should work fine.

Yes my point was more that hooks are not part of the git repo (they're not applied when the repo is cloned). So we need some sort of tools to install them on the first build or similar.

ColtonIdle:
ktlint-gradle might be able to help

https://github.com/jlleitschuh/ktlint-gradle

> addKtlintFormatGitPreCommitHook

I think they recommended (overheard in kotlinglang slack) adding that to the clean task in your gradle project so that it gets added to everyones machine.

vbuberen:
Thanks for sharing. Looks interesting.

cortinico:
I'm adding a note for this in the README 👍

vbuberen:
I am currently playing around with pre-commit hook, which is installed from `clean` task as suggested, which runs both `detekt` and `ktlintCheck` commands. Let's install it automatically, since I don't hope that people will actually read these recomendations.

vbuberen:
Closed by #408

PR #407, opened 2020-08-03
Add a Building section to the README

cortinico:
## :page_facing_up: Context
This is related to #248, I'm adding a Building section to the README to clarify how to setup the local environment for developing the library.

## :pencil: Changes
- Added the `Building` with a mention on the ktlint pre-commit hook
- Added the `gradlew build` suggested command in order to run all the tests locally
- Disabled `GradleDependency` lint check as `gradlew build` is failing on master due to obsolete OkHTTP dependency (that's intentional).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ _A fork of [Chuck](https://github.com/jgilfelt/chuck)_
 * [Snapshots](#snapshots-)
 * [FAQ](#faq-)
 * [Contributing](#contributing-)
+  * [Building](#building-)
 * [Acknowledgments](#acknowledgments-)
 * [License](#license-)
 
@@ -187,6 +188,27 @@ Short `TODO` List for new contributors:
 
 - Increment the test coverage.
 - [Issues marked as `Help wanted`](https://github.com/ChuckerTeam/chucker/labels/help%20wanted)
+
+### Building 🛠
+
+In order to start working on Chucker, you need to fork the project and open it in Android Studio/IntelliJ IDEA.
+
+Before committing we suggest you install the pre-commit hooks with the following command:
+
+```
+./gradlew installGitHook
+```
+
+This will make sure your code is validated against KtLint and Detekt before every commit.
+
+Before submitting a PR please run:
+
+```
+./gradlew build
+```
+
+This will build the library and will run all the verification tasks (ktlint, detekt, lint, unit tests) locally.
+This will make sure your CI checks will pass.
 
 ## Acknowledgments 🌸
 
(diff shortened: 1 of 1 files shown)


vbuberen:
I would put it under `Contributing` section.
Also, as I mentioned in #248 I would like to add automatically installed pre-commit hook instead of hoping that people will read and follow.

cortinico:
> I would put it under `Contributing` section.

It's actually already there (it's a h3 inside the h2 Contributing)

library-no-op/build.gradle
```diff
@@ -15,6 +15,7 @@ android {
     lintOptions {
         warningsAsErrors true
         abortOnError true
+        disable 'GradleDependency'
```
vbuberen, line 18: Why do we need this? To not show warnings about future Gradle versions, like we have now?

README.md
```diff
@@ -188,6 +189,28 @@ Short `TODO` List for new contributors:
 - Increment the test coverage.
 - [Issues marked as `Help wanted`](https://github.com/ChuckerTeam/chucker/labels/help%20wanted)
 
+### Building 🛠
```
vbuberen, line 192: Probably, we should update this section to mention about pre-commit hook from #408

library-no-op/build.gradle
cortinico: > Why do we need this?

We need this as currently lint is failing on master (as we don't run it on CI) because of this.
In my change I'm suggesting to run `./gradlew build` in the readme, but that command will actually fail as of now.

Either we suppress `GradleDependency` globally or we have to comment every dependency in the .gradle file with `//noinspect GradleDependency`. I generally disable this Lint check and let Gradle (or the other tools like Dependabot) take care of the dependencies.

Anyway I'm amending this change as I'll take care of the lint failures in another PR

README.md
cortinico: Updated

README.md
```diff
@@ -188,6 +189,27 @@ Short `TODO` List for new contributors:
...
+
+In order to start working on Chucker, you need to fork the project and open it in Android Studio/IntelliJ IDEA.
+
+Before committing we suggest you install the pre-commit hooks with the following command:
```
vbuberen, line 196: Let's mention that the hook will be installed automatically during the first build.

README.md
```diff
@@ -188,6 +189,28 @@ Short `TODO` List for new contributors:
...
+
+```
+./gradlew installGitHook
+```
+
+This will make sure your code is validated against KtLint and Detekt before every commit.
+The command will run automatically before the `clean` task, so you should have the pre-commit installed by then.
```
vbuberen, line 203:
```suggestion
The command will run automatically before the `clean` task, so you should have the pre-commit hook installed by then.
```

Issue #249, opened 2020-02-22
Add code style config into the project


vbuberen:
## :warning: Is your feature request related to a problem? Please describe
Often contributors struggle with following any particular style since Chucker has no style defined.

## :bulb: Describe the solution you'd like
Create a default code style config and push into the repo, so everyone is able to apply it during work on their contributions without trying to adjust it manually.

## :page_facing_up: Additional context
Probably, we could apply a combination of official [Kotlin Android code style](https://developer.android.com/kotlin/style-guide) from Google along with some others, since the mentioned guide doesn't cover things, like `xml` files formatting.

## :raising_hand: Do you want to develop this feature yourself?

- [ ] Yes
- [x] No

cortinico:
I'm in for having this. I would just love to make sure that he source of thruth is not the codeStyleConfig file. 

Ideally we want to make sure that our CI is able to verify the codestyle, so having the `ktlintCheck` gradle taks running on the CI is a great way to achieve so.

I have nothign against having `.idea` commited files, just make sure we don't rely on IDEs to enforce a common codestyle (aka is a _nice to have_).

ibrahim-iqbal:
Just a heads-up: the repo already has what this issue was asking for. On `develop` today:

- `.editorconfig` at the root
- `detekt-config.yml` at the root, wired via `alias(libs.plugins.detekt) apply false` in the root `build.gradle.kts`
- `ktlint` on the same alias line, so both plugins are available to modules

That covers both halves of the ask — a shared style config in the repo and a checker the CI can run. This can probably be closed unless there's a specific gap I'm missing.

cortinico:
> That covers both halves of the ask — a shared style config in the repo and a checker the CI can run. This can probably be closed unless there's a specific gap I'm missing.

Yup this can be closed now

PR #408, opened 2020-08-03
Add pre-commit hook and task to automatically install it

vbuberen:
## :page_facing_up: Context
This PR implements enhancement #248 to run code checks before commiting the code. Except adding the hook itself I have also added a task to install hook which runs whenever clean task is fired. It means that with such hack every contributor will have hook installed automatically.

## :pencil: Changes
- Added `pre-commit` script file
- Added task to install this script file
- Set `clean` task to depend on install task.

## :no_entry_sign: Breaking
Nothing

Run `Clean` or `Rebuild` task and see that `installGitHook` task is executed as well.
After that hook should run on every commit attempt.

## :stopwatch: Next steps
We might consider integrating formatting task into the hook as well after adding #249

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -1,0 +1,20 @@
+#!/bin/bash
+echo "Running pre-commit checks..."
+
+OUTPUT="/tmp/analysis-result"
+./gradlew detekt ktlintCheck > ${OUTPUT}
+EXIT_CODE=$?
+if [ ${EXIT_CODE} -ne 0 ]; then
+    cat ${OUTPUT}
+    rm ${OUTPUT}
+    echo "*********************************************"
+    echo "               Checks Failed!                 "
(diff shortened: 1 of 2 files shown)


pre-commit
```diff
@@ -0,0 +1,20 @@
...
+echo "Running pre-commit checks..."
+
+OUTPUT="/tmp/analysis-result"
+./gradlew detekt ktlintCheck > ${OUTPUT}
+EXIT_CODE=$?
+if [ ${EXIT_CODE} -ne 0 ]; then
+    cat ${OUTPUT}
+    rm ${OUTPUT}
+    echo "*********************************************"
+    echo "               Checks Failed!                 "
```
cortinico, line 11:
```suggestion
    echo "               Checks Failed!                "
```
nit: line is too long

Issue #259, opened 2020-03-01
Redesign HTTP transaction model


MiSikora:
## :warning: Is your feature request related to a problem? Please describe
Right now model for an HTTP transaction is a big structure that kind of serves as bag for everything. It makes it hard to work on new features that require new fields, different encoding format etc. Some links - #204, #202, #69.

## :bulb: Describe the solution you'd like
I believe that there should be a separate table for transactions and different types of requests and responses. This way it would be easier to add new types of compression and so on.

This would require to not rely in UI layer on DB entities. They should be rather mapped to some common model (most likely a sealed class or an interface) that would be renderable on the view layer.

It would also allow to tidy a bit nullability of some fields. While request or response still might be null, some of their fields shouldn't be (like i.e. URL address).

## :bar_chart: Describe alternatives you've considered
An interesting alternative would be to split HTTP transaction just into three tables (transaction, request, response) and store data as a blob which would be defined in a proto file. There should be no issue with backward compatibility as this is one of main advantages of protobuf. Additional benefit is that new tables would not appear for every new feature and only the proto schema would change.

## :raising_hand: Do you want to develop this feature yourself?
- [x] Yes
- [ ] No

I could work on it but it is rather a big change proposal that requires some discussion, design and approval beforehand. Not sure what the best way would be. This ticket, draft PR, Slack channel or maybe something else?

cortinico:
If I read you correctly, you're suggesting we treat the HTTP entities as immutable. If so, I'm totally in for this.

Currently we do have a `transaction` table that stores a HTTP entity both for request and response and we update it once we collect the data. Ideally having separate tables will allow to make sure we treat those entities as immutable (with all the benefits of immutability).

> This would require to not rely in UI layer on DB entities. They should be rather mapped to some common model (most likely a sealed class or an interface) that would be renderable on the view layer.

Can you elaborate further? Or provide an example (e.g. adding SSL/TLS information as done in #252). How that would look like in your suggestion.

Moreover, let's put Testing in the roadmap as we have much to improve on that side 👍

vbuberen:
> Moreover, let's put Testing in the roadmap as we have much to improve on that side 👍

I thought that we already agreed about it in Slack. On 21st of January I wrote such message for plans after `3.1.0`:

- Cover library with tests completely and force tests run on each PR in our pipeline. We could start with api package as it is the core functionality and move up to UI as a final stage. We definitely need tests to be sure in quality as well as get more credibility from our users. 
[...]

PR #411, opened 2020-08-04
Add port to database and UI.

### LINKED ISSUE #403: URL does not show port number
If a URL includes a port number, it should be shown as part of the url, just like charles proxy

JayNewstrom:
Fixes #403

## :camera: Screenshots
Screenshot_20200803-215033

## :page_facing_up: Context
#403

## :no_entry_sign: Breaking
This will cause the database to be dropped/recreated on the next run.

I updated the automated tests, as well as tested manually via the UI (screenshot attached)

## :stopwatch: Next steps
Maybe add something to release notes about the database being dropped.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
@@ -20,7 +20,6 @@ internal object ShareUtils {
 
     fun getShareText(context: Context, transaction: HttpTransaction, encodeUrls: Boolean): String {
         var text = "${context.getString(R.string.chucker_url)}: ${transaction.getFormattedUrl(encodeUrls)}\n"
-        text += "${context.getString(R.string.chucker_port)}: ${transaction.port}\n"
         text += "${context.getString(R.string.chucker_method)}: ${transaction.method}\n"
         text += "${context.getString(R.string.chucker_protocol)}: ${transaction.protocol}\n"
         text += "${context.getString(R.string.chucker_status)}: ${transaction.status}\n"
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
@@ -94,7 +94,7 @@ internal object ShareUtils {
             // try to keep to a single line and use a subshell to preserve any line breaks
             curlCmd += " --data $'${requestBody.replace("\n", "\\n")}'"
         }
-        curlCmd += (if (compressed) " --compressed " else " ") + transaction.getFormattedUrl(false)
+        curlCmd += (if (compressed) " --compressed " else " ") + transaction.getFormattedUrl(encode = false)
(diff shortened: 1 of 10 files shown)


vbuberen:
I guess this is not what was requested in #403 
The idea was to show port as part of URL, like Charles Proxy does. 

So the idea is to add port in to `URL` field on `Overview` and on `Network` screen, since it is really a part of URL

JayNewstrom:
I've update the port to be stored/shown with the URL. This makes it so that it behaves more inline with the other parts of the URL, and also makes it so there's no breaking database changes.

vbuberen:
I have 2 concerns here:
1. We might not want to always show port. For example, Charles Proxy shows port numbers not always, but when they are really part of the url (at least this is what it looks like to me). With current implementation we will always show `443` or `80` for https/http requests respectively.
2. If we add port into url we should also show the same url on the main screen with list of transactions.

JayNewstrom:
> If we add port into url we should also show the same url on the main screen with list of transactions.
If we want to add it to the transactions list, we'd have to read/parse the url in `HttpTransactionTuple`, which I'm guessing is not desired for scrollability/speed. If you decide it should go there as well after this new info, let me know and I'll update it.

JayNewstrom:
The other (hacky) thing I could do is store the port with the `host`, in the displayable format, since it looks like that is only used for that one thing.

vbuberen:
> If we want to add it to the transactions list, we'd have to read/parse the url in `HttpTransactionTuple`, which I'm guessing is not desired for scrollability/speed. If you decide it should go there as well after this new info, let me know and I'll update it.

No need to update, from my point of view. We are going to redesign our DB entities for 4.x release (which we plan after 3.3.0). And in that change of DB it highly likely that we will have separate entities for transaction itself, request and response (at least this is what I am still thinking about and we discuss in #259 )
Except that I am going to do some changes on that transactions list as well.

As to your change about omitting 443 and 80 - looks good. I am curious to get more opinions there, but for me current options looks good enough.

JayNewstrom:
Sounds good. Let me know if there's anything else I can do.

library/src/main/java/com/chuckerteam/chucker/internal/support/ShareUtils.kt
```diff
@@ -94,7 +94,7 @@ internal object ShareUtils {
             // try to keep to a single line and use a subshell to preserve any line breaks
             curlCmd += " --data $'${requestBody.replace("\n", "\\n")}'"
         }
-        curlCmd += (if (compressed) " --compressed " else " ") + transaction.url
+        curlCmd += (if (compressed) " --compressed " else " ") + transaction.getFormattedUrl(false)
```
MiSikora, line 97: I suggest using a named parameter here.

JayNewstrom: Sure, updated.

ColtonIdle:
Looks great and yes, having it part of the URL matches what I've gotten used to with Charles, so I think that makes sense.

PR #418, opened 2020-08-07
Make some dependencies api in library build.gradle

### LINKED ISSUE #415: Audit build.gradle for items that should be api instead of implementation
Right now everything in `library/build.gradle` is defined as `implementation`. Some of these should be `api`. Any types that can be accessed via public APIs should be declared as `api`. For instance, because `ChuckerInterceptor` is public, and implements the OkHttp interface `Interceptor`, OkHttp should be declared as an `api` dependency in the build.gradle.

JayNewstrom:
Fixes #415

These are the only 2 dependencies that have their types exposed by Chucker public types. This ensures that downstream apps that use these APIs will compile.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -40,7 +40,7 @@ android {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
+    api "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
 
     implementation "com.google.android.material:material:$materialComponentsVersion"
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutVersion"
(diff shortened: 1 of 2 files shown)


library/build.gradle
```diff
@@ -40,7 +40,7 @@ android {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
+    api "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
```
MiSikora, line 43: How Kotlin is exposed?

JayNewstrom: This is one I went on the side of caution for. The only place I could see that might have potential impact was `ChuckerInterceptor` where the set of headers to redact is exposed. I don't have a java project to easily test this with.

In practice, this probably isn't an issue.

MiSikora: Yes, there is no need for that. `kotlin.collections.Set` is seen as `java.util.Set` from Java.

vbuberen (approved):
Thanks for contribution. LGTM! 🚢

PR #423, opened 2020-08-13
Add failing test demonstrating #422

koral--:
PR only for bug #422  reproduction purposes. Not for direct merging.

Issue #413, opened 2020-08-05
Crash with RuntimeException: StrictMode ThreadPolicy violation


noloman:
## :writing_hand: Describe the bug

I'm using Chucker in an app with Dagger 2 and Android's StrictMode with the option `StrictMode.ThreadPolicy.Builder().detectDiskWrites()`, and I get the following crash: `java.lang.RuntimeException: StrictMode ThreadPolicy violation` pointing to my this code in my `NetworkModule`:
```
@Provides
@Singleton
fun provideChuckerInterceptor(context: Context): ChuckerInterceptor = ChuckerInterceptor(context)
```
Nevertheless removing the `detectDiskWrites()` removes the crash, but I still want to have this option enabled for my app.

Thanks for the great library!

vbuberen:
Thanks for reporting. Yes, this issue is present in 3.2.0, but it is already resolved in 3.3.0-snapshot. So feel free to use snapshot if you need to have strict mode enabled. Snapshot has more goodies besides this fix, so give it a try :)

MiSikora:
I don't believe this is the same as #311 or #318. Unless this is about some other fix I don't recall?

I'd appreciate, @noloman, if you could share the stack trace. I'm not sure about the snippet you showed, as the only potential IO thing that happens there is DB creation/deletion and opening SharedPreferences. Also, this should not crash the app under the `StrictMode` policy.

Most likely what happens is that you read your response on the main thread. Chucker does not make any assumptions about the thread that your app uses and it will write to and read from temporary, internal files while you read your responses from the backend. In such a case it is a good thing that this was caught.

vbuberen:
Right, I might have mixed something about issues in Chucker and some other project. Thanks for catching.

In such case yeah, we need stacktrace to understand what happens.

koral--:
I guess that crash can be caused by `penaltyDeath()` in StrictMode configuration and the violation may come from here:
https://github.com/ChuckerTeam/chucker/blob/develop/library/src/main/java/com/chuckerteam/chucker/internal/support/AndroidCacheFileFactory.kt#L12 (in turn called from here: https://github.com/ChuckerTeam/chucker/blob/develop/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt#L64)
It performs I/O: http://androidxref.com/6.0.0_r1/xref/frameworks/base/core/java/android/app/ContextImpl.java#createFilesDirLocked
[...]

PR #427, opened 2020-08-17
File creation fix

### LINKED ISSUE #414: Potential NPE during AndroidCacheFileFactory initialization
The current code at: https://github.com/ChuckerTeam/chucker/blob/develop/library/src/main/java/com/chuckerteam/chucker/internal/support/AndroidCacheFileFactory.kt#L17 assumes that `context.cacheDir` returns non-null value. However, on Android older than 7.0 this method can return `null`: http://androidxref.com/6.0.0_r1/xref/frameworks/base/core/java/android/app/ContextImpl.java#425

## :bomb: Steps to reproduce
1. Cause cache dir to be not creatable eg. create regular file with the same name beforehand or fill entire disk space.
2. Invoke `ChuckerInterceptor` constructor.

## :wrench: Expected behavior
Not sure what is that cache used for. If it is mandatory, NPE could be wrapped by IOException.

## :camera: Screenshots
N/A

## :iphone: Tech info
 - OS: < 7.0
 - Chucker version: at least one which is built from commit https://github.com/ChuckerTeam/chucker/commit/392a63b8043f3c967558a49bf84db481870537a2 or newer.

I haven't encountered this issue at runtime. Just realized that it may happen during https://github.com/ChuckerTeam/chucker/issues/413 investigation.

### LINKED ISSUE #422: IOException when app uses 3rd party SDK
This issue arises from #394 

After fixes for #394 the reporter [provided update](https://github.com/ChuckerTeam/chucker/issues/394#issuecomment-671868443) that another exception appeared:

> ```
>  java.io.IOException: No such file or directory
>         at java.io.UnixFileSystem.createFileExclusively0(Native Method)
>         at java.io.UnixFileSystem.createFileExclusively(UnixFileSystem.java:281)
>         at java.io.File.createNewFile(File.java:1008)
>         at com.chuckerteam.chucker.internal.support.AndroidCacheFileFactory.create(AndroidCacheFileFactory.kt:21)
>         at com.chuckerteam.chucker.internal.support.AndroidCacheFileFactory.create(AndroidCacheFileFactory.kt:15)
>         at com.chuckerteam.chucker.api.ChuckerInterceptor.multiCastResponseBody(ChuckerInterceptor.kt:184)
>         at com.chuckerteam.chucker.api.ChuckerInterceptor.intercept(ChuckerInterceptor.kt:92)
> ```
> It happens even after removing the application from device and reinstalling a new one.

## :wrench: Expected behavior
Chucker works normally and shows transactions.

## :iphone: Tech info
 - Chucker version: 3.3.0-SNAPSHOT (as of 11th of August 2020)

We should investigate the issue to either fix it or to get all the info we might need to deal with possible problems with other 3rd party SDKs.

MiSikora:
## :page_facing_up: Context

There are open issues - #422, #414.

## :pencil: Changes

`FileFactory` is now IO safe and typesafe. I removed the interface in favor of a concrete implementation. It was internal anyway and it was misleading in tests of `ChuckerInterceptor` because fake implementation did not take into account stuff like `createNewFile()`, that was the cause of #422. 

One more thing that I changed in the implementation of the factory is that access to the parent directory. It is now lazy, so it should help with #413 since the cache directory will no longer be accessed during initialization. It won't fix it as there are other files being used on the main thread but it is one step forward.

~~I'm not particularly happy about the fact that `TeeSource` accepts now `File?` as an input parameter but I didn't want to overcomplicate things by introducing another source that just counts bytes downloaded.~~ - This isn't the case anymore. See https://github.com/ChuckerTeam/chucker/pull/427#issuecomment-677716291.

## :paperclip: Related PR

I copied the test for the file factory from #423.

## :no_entry_sign: Breaking

No.

In practice not sure. Maybe @CodeBreak524 can test with a dependency from my branch on JitPack (`com.github.MiSikora.chucker:library:file-creation-fix-SNAPSHOT`). I didn't write new tests for `ChuckerInterceptor` as old ones cover this case due to using real file factory in tests.

## :stopwatch: Next steps

Closes #422
Closes #414
Closes #423


library/src/main/java/com/chuckerteam/chucker/internal/support/FileFactory.kt
```diff
@@ -1,8 +1,29 @@
...
+
+    fun create(filename: String): File? = try {
+        val directory = getFileDirectory()
+            ?: throw FileNotFoundException("Failed to create directory for temporary Chucker files")
+        File(directory, filename).apply {
+            if (exists()) {
+                delete()
+            }
+            parentFile?.mkdirs()
+            createNewFile()
```
koral--, line 23: Shouldn't we fail fast here if this invocation returned `false`?

MiSikora: I'm not sure. `createNewFile()` returns `false` only if the file already exists according to documentation. In that case, it might be more accurate to do something like this.

```kotlin
if (exists() && !delete()) {
  throw IOException("Failed to delete temporary Chucker file $file")
}
```

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -234,28 +233,26 @@ class ChuckerInterceptor internal constructor(
...
     ) : TeeSource.Callback {
 
-        override fun onClosed(file: File, totalBytesRead: Long) {
-            val buffer = try {
-                readResponseBuffer(file, response.isGzipped)
-            } catch (e: IOException) {
-                null
-            }
+        override fun onClosed(file: File?, totalBytesRead: Long) {
+            val buffer = file?.let { readResponseBuffer(it, response.isGzipped) }
```
vbuberen, line 237: I would suggest to avoid `?.let { }` here and use just `if (file != null) {...` to not introduce additional useless variable multiple times (as this fun will be called for each transaction).
Suggestion is based on [this article](https://medium.com/mobile-app-development-publication/kotlin-dont-just-use-let-7e91f544e27f) and recommendation 1.1 from the mentioned article.

library/src/main/java/com/chuckerteam/chucker/internal/support/TeeSource.kt
```diff
@@ -18,12 +18,12 @@ import java.io.IOException
...
 internal class TeeSource(
     private val upstream: Source,
-    private val sideChannel: File,
+    private val sideChannel: File?,
     private val callback: Callback,
     private val readBytesLimit: Long = Long.MAX_VALUE
 ) : Source {
     private val sideStream = try {
-        Okio.buffer(Okio.sink(sideChannel))
+        sideChannel?.let { Okio.buffer(Okio.sink(it)) }
```
vbuberen, line 26: Same suggestion about using `if (sideChannel != null)` as in the previous comment.

vbuberen (commented):
Thanks for such a good change. Looks good to me. 
Only a few minor comments with suggestions for micro-optimizations.

Also, I see a typo in PR description with wrong PR mentioned in the first part in `that was the cause of #420.` Please update it as well to have proper history descriptions.

MiSikora:
Ok, I applied your suggestions. I would appreciate it if you can give feedback on https://github.com/ChuckerTeam/chucker/pull/427#discussion_r471637023 from @koral-- and how you'd like to handle it.

MiSikora:
Coverted it to a draft because I discovered some issues.

MiSikora:
In 130d59d I decoupled tee operation from reporting functionality. Reporting is now a part of a `ReportingSink` class. My main motivation was that `TeeSource` started to become too complex and was had too many responsibilities.

cortinico:
> CI fails for now because I added a functional interface and #438 needs to be merged first. I'll rebase after. Unless you think that `() -> File?` and a type alias will suffice.

Since we're on 1.4, let's go with a functional interface 👌

cortinico (approved):
Left some nits. Overall the change looks good to me. 

Thank you very much for taking the time to do this change @MiSikora 🙏 
I'd love to get a pass also from @vbuberen if possible before merging 👌

library/src/main/java/com/chuckerteam/chucker/internal/support/CacheDirectoryFactory.kt
```diff
@@ -0,0 +1,7 @@
+package com.chuckerteam.chucker.internal.support
+
+import java.io.File
+
+internal fun interface CacheDirectoryFactory {
+    fun create(): File?
```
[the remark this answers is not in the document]
MiSikora: 👍  I even named it `Provider` in the first implementation, but didn't want to add a mental overhead of differentiation between factories and providers.

MiSikora:
Great, thanks! I added also one test for a missing cache in the first place.

vbuberen (approved):
LGTM!

PR #363, opened 2020-06-05
Enable Explicit Api mode for Chucker

cortinico:
## :page_facing_up: Context

I'm enabling strict explicit API check for Chucker. This will help us make sure we specify visibility of whatever we expose.

This feature is available only on Kotlin 1.4x so we can't merge this at this point.

## :stopwatch: Next steps
We can already address the warning actually. Once we bump to 1.4, we can just enable explicit API and have a green build.

PR #430, opened 2020-08-18
Update Kotlin to 1.4.0

vbuberen:
## :page_facing_up: Context
Update Chucker to use Kotlin 1.4.0. It seems that after this update we don't need different Kotlin versions in our workflow anymore.

## :pencil: Changes
- Update Kotlin to 1.4.0
- Remove environment variable for different Kotlin versions from Github Actions workflow.

## :paperclip: Related PR
Unblocks #363 

## :no_entry_sign: Breaking
Nothing expected.

## :stopwatch: Next steps
Update coroutines dependency to 1.3.9, which already switched to Kotlin 1.4.0

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -11,15 +11,6 @@ on:
 jobs:
   test:
     runs-on: [ubuntu-latest]
-    strategy:
-      fail-fast: false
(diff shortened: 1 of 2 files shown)


olivierperez (approved):
LGTM :+1:

Issue #425, opened 2020-08-14
response json body is not show when parse error


czy1121:
## :writing_hand: Describe the bug
response json body is not show

## :bomb: Steps to reproduce

```
data class SomeData(val someValue: Boolean)

interface SomeService {
    @GET("test")
    fun config(): Call<SomeData>  
}
```

if server return `{"someValue": true}`, i will see it in chucker,
if server return `{"someValue": 1}`, i can not see it in chucker, i need it

MiSikora:
Can you please create a sample test that would show this behaviour? This one passes correctly.

```kotlin
data class SomeData(val someValue: Boolean)

interface SomeService {
    @GET("/")
    fun config(): Call<SomeData>
}

@ParameterizedTest
@EnumSource(value = ClientFactory::class)
fun test(factory: ClientFactory) {
    val body = Buffer().apply { writeUtf8("""{"someValue": 1}""") }
    server.enqueue(MockResponse().setBody(body).addHeader("Content-Type", "application/json"))

    val service = Retrofit.Builder()
            .client(factory.create(chuckerInterceptor))
            .addConverterFactory(GsonConverterFactory.create())
            .baseUrl(server.url("/"))
            .build()
            .create<SomeService>()

    val latch = CountDownLatch(1)
    service.config().enqueue(object : Callback<SomeData> {
        override fun onFailure(call: Call<SomeData>, t: Throwable) {
            latch.countDown()
        }

        override fun onResponse(call: Call<SomeData>, response: Response<SomeData>) {
            throw AssertionError()
        }
    })

    latch.await()
    val transaction = chuckerInterceptor.expectTransaction()

    assertThat(transaction.responseBody).isEqualTo("""{"someValue": 1}""")
}
```

If you use `execute` instead of `enqueue` you need to catch exceptions that your JSON converter might throw. Alternatively you should wrap your data in `Response` and use `fun config(): Call<Response<SomeData>>`.

vbuberen:
Thanks @MiSikora for such a fast reply.

My suggestion is also about JSON converter issues there.

czy1121:
@MiSikora 
i use Moshi 

when server return {"someValue": 1}, Moshi throw JsonDataException, it's ok

i want `{"someValue": 1}` show in chucker, but not
![Lark20200815-034041](https://user-images.githubusercontent.com/1754323/90286787-2207d780-dea9-11ea-905c-e420e3ed8984.jpeg)

[...]

PR #443, opened 2020-08-25
Add a flag to always read response body fully

### LINKED ISSUE #432: Allow to read the whole response body irrespective of the consumer

In the light of the discussion on Slack there is a reasonable demand to read the response body even if a client did not wish to consume it or failed to consume it. This is also kind of a legacy from Chuck that is expected to be preserved.

https://kotlinlang.slack.com/archives/CRWD6370R/p1597269092105200
https://kotlinlang.slack.com/archives/CRWD6370R/p1597812748114900

I'd like to introduce a new flag for Chucker configuration that would read the whole response body regardless of the underlying work of external libraries like Retrofit or JSON converters. Conceptually it might look like this. If the `alwaysReadResponseBody` flag is set to `true` then upon creating the upstream `TeeSource` would be wrapped around the `DepletingSource`. This might also help with #425.

```kotlin
internal class DepletingSource(delegate: Source) : ForwardingSource(delegate) {
  override fun close() {
    try {
      Okio.buffer(delegate()).readAll(Okio.blackhole())
    } catch (_: IOException) {

    }
    super.close()
  }
}
```

I don't see any feasible alternatives. One way would be to read the whole body before passing it to a client but this was the cause of OOM issues and #263 in the past. Turning on the flag would bring back #263. For that reason, I believe, it should default to `false`.

- [x] Yes
- [ ] No

MiSikora:
## :page_facing_up: Context

There is an issue - #432.

## :pencil: Changes

I added `alwaysReadResponseBody` flag to `ChuckerInterceptor` that lets users decide if they want to consume the whole body from responses. It is an opt-in feature because it can drastically change the behavior of an application between QA and production builds.

Sources are depleted with a new `DepeletingSource` class that is added conditionally based on the flag.

I also did some housekeeping on tests.
* The temporary directory is now a property of `ChuckerInterceptorTest` instead of an input argument of test functions.
* Okio's segment size is now a test constant.

## :no_entry_sign: Breaking

No.

There are unit tests to cover this functionality. To tests this manually you can use the sample app, turn on the flag, and see that the response body from Postman call is fully available.

## :stopwatch: Next steps

Closes #432.
I'm not sure about #425. It should not happen in the first place given the info from the author. This PR helps if the response body is longer than 8 KiB and parsing error happens before reading the last segment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,12 @@ val chuckerInterceptor = ChuckerInterceptor(
         // The max body content length in bytes, after this responses will be truncated.
         maxContentLength = 250000L,
         // List of headers to replace with ** in the Chucker UI
-        headersToRedact = setOf("Auth-Token"))
+        headersToRedact = setOf("Auth-Token"),
+        // Read response bodies fully even in case of not consuming all of the bytes by the client.
+        // This is useful in case of parsing errors or when the response body
+        // is closed before being read like in Retrofit with Void and Unit types.
+        alwaysReadResponseBody = true
+)
 
 // Don't forget to plug the ChuckerInterceptor inside the OkHttpClient
 val client = OkHttpClient.Builder()
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ val chuckerInterceptor = ChuckerInterceptor(
         maxContentLength = 250000L,
         // List of headers to replace with ** in the Chucker UI
         headersToRedact = setOf("Auth-Token"),
-        // Read response bodies fully even in case of not consuming all of the bytes by the client.
+        // Read the whole response body even when the client does not consume the response completely.
         // This is useful in case of parsing errors or when the response body
         // is closed before being read like in Retrofit with Void and Unit types.
         alwaysReadResponseBody = true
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -95,7 +95,11 @@ val chuckerInterceptor = ChuckerInterceptor(
         // The max body content length in bytes, after this responses will be truncated.
         maxContentLength = 250000L,
         // List of headers to replace with ** in the Chucker UI
-        headersToRedact = setOf("Auth-Token"))
+        headersToRedact = setOf("Auth-Token")),
+        // Read response bodies fully even in case of not consuming all of the bytes by the client.
+        // This is useful in case of parsing errors or when the response body
+        // is closed before being read like in Retrofit with Void and Unit types.
```
MiSikora, line 101: I'm not sure if mentioning Retrofit here makes sense.

vbuberen: I guess we can leave it for now.

README.md
```diff
@@ -95,7 +95,12 @@ val chuckerInterceptor = ChuckerInterceptor(
         // The max body content length in bytes, after this responses will be truncated.
         maxContentLength = 250000L,
         // List of headers to replace with ** in the Chucker UI
-        headersToRedact = setOf("Auth-Token"))
+        headersToRedact = setOf("Auth-Token"),
+        // Read response bodies fully even in case of not consuming all of the bytes by the client.
```
vbuberen, line 99: I suggest to simplify a little bit. What about changing to something like:
```suggestion
        // Read the whole response body even when the client does not consume the response completely.
```
?

vbuberen (approved):
LGTM! 🚢

ColtonIdle:
@MiSikora thank you. This is awesome. Good to be able to have the "same behavior" as chuck and charles.

PR #450, opened 2020-09-15
Fix classes visibility

vbuberen:
## :camera: Screenshots
Screenshot 2020-09-15 at 10 43 41

## :page_facing_up: Context
Found out that in docs generated by Dokka we have more than `api` package available (see the attached screenshot). These packages inside `internal` package shouldn't be visible to users, as well as classes in it. And such classes shouldn't be in docs as well.

## :pencil: Changes
- Updated visibility modifier for 2 classes.

Run `dokkaHtml` command and see that generated doc has only `api` package.

## :stopwatch: Next steps
Documentation rework is on me, so I will improve documentation when I get back to these tasks soon.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/ChuckerFileProvider.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/ChuckerFileProvider.kt
@@ -6,4 +6,4 @@ import androidx.core.content.FileProvider
  * We need our own subclass so we don't conflict with other [FileProvider]s
  * See: https://github.com/ChuckerTeam/chucker/issues/409
  */
-class ChuckerFileProvider : FileProvider()
+internal class ChuckerFileProvider : FileProvider()
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/ProtocolResources.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/ProtocolResources.kt
@@ -4,7 +4,7 @@ import androidx.annotation.ColorRes
 import androidx.annotation.DrawableRes
 import com.chuckerteam.chucker.R
 
-sealed class ProtocolResources(@DrawableRes val icon: Int, @ColorRes val color: Int) {
+internal sealed class ProtocolResources(@DrawableRes val icon: Int, @ColorRes val color: Int) {
     class Http : ProtocolResources(R.drawable.chucker_ic_http, R.color.chucker_color_error)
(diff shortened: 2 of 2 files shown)


MiSikora (approved):

library/src/main/java/com/chuckerteam/chucker/internal/support/ChuckerFileProvider.kt
```diff
@@ -6,4 +6,4 @@ import androidx.core.content.FileProvider
  * We need our own subclass so we don't conflict with other [FileProvider]s
  * See: https://github.com/ChuckerTeam/chucker/issues/409
  */
-class ChuckerFileProvider : FileProvider()
+internal class ChuckerFileProvider : FileProvider()
```
cortinico, line 9: Can you verify that this is not crashing at runtime?
This class is referenced inside the AndroidManifest and I'm unsure if `internal` visibility is valid here.

vbuberen: I was worrying about it as well. However, I had no issues on API 19 and 30 when tested file sharing in our sample app.

MiSikora: It should behave the same. `internal` compiles down to `public` so in the byte code it is the same thing unless I'm missing something.

cortinico: > internal compiles down to public so in the byte code it is the same thing unless I'm missing something.

It should actually mangle the name of the class to be `ChuckerFileProvider$com_github_ChuckerTeam_Chucker_library` if I'm not mistake (at least that's [what happens for method names](https://ncorti.com/blog/name-mangling-in-kotlin))

MiSikora: From my quick check, it should be ok. 

```kotlin
internal class Foo {
    internal fun foo() = Unit
}

internal fun foo() = Unit
```

```sh
kotlinc Foo.kt
javap Foo.class
```

```java
Compiled from "Foo.kt"
public final class Foo {
  public final void foo$main();
  public Foo();
}
```

cortinico: TIL 👍  Then it's good to go for me

cortinico (approved):

PR #451, opened 2020-09-16
Enable transaction file sharing

### LINKED ISSUE #404: Feature Request: Share as file
In Chucker we can save a file, and we can share from the app. Sometimes I want to share as a file. It would be great if I could do that in one click.

MiSikora:
## :camera: Screenshots

Screenshot_20200916-123520

## :page_facing_up: Context

There is an issue - #404.

## :pencil: Changes

I added a way to share a single transaction as a file. I also refactored sharing flow. `ShareUtils` are removed in favour of `Sharable` interface that is responsible for creating shared content that can be written to a file or wherever.

Tests for sharing run now on Robolectric to use real resources instead of a mocked Context.

## :no_entry_sign: Breaking

No.

Run the app and check sharing capabilities. Old ones should stay the same. Single file sharing should be available from transaction details.

## :stopwatch: Next steps

Closes #404


library/src/main/java/com/chuckerteam/chucker/internal/support/Sharable.kt
```diff
@@ -0,0 +1,72 @@
...
+import android.content.Intent
+import android.widget.Toast
+import androidx.core.app.ShareCompat
+import androidx.core.content.FileProvider
+import com.chuckerteam.chucker.R
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
+internal interface Sharable {
+    fun toSharableContent(context: Context): String
```
MiSikora, line 15: I'm not very fond of passing `Context` here. It makes implementations that do not use Android (like cURL one) dependent on it and a little bit harder to test.

On the other hand passing `Context` as an argument to the implementation might lead to leaks (even though it wouldn't do it ATM).

vbuberen: I don't think that passing `Context` here is an issue now. We can always reiterate on this decision later.

vbuberen (approved):
Looks good to me.

Issue #466, opened 2020-10-02
3.3.0 is binary-incompatible with 3.2.0


autonomousapps:
## :writing_hand: Describe the bug
I have a project that uses Chucker, which also uses a library that uses Chucker. The library is still using 3.2.0, and when I updated my app to 3.3.0 and ran it, it crashed at runtime with this error:

```
java.lang.NoSuchMethodError: No direct method <init>(Landroid/content/Context;Lcom/chuckerteam/chucker/api/ChuckerCollector;JLjava/util/Set;ILkotlin/jvm/internal/DefaultConstructorMarker;)V in class Lcom/chuckerteam/chucker/api/ChuckerInterceptor; or its super classes (declaration of 'com.chuckerteam.chucker.api.ChuckerInterceptor' appears in /data/app/~~GJiF7AJl427rbyfaLOyo1A==/com.myproject-GIw2-OO7zIRGQFD6KsWDQg==/base.apk)
```

According to @cortinico, 
> Seems like the ChuckerInterceptor(Context, ChuckerCollector, Set) is missing.

## :bomb: Steps to reproduce

1. Have a library with Chucker 3.2.0.
2. Have an app with Chucker 3.3.0, and which uses the above library.
3. Launch app.
4. Observe crash.

## :wrench: Expected behavior
App should not crash.

## :page_facing_up: Additional context

Adding the missing constructor will make it easier to migrate to Chucker 3.3.0.

autonomousapps:
Btw, I haven't formalized it yet, but it would be possible to use [my plugin](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin) to get a snapshot of your ABI at any given moment, and you could use that to compare across versions. You'd run `./gradlew :lib:abiAnalysisDebug` and then look at the output at `build/reports/dependency-analysis/debug/intermediates/abi-dump.txt`. I plan to make this easier in the future. If you do try this, let me know how it goes, as having a real use-case to discuss would make it easier to build out the feature fully.

MiSikora:
I think some information is missing. These are constructors from `3.2.0` and what they compiles down to.

https://github.com/ChuckerTeam/chucker/blob/a6ed2d4296d3d1ff700a0965f012f6f4d14dfb92/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt#L39-L65

```java
public final class com.chuckerteam.chucker.api.ChuckerInterceptor implements okhttp3.Interceptor {
[...]

PR #483, opened 2020-10-23
Remove deprecated constructor and use exposed builder pattern

### LINKED ISSUE #462: Create ChuckerInterceptor from builder

Having multi-arg constructor is nice from Kotlin perspective but not that great from Java and binary compatibility. I'd like to introduce in 4.x builder pattern.

It would look more or less like this. I omitted most input parameters for brevity.

```kotlin
class ChuckerInterceptor internal constructor(builder: Builder) : Interceptor {
  private val context = builder.context
  private val collector = builder.collector
  private val maxContentLength = builder.maxContentLength

  public constructor(context: Context) : this(Builder(context))

  class Builder(internal val context: Context) {
    internal var collector = ChuckerCollector(context)
    internal var maxContentLength = 250_000L

    fun collector(collector: ChuckerCollector) = apply { this.collector = collector }
    
    fun maxContentLength(maxContentLength: Long) = apply { this.maxContentLength = maxContentLength }

    fun build() = ChuckerInterceptor(this) 
  }
}
```

This approach is nice because users that don't need customization can create an interceptor with `ChuckerInterceptor(context)` and others can customize it nicely from both Kotlin and Java worlds.

I know there was a proposal for the DSL approach in #141. It would still be viable. We could hide the builder from Kotlin and hide DSL from Java. However, I don't see much value in DSL here and I feel like adding DSL would only be "fancy" and not useful.

We could make migration easier by deprecating the current constructor in `3.3.1` and by adding a builder to which we encourage migration.

One downside of this is that we would encourage users to migrate away from the constructor even if they use one that accepts only `Context`, which would be available in `4.x`. I don't think there is any way around it, unfortunately. This could be explained more in the deprecation message.

- [x] Yes
- [ ] No

MiSikora:
## :page_facing_up: Context

There is an issue - #462.

## :pencil: Changes

This implementation uses more robust builder pattern in terms of future development. It should be used with `4.x`. There is one inconvenience when it comes to default values but it is not that big of a deal. I'll describe it more in the comments to the code.

## :no_entry_sign: Breaking

Yes it breaks both API and ABI, but it is for `4.x`.

~~I suggest reviewing commits separately. With 5e7d733 you can check if `ReplaceWith` works correctly.~~

Nothing special to test.

## :stopwatch: Next steps

Closes #462

Helps with #466. It does not solve it, but I don't think we can fix it unfortunately. We can only make sure that in the future we don't fall into the same trap. Though suggestion with adding ABI compatibility tool would be a nice addition.

We should consider doing the same with `ChuckerCollector` and `RetentionManger`. They do not change that dynamically, so there might be no reason to switch there to the builder pattern but on the other hand it might pose at some point in the future same issues as #466.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -1,6 +1,7 @@
 package com.chuckerteam.chucker.api
 
 import android.content.Context
+import androidx.annotation.VisibleForTesting
 import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
 import com.chuckerteam.chucker.internal.support.CacheDirectoryProvider
 import com.chuckerteam.chucker.internal.support.DepletingSource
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -28,60 +28,29 @@ import kotlin.jvm.Throws
...
-    )
-
-    private val io: IOUtils = IOUtils(context)
-    private val headersToRedact: MutableSet<String> = headersToRedact.toMutableSet()
+    public constructor(context: Context) : this(Builder(context))
+
+    private val context = builder.context
+    private val collector = builder.collector ?: ChuckerCollector(context)
+    private val maxContentLength = builder.maxContentLength
+    private val cacheDirectoryProvider = builder.cacheDirectoryProvider ?: CacheDirectoryProvider { context.filesDir }
```
MiSikora, line 50: This is the inconvenience that I mentioned. Default fields are declared here instead of builder. It is only due to test purposes. Instantiating `ChuckerCollector` would require a lot of mocking of `Context` that I don't want to be botheredwith.

On the other hand it is nice to not create stuff eagerly in the builder.

vbuberen:
I like that we will have builder patter in Chucker soon.
However, I have a few concerns/suggestions about this PR.

1. I don't like how we are removing existing constructor without proper deprecation process. It is not necessary that we need `ReplaceWith` point to something specific, but at least deprecation messages, like we did for throwables would be good. 

2. With introduced changes and added deprecation messages (if we add some) I believe that we should change the version to `3.4.0`, not `3.3.1` to follow semantic versioning approach.

3. While looking through code I thought that `ChuckerInterceptor` looks quite filled with different stuff now. What if we just create some interface for the public API, which will be the single entry point for user? In such case we could create some separate Builder class to implement this interface and to build and return an actual configured `ChuckerInterceptor` object. Might be an overkill and lead us to more breaking changes, but we might end up with better separation of concerns.

MiSikora:
> 1. I don't like how we are removing existing constructor without proper deprecation process. It is not necessary that we need `ReplaceWith` point to something specific, but at least deprecation messages, like we did for throwables would be good.
> 2. With introduced changes and added deprecation messages (if we add some) I believe that we should change the version to `3.4.0`, not `3.3.1` to follow semantic versioning approach.

I'm not sure if I understand what you mean by `proper deprecation process`. If it is connected to your second point I get it. Otherwise can you explain how should it look like? Constructor is marked as `@Deprecated` and has a message.

> 3. While looking through code I thought that `ChuckerInterceptor` looks quite filled with different stuff now. What if we just create some interface for the public API, which will be the single entry point for user? In such case we could create some separate Builder class to implement this interface and to build and return an actual configured `ChuckerInterceptor` object. Might be an overkill and lead us to more breaking changes, but we might end up with better separation of concerns.

Can you create some snippet? I think I don't get the full picture.

vbuberen:
> I'm not sure if I understand what you mean by `proper deprecation process`.

I mean that we are just removing some existing part of our public API without any warnings. Yes, I agree that we don't make people do a ton of changes, but still, it would be better if we had some announcement/warning and some grace period when we have both deprecated constructor and the brand new builder. It feels kind of tricky since we have just one `ChuckerInterceptor` class which is being edited in this PR, but still.



> Can you create some snippet? I think I don't get the full picture.

Yes, I should do some, probably. Also need to connect all dots to prove that the concept is valid.

MiSikora:
> I mean that we are just removing some existing part of our public API without any warnings. Yes, I agree that we don't make people do a ton of changes, but still, it would be better if we had some announcement/warning and some grace period when we have both deprecated constructor and the brand new builder.

Maybe it got lost in the whole PR description. After merging this PR without squashing I want to cherry-pick 5e7d733 onto `3.x` where old constructor is still present and just deprecated. Users will have a grace period for as long as `4.x` is not released.

vbuberen:
Yes, looks like I missed the point about deprecations in one of commits.

In such case I am fine with the approach. But still, we would need to use `3.4.0` instead of `3.3.1` as we are introducing some new deprecations and some new API.

MiSikora:
Yes, it's a little bit unfortunate that we'd have to make next release with a minor bump but I want to remove the burdon of `@JvmOverloads` constructor from `4.x`.

vbuberen (approved):
Reviewed this PR by commits (as it supposed to happen) and everything looks fine and reasonable for me.

As to my suggestion about additional interfaces - we can come back to it later when we do other changes to transaction processing.

cortinico:
Great job 👍 Also sorry for being late to the game here.

> Though suggestion with adding ABI compatibility tool would be a nice addition.

+1 on this. I actually wanted to take a stab to add https://github.com/Kotlin/binary-compatibility-validator to Chucker in the near future (that will prevent future problems like this one).

> 2\. With introduced changes and added deprecation messages (if we add some) I believe that we should change the version to `3.4.0`, not `3.3.1` to follow semantic versioning approach.

Agree here on what @vbuberen mentioned. If we merge this, we need to release `3.4.0` as a deprecation (of our prominent API) is worth a minor bump rather than a patch-fix bump.

Imho this PR should also be split in 2 separate PRs: 5e7d733 should be pointed against `master` while 6d4a8b4 should be pointed against `4.x` (or however we wish to call it). As a rule of thumb is also useful for our users to have separate PRs that are easier to browse.

MiSikora:
> Imho this PR should also be split in 2 separate PRs: [5e7d733](https://github.com/ChuckerTeam/chucker/commit/5e7d733ad73b698e8f2e9cd75bed5480cb6e4cca) should be pointed against `master` while [6d4a8b4](https://github.com/ChuckerTeam/chucker/commit/6d4a8b47429ce571a2d65970cf796775b08058b6) should be pointed against `4.x` (or however we wish to call it). As a rule of thumb is also useful for our users to have separate PRs that are easier to browse.

I thought that we treat now our `develop` as a `4.x`. That's why I wanted to merge these two commit without squashing them, and then create a separate PR for `3.x` with some of commits that should also land in `3.x` `cherry-picked` from `develop`.

cortinico:
> I thought that we treat now our `develop` as a `4.x`.

_(I wrote master but I meant develop sorry)_

Yup I'm fine either way. But I would still advice we follow this flow:
- Create a PR with 5e7d733 & Merge it on develop
- Release version to 3.4.0 (<- this will be the point where we start the `3.x` branch).
- Create a new PR with 6d4a8b4 & Merge it on develop

MiSikora:
Ok, got it. I was just a little bit confused because we already have `3.x` but I we didn't anticipate next minor bump so it's all fine then. I'll split this into two separate PRs.

cortinico:
> Ok, got it. I was just a little bit confused because we already have `3.x`

Sorry, I also haven't noticed that. In that case, let's create a PR with 5e7d733 & Merge it on `3.x`

MiSikora:
Ok, I added `@VisibleForTesting`. 6d4a8b4 changed to a923a8f.

MiSikora:
Ok, I changed PR target. I think after merging #491 it will automatically retarget to `develop` branch. If not I'll update it then.

MiSikora:
Some conflicts popped up. I'll resolve them after `3.4.0` is released.

cortinico:
Let's rebase this on top of #496 once it's merged

cortinico (approved):
LGTM 👍

Issue #421, opened 2020-08-12
StrictMode policy violation: android.os.strictmode.LeakedClosableViolation


Pjnous:
Hello,

We have added the strict mode policy to our project:

 if (BuildConfig.DEBUG) {
            StrictMode.setVmPolicy(
                StrictMode.VmPolicy.Builder()
                    .detectLeakedSqlLiteObjects()
                    .detectLeakedClosableObjects()
                    .penaltyLog()
                    .build()
            )
        }

Since then I have noticed in Logcat

android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.

The full stacktrace is:

 D/StrictMode: StrictMode policy violation: android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
        at android.os.StrictMode$AndroidCloseGuardReporter.report(StrictMode.java:1877)
        at dalvik.system.CloseGuard.warnIfOpen(CloseGuard.java:286)
        at java.io.FileInputStream.finalize(FileInputStream.java:493)
        at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:289)
        at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:276)
        at java.lang.Daemons$Daemon.run(Daemons.java:137)
        at java.lang.Thread.run(Thread.java:919)
     Caused by: java.lang.Throwable: Explicit termination method 'close' not called
        at dalvik.system.CloseGuard.open(CloseGuard.java:237)
        at java.io.FileInputStream.<init>(FileInputStream.java:174)
        at okio.Okio__JvmOkioKt.source(JvmOkio.kt:178)
        at okio.Okio.source(Unknown Source:1)
        at com.chuckerteam.chucker.api.ChuckerInterceptor$ChuckerTransactionTeeCallback.readResponseBuffer(ChuckerInterceptor.kt:251)
        at com.chuckerteam.chucker.api.ChuckerInterceptor$ChuckerTransactionTeeCallback.onSuccess(ChuckerInterceptor.kt:239)
        at com.chuckerteam.chucker.internal.support.TeeSource.close(TeeSource.kt:69)
        at okio.RealBufferedSource.close(RealBufferedSource.kt:498)
        at okio.RealBufferedSource.close(RealBufferedSource.kt:498)
[...]

PR #488, opened 2020-10-31
Fix for not closed GzipSource

### LINKED ISSUE #472: StrictModeViolation: GzipSource is not closed
```
Caused by: java.lang.Throwable: Explicit termination method 'end' not called
   at dalvik.system.CloseGuard.open(CloseGuard.java:237)
   at java.util.zip.Inflater.<init>(Inflater.java:122)
   at okio.GzipSource.<init>(GzipSource.kt:44)
   at com.chuckerteam.chucker.internal.support.IOUtils.getNativeSource(IOUtils.kt:59)
   at com.chuckerteam.chucker.api.ChuckerInterceptor.processRequest(ChuckerInterceptor.kt:132)
   at com.chuckerteam.chucker.api.ChuckerInterceptor.intercept(ChuckerInterceptor.kt:97)
```

## :bomb: Steps to reproduce
```
StrictMode.setVmPolicy(
  VmPolicy.Builder()
    .detectLeakedClosableObjects()
    .penaltyLog()
    .penaltyDeath()
)
```

Call an endpoint using gzip compression.

## :wrench: Expected behavior
No violation. `GzipSource` is closed.

## :iphone: Tech info
 - Chucker version: 3.3.0

vbuberen:
## :page_facing_up: Context
There is an issue #472 
Issue appears with StrictMode enabled. We couldn't catch the issue earlier, because sample app has no gzip request with body, so code with error wasn't reachable.

## :pencil: Changes
- Fixed the issue by adding `use { }` function to GzipSource in `getNativeSource()` function.
- Updated tests to use real `Buffer()`, since Mockk can't properly react to `close()` call.
- Added a fake Gzip request which I fill with some dummy request body and header telling that response is with gzip encoding. It returns 405, but allows to reproduce the issue.
- Added StrictMode to the sample app, because this is not the first time we have issues reported by people with this mode enabled (for example #421)
- (UPD: after review) Renamed existing gzip request to explicitly mention that it returns just gzip response.

To reproduce the issue remove `use` from `getNativeSource()` and use just GzipSource directly. With StrictMode enabled sample will crash when `fakeGzipRequest()` is called.

## :stopwatch: Next steps
We might want to improve the part of request processing and cover it with some more tests.
Closes #472

### DIFF
--- a/sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
+++ b/sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
@@ -82,7 +82,8 @@ class HttpBinClient(
             stream(500).enqueue(cb)
             streamBytes(2048).enqueue(cb)
             image("image/png").enqueue(cb)
-            gzip().enqueue(cb)
+            gzipResponse().enqueue(cb)
+            gzipRequest(Data("Some gzip request")).enqueue(cb)
             xml().enqueue(cb)
             utf8().enqueue(cb)
             deflate().enqueue(cb)
@@ -99,7 +100,6 @@ class HttpBinClient(
         downloadSampleImage(colorHex = "fff")
         downloadSampleImage(colorHex = "000")
         getResponsePartially()
-        fakeGzipRequest()
     }
 
     internal fun initializeCrashHandler() {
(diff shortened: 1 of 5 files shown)


sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
```diff
@@ -127,6 +128,24 @@ class HttpBinClient(
         )
     }
 
+    private fun fakeGzipRequest() {
```
MiSikora, line 131: We can avoid custom function with a header on a Retrofit function. We could also use `/post` instead of `/gizp` to get back the result instead of `405`.

```kotlin
@POST("/post")
@Headers("Content-Encoding: gzip")
fun postGzipped(@Body body: Data): Call<Any?>
```

MiSikora (commented):
Left one minor thing.

As far as the requests revamp go I think we should go with a similar route to responses processing.

sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
vbuberen: Yeah, this one looks nicer. Added `/gzip` just to have a visual clue on what request it is.

MiSikora (approved):

PR #491, opened 2020-11-03
Use builder pattern for ChuckerInterceptor in a backwards compatible way

MiSikora:
## :page_facing_up: Context

It is a follow up to the discussion in #483. It also addresses #462 and #466.

## :pencil: Changes

I added a builder pattern in a `3.x` compatible way. This way users will be informed of changes and have more time to adjust their code. Unfortunately, `ReplaceWith` is not ideal. `imports` do no work properly with default arguments but IMHO it is a minor inconvenience (basically, only `ChuckerCollector` needs to be manually imported). Also, it will create builder with all fields. I've tried to split constructor into different ones with different `ReplaceWith` strategies but it is not possible due to named arguments and source compatibility in Kotlin.

## :paperclip: Related PR

#462

## :no_entry_sign: Breaking

No.

You can check if `ReplaceWith` works correctly.

## :stopwatch: Next steps

Prepare `3.4.0` release and `3.x` branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Please add your entries according to this format.
 
 ## Unreleased
 
+### Added
+
+* `ChuckerInterceptor.Builder` for fluent creation of the interceptor. It will also help us with preserving binary compatibility in future releases of `4.x`. [#462]
+
 ### Changed
 
 * Bumped `targetSDK` and `compileSDK` to 30 (Android 11).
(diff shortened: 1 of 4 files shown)


MiSikora:
@cortinico 

I'm not sure if that is what you wanted. Should I create a separate branch with just 2ead7cb `cherry-picked` or bringing all other commits is fine?

vbuberen (approved):
You are right, we still need to merge other commits into `3.x` as well as a part of `3.4.0` update. As to the strategy I believe that you should use `Rebase and merge` here to save the history. 

With different PRs we will make branches `develop` and `3.x` different anyway and it will add complexity with next merges into `3.x` in future. I assume we will have to do 2 PRs (into `develop` and into `3.x`) in future or play around with cherry-picking if something needs to land in both branches.

MiSikora:
Changed branch target in order to make `3.x` easier as we discussed on Slack.

vbuberen (approved):
Tested - `ReplaceWith` works fine.

PR #500, opened 2020-11-11
Update OkHttp and bump min SDK to 21

MiSikora:
## :page_facing_up: Context

This is a consequence of updating OkHttp that is planned for Chucker `4.x`.

## :pencil: Changes

- Bump OkHttp to 4.9.0 and replace deprecated methods.
- Bump Retrofit to 2.9.0.
- Fix Lint issues due to min SDK bump (and some other minor ones).
- Remove OkHttp and Retrofit from Dependabot ignore list.
- Update README about Java 8 support.
- Enable Java 8 in the sample.

## :no_entry_sign: Breaking

Yes, min SDK is higher and Java 8 support is needed, but this is expected.

Not sure. Run the app and check it? Otherwise let the CI do its job.

## :stopwatch: Next steps

Maybe enable Lint checks on GH Actions?


README.md
```diff
@@ -60,7 +74,7 @@ You can find older version of Chucker here: [![JitPack](https://jitpack.io/v/Chu
 Don't forget to check the [changelog](CHANGELOG.md) to have a look at all the changes in the latest version of Chucker.
 
 * Compatible with **OkHTTP 4**
-* **API >= 16** compatible
+* **API >= 21** compatible
```
cortinico, line 77: Should we add a note that the last API 16+ version is Chucker `3.4.0`?

cortinico (approved):
Went through the whole diff and it looks good on my end 👍

README.md
MiSikora: IMHO readme and all of the stuff should be about the current state of the repository at a specific commit. If someone wants to know about `3.4.0` or whichever version they should check the specific commit or a tag.

cortinico: > IMHO readme and all of the stuff should be about the current state of the repository at a specific commit

As a general rule, I agree.

However the Min SDK bump can be a bummer for several users. Offering a previous version as an alternative could be a nice way to don't entice them away (similarly to what happened with the AndroidX migration).

MiSikora: Ok, makes sense. Would you see it as a separate section in readme?

We should probably add to this info a period for how long we plan to support `3.x` actively. If I remember correctly it was supposed to be half a year? @vbuberen do you remember? I think you should know better.

cortinico: Yeah even a separate section would work 👌

vbuberen: We mentioned bumping minSDK in 3.3.0 release https://github.com/ChuckerTeam/chucker/blob/develop/CHANGELOG.md#version-330-2020-09-30

We can simply mention that Chucker 3.x has minSDK 16 and Chucker 4.x has minSDK 21. It will save us from doing some changes when we need to release something like 3.4.1, for example.

vbuberen (approved):
Looks good to me.

Had to check changelogs for OkHttp and Retrofit quite thoroughly, because such version jump is quite scary.

As to next steps - we could consider switching to OkHttp BOM.

library/build.gradle
```diff
@@ -28,6 +28,8 @@ android {
     lintOptions {
         warningsAsErrors true
         abortOnError true
+        // We don't want to impose RTL on consuming applications.
+        disable 'RtlEnabled'
```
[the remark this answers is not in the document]
MiSikora: Since we use attributes like `end` or `start` Lint complains about RTL not being enabled in the manifest. But this can be done only on the application level. This is related to this PR in a sense that we don't use `left` or `right` anymore.

MiSikora:
Ok, I updated PR with code review fixes and added info about Android 16+ to FAQ.

PR #505, opened 2020-11-21
Update README.md

okanaydin:
fix the typo


vbuberen:
Good catch!

Thanks for your contribution.

PR #508, opened 2020-11-24
Update Sample app layout

vbuberen:
## :camera: Screenshots
| Before | After |
| --- | --- |
|  |  |

| Before | After |
| --- | --- |
| |  |

## :page_facing_up: Context
Saw that we forgot to remove `Trigger an exception` button when were removing throwables funcitonality. So I removed this button. Since I was editing this layout anyway I decided to improve it to make it more convenient and to look better on different screens. 
I believe that updated layout should be more convenient for users of big screens. Additionally, since max width for buttons and text is limited now, in landscape mode this screen looks slightly better.

## :pencil: Changes
- Removed `Trigger an exception` button.
- Updated elements position and properties.

## :stopwatch: Next steps
Once this PR is merged I will record new gifs.

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -55,14 +55,17 @@ dependencies {
     debugImplementation project(':library')
     releaseImplementation project(':library-no-op')
 
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
+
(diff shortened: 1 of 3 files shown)


cortinico (approved):
LGTM 👍

MiSikora (approved):

PR #511, opened 2020-11-28
Fix broken lint and add it to GH Actions

cortinico:
## :page_facing_up: Context
Lint is currently broken on `develop`. I'm fixing it and I've added a job on GH Actions to run it on pre-merge. If you happen to run `gw check` to run all the verifications, it will fail.

## :pencil: Changes
- Disabled `AcceptsUserCertificates` as we intentionally want to allow user CAs
- Fixed error on `android:networkSecurityConfig`

## :stopwatch: Next steps
I believe we should update the pre-commit hook to run `gw check`

### DIFF
--- a/sample/src/main/res/values/dimens.xml
+++ b/sample/src/main/res/values/dimens.xml
@@ -1,6 +1,6 @@
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
     <!-- Default screen margins, per the Android Design guidelines. -->
-    <dimen name="norm_grid_size">8dp</dimen>
+    <dimen name="norm_grid_size" tools:ignore="UnusedResources">8dp</dimen>
     <dimen name="doub_grid_size">16dp</dimen>
     <dimen name="max_width">500dp</dimen>
 </resources>
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -55,6 +55,27 @@ jobs:
     - name: Stop Gradle
       run: ./gradlew --stop
 
+  lint:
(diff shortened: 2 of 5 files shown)


vbuberen (commented):
I am not sure I understand what is `gw check`?

cortinico:
> I am not sure I understand what is `gw check`?

`gw` is a shorthand for `./gradlew` (a bash alias).
`check` is a [lifecycle task](https://docs.gradle.org/current/userguide/base_plugin.html#sec:base_tasks) to run all the verifications of a project (such as Detekt, KtLint, tests, etc.)

vbuberen:
Understood. Wanted to be sure.

vbuberen (approved):
LGTM 🚢

sample/src/main/res/values/dimens.xml
```diff
@@ -1,6 +1,6 @@
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
     <!-- Default screen margins, per the Android Design guidelines. -->
-    <dimen name="norm_grid_size">8dp</dimen>
+    <dimen name="norm_grid_size" tools:ignore="UnusedResources">8dp</dimen>
```
koral--, line 3: Why it is ignored?
Shouldn't be just removed if it is really unused? 
Or maybe it is preserved for backwards compatibility?

cortinico: > Why it is ignored?

It's ignored as it's our base grid size. All the others size are derived from this. I'd rather keep it otherwise `doub_grid_size` would make little sense

MiSikora (approved):

Issue #202, opened 2020-01-24
Support brotli compression


jakoss:
Did you consider to support brotli compression in your library? OkHttp is supporting it using `BrotliInputStream` from https://github.com/google/brotli/blob/master/java/org/brotli/dec/BrotliInputStream.java

vbuberen:
Yeah, great idea. Going to include it into our roadmap. Thank you.

PR #519, opened 2020-12-09
Switch to OkHttp BOM in the library module

vbuberen:
## :page_facing_up: Context
Since we are going to add more OkHttp packages, for example, to implement #202 I suggest us to switch to BOM for OkHttp.
I think that we are fine with using BOM only in the `library` module, since in `library-no-op` and in `sample` we use only one package from OkHttp set.

## :pencil: Changes
- Switched to BOM for OkHttp dependencies in `build.gradle` file

## :no_entry_sign: Breaking
Nothing breaking expected


cortinico (approved):
Great 👍  LGTM

Issue #286, opened 2020-03-25
Add filters for transactions list


vbuberen:
## :warning: Is your feature request related to a problem? Please describe
Currently in Chucker it is hard to find only GET items or any other items by some property (time, size, status, etc.). So it might be a good case to add filtering, especially as there are partially the same feature requests appear (for example #273).

## :bulb: Describe the solution you'd like
On the main Chucker screen with list of transactions add a button which will call a `BottomSheet` with the following filters:
- `Checkboxes` for GET, POST, PUT, etc., so users could select only one or multiple types they need.
- `Checkboxes` for status codes (200, 500, etc.)
- `EditTexts` with min and max request duration took.

List of filters might expand with additional users feedback on which filters they would like to see.

This `BottomSheet` should also be with buttons to apply and reset filters. `Apply` will apply filters and close the `BottomSheet`, while `Reset` should only clear filters and leave `BottomSheet` opened, since I see use cases for reset mainly as a sequence of clearing and selecting new filters right away.

As to search currently present on that main screen - it should do search among filtered items (e.g. on any currently visible list of transactions).

## :page_facing_up: Additional context
Such feature should be implemented with minimal amount of changes to entities or other logic to not make library core more complex. 
Also, I would like to switch to `ListAdapter` with `DiffUtil` first to provide nice and efficient lists update.

## :raising_hand: Do you want to develop this feature yourself?

- [x] Yes
- [ ] No

cortinico:
That sounds like a great addition!

I think it's probably better if we find an agreement here about the list of filters we want to implement. The one you suggested sound already great.

NULLPointerGuy:
Few what I could think of is 
1.Filter based on Http status code.
2.Filter based on Http or Https
3.Filter based on the Http request method.
4.Filter based on the duration of API call
5.Filter based on the size of the response content.

alorma:
I would say to add a filter also per domain and content type.

We want to get rid of images on chucker, and add ability to filter by `media.xxx.yyy` could be a good addition

Sloy:
I was just talking with @alorma about proposing a similar feature, and he linked me to this issue 😂 

We would like to have some filtering capabilities too. In our case, filtering domains and Content-Type would be pretty handy.

cortinico:
@ArjanSM I think this can be implemented, but I would suggest we maybe add filtering using some custom queries? like `method:GET` would filter only for transactions that GET methods or so, what do you think?

[...]

PR #520, opened 2020-12-13
Use ListAdapter for TransactionsAdapter

vbuberen:
## :page_facing_up: Context
Something that I mentioned in different chats or issues (for example [here](https://github.com/ChuckerTeam/chucker/issues/286)), but never delivered - updating transactions list nicely with `DiffUtil` via `ListAdapter`

## :pencil: Changes
- Switched from regular `RecyclerView.Adapter` to `ListAdapter`

## :no_entry_sign: Breaking
Nothing breaking expected

Run the sample app and check how transactions list updates when some network activity happens and during the search

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TransactionDiffCallback.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TransactionDiffCallback.kt
@@ -11,4 +11,7 @@ internal object TransactionDiffCallback : DiffUtil.ItemCallback<HttpTransactionT
     override fun areContentsTheSame(oldItem: HttpTransactionTuple, newItem: HttpTransactionTuple): Boolean {
         return oldItem == newItem
     }
+
+    // Overriding function is empty on purpose to avoid flickering by default animator
+    override fun getChangePayload(oldItem: HttpTransactionTuple, newItem: HttpTransactionTuple) {}
 }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/TransactionDiffCallback.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/TransactionDiffCallback.kt
@@ -13,5 +13,5 @@ internal object TransactionDiffCallback : DiffUtil.ItemCallback<HttpTransactionT
     }
 
     // Overriding function is empty on purpose to avoid flickering by default animator
-    override fun getChangePayload(oldItem: HttpTransactionTuple, newItem: HttpTransactionTuple) {}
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/support/TransactionDiffCallback.kt
```diff
@@ -0,0 +1,14 @@
+package com.chuckerteam.chucker.internal.support
+
+import androidx.recyclerview.widget.DiffUtil
+import com.chuckerteam.chucker.internal.data.entity.HttpTransactionTuple
+
+internal object TransactionDiffCallback : DiffUtil.ItemCallback<HttpTransactionTuple>() {
```
MiSikora, line 6: We might consider overriding `getChangePayload` and return i.e. `Unit` from it. This will prevent item animation change and remove flicker.

vbuberen: You mean that we could use onBind with payload in adapter to avoid blinking the whole item, but update only changed fields?

MiSikora: Yes. :)

vbuberen: Got it. Interesting trick. Checked it - with light theme the flicker is less visible for me, but with dark theme your suggestion looks better 👍🏻

MiSikora (approved):

library/src/main/java/com/chuckerteam/chucker/internal/support/TransactionDiffCallback.kt
```diff
@@ -0,0 +1,17 @@
...
+    override fun areItemsTheSame(oldItem: HttpTransactionTuple, newItem: HttpTransactionTuple): Boolean {
+        return oldItem.id == newItem.id
+    }
+
+    override fun areContentsTheSame(oldItem: HttpTransactionTuple, newItem: HttpTransactionTuple): Boolean {
+        return oldItem == newItem
+    }
+
+    // Overriding function is empty on purpose to avoid flickering by default animator
+    override fun getChangePayload(oldItem: HttpTransactionTuple, newItem: HttpTransactionTuple) {}
```
MiSikora, line 16:
```suggestion
    override fun getChangePayload(oldItem: HttpTransactionTuple, newItem: HttpTransactionTuple) = Unit
```

PR #526, opened 2020-12-21
Centralise all logging in Chucker

MiSikora:
## :page_facing_up: Context

Some parts of the code used `printStackTrace()` to log failures. I wanted to centralise logging point that can be used in Chucker. This also gives an opportunity to let users set an external logger in the future.

## :pencil: Changes

I changed `Logger` to an interface and made an instance of it available in `Chucker` object. Companion object of `Logger` delegates then to this property.

I also added a test extension that sets no-op logger. It is needed as we can't have a centralised logger that is not static due to logging in Android components as well.

One more thing – I removed "Chucker" from log messages. It is present in the tag anyway.

## :paperclip: Related PR

#452

## :no_entry_sign: Breaking

No.

Nothing special to check. Rely on unit tests.

## :stopwatch: Next steps

N/A

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -7,6 +7,7 @@ import com.chuckerteam.chucker.internal.support.CacheDirectoryProvider
 import com.chuckerteam.chucker.internal.support.DepletingSource
 import com.chuckerteam.chucker.internal.support.FileFactory
 import com.chuckerteam.chucker.internal.support.IOUtils
+import com.chuckerteam.chucker.internal.support.Logger
 import com.chuckerteam.chucker.internal.support.ReportingSink
 import com.chuckerteam.chucker.internal.support.TeeSource
 import com.chuckerteam.chucker.internal.support.contentType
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -183,7 +184,7 @@ public class ChuckerInterceptor private constructor(
     private fun createTempTransactionFile(): File? {
         val cache = cacheDirectoryProvider.provide()
         return if (cache == null) {
-            IOException("Failed to obtain a valid cache directory for Chucker transaction file").printStackTrace()
+            Logger.warn("Failed to obtain a valid cache directory transaction file")
```
cortinico, line 187: Should we generate an `Exception` here and call `Logger.error` instead?

cortinico (approved):
Sorry for the late review but I was 🎅 -ing. 
LGTM other than a minor comment 👍

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
MiSikora: I removed `Exception` because I thought it was used only to avoid `println` as there was no access here to a logger. I'm fine either way and can bring it back if you think it helps.

cortinico: I'm unsure about the value added tbh. I'm fine either way 👍

MiSikora: Ok, I'll leave it them. Just will fix a typo that I just noticed.

vbuberen:
Wanted to take part in review as well 😕

MiSikora:
Sorry, I didn't think this to be an important change to require review from everyone. If you have any comments leave them and I'll make a subsequent PR. :)

vbuberen:
Ok, no problem. I thought that changes in classes with our public API would require more maintainers to check.

Issue #523, opened 2020-12-17
Improved handling of encoded/binary bodies


mattprecious:
## :bulb: Describe the solution you'd like
It would be great if we could optionally provide a custom decoder to handle bodies that aren't plain text. We use protocol buffers, so I would like to be able to print something useful inside Chucker instead of `(encoded or binary body omitted)`.

## :bar_chart: Describe alternatives you've considered
None

## :page_facing_up: Additional context
An interface with a function that takes the `Request`/`Response` and returns a string would probably be enough.

Something like this, but with better naming:
```kt
interface BodyDecoder { // Converter?
  fun decodeRequest(request: Request): String?
  fun decodeResponse(response: Response): String?
}
```
Or maybe you do converter factories like retrofit :shrug:

## :raising_hand: Do you want to develop this feature yourself?

- [ ] Yes
- [x] No

cortinico:
That's definitely and valuable extension point for Chucker 4.x

MiSikora:
I was thinking about it for some time as well. The thing I haven't figured out is how we could get correct type for a decoder.

@mattprecious How would getting a concrete decoder look like for this interface? Retrofit relies on concrete types and reflection. We, unfortunately, don't have this information.

mattprecious:
Yeah, decoding with proper names and type information wouldn't be straightforward. However, it is possible to decode a proto without that information. My ideas for the steps I would take to incorporate this into our app:

1) Print out the bytes in utf8 hex. This string can be pasted into a tool (like, self-plug: [protogram](https://mattprecious.github.io/protogram)) to decode into something somewhat useful. You don't get class/field/enum names, but it's still very useful information and you can manually map tag/value numbers back to useful names by opening the proto definition yourself if needed.
2) Incorporate protogram into our debug app to save you the step of copying and pasting.
3) If this was being heavily used and could justify the investment, we could use reflection or code-gen something based on our retrofit interface(s) and use the path to know which adapters to use. That's a lot of work, but it sounds plausible.

If we only receive the `ByteString` then we lose valuable information for decoding like the path, content type, response code, etc.

MiSikora:
Ok, thanks for the feedback. Will think about it.

[...]

PR #527, opened 2020-12-21
Refactor ChuckerInterceptor processing internals

MiSikora:
## :page_facing_up: Context

I started working on #523 but it was rather hard to give some public API for custom requests and responses processing.

## :pencil: Changes

There is a lot of code changes but most of it is just refactor. I deleted `IOUtils` class in favour of extensions on `Okio` and `OkHttp`. I also added two classes – `RequestProcessor` and `ResponseProcessor` to remove direct complexity in `ChuckerInterceptor`. These classes do too much ATM anyway but it is some start.

I fixed also some bugs.
- One shot and duplex requests are no longer processed by `ChuckerInterceptor` as it can't do it anyway.
- Plain text response body is no longer incorrectly set when processing requests.
- Requests headers are redacted when processing requests as well. Previously they were only redacted on response received, but this was a potential source of bugs due to exceptions thrown when getting the response.
- Known supported encodings no longer imply that request bodies are plain text. This was false as gzipped body does not necessarily mean plain text.

Some other changes.
- `HttpTransaction` has `isRequestBodyPlainText` and `isResponseBodyPlainText` now set to `false` by default. I feel it is better to rely on explicitly detecting plain text rather than assuming it. The implication is different message in export files for bodies without content.
- IO exceptions handling while processing requests.
- `chucker_body_unexpected_eof` message is removed as it is now impossible to read too much from in-memory buffers while processing the response.

## :no_entry_sign: Breaking

No.

I added some unit tests for request processing in `ChuckerInterceptor`. You can run the sample and check if everything is fine.

## :stopwatch: Next steps

I'll need to resolve some conflicts after merging #526. Or remove logging changes from this PR if #526 does not get accepted.


cortinico:
> I fixed also some bugs.
> 
> * One shot and duplex requests are no longer processed by `ChuckerInterceptor` as it can't do it anyway.
> * Plain text response body is no longer incorrectly set when processing requests.
> * Requests headers are redacted when processing requests as well. Previously they were only redacted on response received, but this was a potential source of bugs due to exceptions thrown when getting the response.
> * Known supported encodings no longer imply that request bodies are plain text. This was false as gzipped body does not necessarily mean plain text.

I have to admit that this has been sitting in my Inbox for a month 😅  (I was also pretty busy). Sorry for that.

Could we split this PR out in smaller ones given that is a ~1K LOC diff? Ideally have a refactoring one and others for every PR?

vbuberen:
Joining here (needed some break). Agree with Nico here about splitting the PR for more manageable size.

MiSikora:
I don't think it will shave off much from this PR because most of it is moving internals of `ChuckerInterceptor` to separate classes. But I'll try to split it as best as possible when I find some time.

MiSikora:
Closing this one. I started splitting it into smaller PRs.

PR #538, opened 2021-01-27
Fix setting request body plain text in transaction

MiSikora:
## :page_facing_up: Context

There is a bug while processing requests.

## :pencil: Changes

I set a correct property on a transaction. Previously it used response, now it uses request.

## :paperclip: Related PR

#527

## :no_entry_sign: Breaking

No.

I added a test that covers the bug.


cortinico (approved):
Nice catch 👌

vbuberen (approved):

PR #541, opened 2021-01-29
Auto cancel workflow when new push happens

vbuberen:
## :page_facing_up: Context
Github Actions has no feature to cancel previous workflow execution if some new push even happens triggering the same workflow again. In Chucker case it happens when somebody merges multiple PRs to `develop` one after another firing multiple runs of checks and snapshot publishing. It is not a big deal, but not efficient.
With suggested action: https://github.com/styfle/cancel-workflow-action we can improve efficiency by cancelling previous running builds.

## :pencil: Changes
- Modified `pre-merge` and `publish-snapshot` actions to automatically cancel workflows if new trigger event happened.
- Added a few empty lines to separate steps in workflows.

## :no_entry_sign: Breaking
It might be that pushing code to different PRs quite fast will cancel `pre-merge` running for some other build, but not sure if we have such cases.

- Open some PR and push some update to it while `pre-merge` is still running.
- Merge a few PRs into `develop` one after another and see that only latest merged gets to completion.

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -15,8 +15,14 @@ jobs:
     runs-on: [ubuntu-latest]
 
     steps:
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.7.0
+      with:
+        access_token: ${{ github.token }}
+
     - name: Checkout Repo
       uses: actions/checkout@v2
+
     - name: Cache Gradle Folders
       uses: actions/cache@v2
       with:
@@ -40,6 +46,7 @@ jobs:
     steps:
     - name: Checkout Repo
       uses: actions/checkout@v2
+
     - name: Cache Gradle Folders
       uses: actions/cache@v2
       with:
@@ -61,6 +68,7 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
+
       - name: Cache Gradle Folders
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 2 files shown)


cortinico:
> * Merge a few PRs into `develop` one after another and see that only latest merged gets to completion.

I think it's a good addition but I'm afraid of one scenario: if you merge multiple PRs, you actually want to know if one of those broke `develop`. Moreover, you can access all the snapshots that got created in between. If we stop the previous workflow, we will know only the resulting outcome.

I think this will be helpful for PRs instead. Perhaps we can limit this to run only if the originating event is a `pull_request`.

MiSikora:
I also wouldn't cancel quality checks. They are useful regardless of the previous stage. Unless we require updating the branch before merging.

> Moreover, you can access all the snapshots that got created in between.

Aren't snapshots volatile and only the latest one is available?

cortinico:
> Aren't snapshots volatile and only the latest one is available?

As from Maven specifications, yes.

On Sonatype, they're however stored using a timestamp classifier:

Screenshot 2021-01-29 at 20 14 35

Seems like Sonatype is keeping the last 10 `-SNAPSHOT` version and allows you to retrieve them if you specify the timestamp/revision number. They're all listed here: https://oss.sonatype.org/service/local/repositories/snapshots/content/com/github/chuckerteam/chucker/library/4.0.0-SNAPSHOT/

That being said, I don't know if we will ever need to grab a "older" snapshot so I'm fine either way.

vbuberen:
Ok, I will try to update my PR to use this step only in pull requests.

vbuberen:
Updated to trigger Cancel only when there were new changes to pull request pushed.

MiSikora (approved):

cortinico (approved):

PR #542, opened 2021-01-29
Request processing refactor

MiSikora:
## :page_facing_up: Context

`ChuckerInterceptor` was doing too much and it would be hard to integrate #523 in a sane manner.

## :pencil: Changes

I refactored code responsible for requests processing.
* Added `LimitingSource` that replaces method on `IOUtils` that read limited amount of content. Due to that `chucker_body_unexpected_eof` message is removed as it is now impossible to read too much from in-memory buffers while processing the response.
* Added `RequestProcessor` that moves away request logic from the interceptor.
* `HttpTransaction` has `isRequestBodyPlainText` is now set to false by default and changed to true only when plain text is detected.
* Added interceptor tests for request processing.
* Added logging for handling exceptions while processing requests.

## :paperclip: Related PR

#527 

## :no_entry_sign: Breaking

No.

Added some tests. Other than that, sample app should behave the same.

## :stopwatch: Next steps

Similar PR for response to follow. Also PRs fixing some other issues mentioned in #527.


vbuberen:
Sorry, I occasionally merged this PR instead of my PR with workflow update 🙈🤦🤦🤦🤦

vbuberen:
Restore the branch. Reverting pull request it probably too much.

PR #543, opened 2021-01-30
Request processing refactor

MiSikora:
## :page_facing_up: Context

Due to mishaps in #542 I'm putting this for review. If there are any comments I'll fix them in this PR and then create a new one with fixes only that target develop. Below is the copy of description from #542.

`ChuckerInterceptor` was doing too much and it would be hard to integrate #523 in a sane manner.

## :pencil: Changes

I refactored code responsible for requests processing.
* Added `LimitingSource` that replaces method on `IOUtils` that read limited amount of content. Due to that `chucker_body_unexpected_eof` message is removed as it is now impossible to read too much from in-memory buffers while processing the response.
* Added `RequestProcessor` that moves away request logic from the interceptor.
* `HttpTransaction` has `isRequestBodyPlainText` is now set to false by default and changed to true only when plain text is detected.
* Added interceptor tests for request processing.
* Added logging for handling exceptions while processing requests.

## :paperclip: Related PR

#527 

## :no_entry_sign: Breaking

No.

Added some tests. Other than that, sample app should behave the same.

## :stopwatch: Next steps

Similar PR for response to follow. Also PRs fixing some other issues mentioned in #527.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
@@ -36,8 +36,7 @@ internal class RequestProcessor(
     private fun processBody(request: Request, transaction: HttpTransaction) {
         val body = request.body ?: return
 
-        val isEncodingSupported = request.headers.hasSupportedContentEncoding
-        if (!isEncodingSupported) {
+        if (!request.headers.hasSupportedContentEncoding) {
             return
         }
 
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
@@ -14,11 +14,10 @@ internal class RequestProcessor(
     private val collector: ChuckerCollector,
     private val maxContentLength: Long,
 ) {
-    fun process(request: Request, transaction: HttpTransaction): Request {
+    fun process(request: Request, transaction: HttpTransaction) {
         processMetadata(request, transaction)
         processBody(request, transaction)
         collector.onRequestSent(transaction)
-        return request
     }
 
     private fun processMetadata(request: Request, transaction: HttpTransaction) {
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
@@ -40,12 +40,13 @@ internal class RequestProcessor(
             return
         }
 
-        val limitingSource = try {
+        val requestSource = try {
             Buffer().apply { body.writeTo(this) }
         } catch (e: IOException) {
             Logger.error("Failed to read request payload", e)
             return
-        }.uncompress(request.headers).let { LimitingSource(it, maxContentLength) }
+        }
+        val limitingSource = LimitingSource(requestSource.uncompress(request.headers), maxContentLength)
 
         val contentBuffer = Buffer().apply { limitingSource.use { writeAll(it) } }
         if (!contentBuffer.isProbablyPlainText) {
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -62,7 +62,9 @@ public class ChuckerInterceptor private constructor(
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
         val transaction = HttpTransaction()
-        val request = requestProcessor.process(chain.request(), transaction)
+        val request = chain.request()
+
+        requestProcessor.process(request, transaction)
 
         val response = try {
             chain.proceed(request)


library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
```diff
@@ -0,0 +1,66 @@
...
+    private val context: Context,
+    private val collector: ChuckerCollector,
+    private val maxContentLength: Long,
+) {
+    fun process(request: Request, transaction: HttpTransaction): Request {
+        processMetadata(request, transaction)
+        processBody(request, transaction)
+        collector.onRequestSent(transaction)
+        return request
+    }
```
cortinico, line 22:
```suggestion
    fun process(request: Request, transaction: HttpTransaction): Request =
        request.apply {
            ....
        }
```

nit: can be rewritten with

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    private fun processBody(request: Request, transaction: HttpTransaction) {
+        val body = request.body ?: return
+
+        val isEncodingSupported = request.headers.hasSupportedContentEncoding
+        if (!isEncodingSupported) {
+            return
+        }
```
cortinico, line 42:
```suggestion
        if (!request.headers.hasSupportedContentEncoding) {
            return
        }
```

cortinico: I think that's clear enough from the variable naming

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        val limitingSource = try {
+            Buffer().apply { body.writeTo(this) }
+        } catch (e: IOException) {
+            Logger.error("Failed to read request payload", e)
+            return
+        }.uncompress(request.headers).let { LimitingSource(it, maxContentLength) }
```
cortinico, line 49: nit: Here we mix multi line lambdas with inline lambdas that is not exactly elegant. Can you add a couple of new lines?

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        val contentBuffer = Buffer().apply { limitingSource.use { writeAll(it) } }
+        if (!contentBuffer.isProbablyPlainText) {
+            return
+        }
+
+        transaction.isRequestBodyPlainText = true
```
cortinico, line 56: _Commenting here but not exactly related to this line:_

Have you verified that this works fine with images (even with truncated ones)? I also believe that `isRequestBodyPlainText` is probably a confusing name within the image scenario. Maybe we can come up with a better name?

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
MiSikora: I don't mind changing it to `apply` but it looks strange to me. Typically I use `apply` to set properties on an applied object. Maybe `with` is better?

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
MiSikora: I want to make sure I understand you correctly because there is only one lambda. Do you mean `try/catch` expression that is mixed with a one liner lambda next to it or generally in the project?

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
MiSikora: Not sure I got that. We never processed images in requests so they won't pass through supported encoding check.

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
cortinico: > Maybe `with` is better?

Perhaps a `let` also works

MiSikora: Ah, with `with` and `let` I'd have to return explicitly from lambda. Seems `run` is the best. This got unnecessarily mind-boggling. :D

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
cortinico: 🤦 I thought we were looking at responses my bad. Closing here.

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
cortinico: > because there is only one lambda

My bad. I meant _block_ and it was related to this statement specifically. Also it's a nit so I'll leave it to your final decision :)

Maybe something like:

```
}.uncompress(request.headers).let { 
    LimitingSource(it, maxContentLength) 
}
```

or even
```

cortinico (approved):
LGTM 👍

vbuberen (approved):

PR #544, opened 2021-01-30
Do not process one shot and duplex request bodies

MiSikora:
## :page_facing_up: Context

Processing one shot request bodies is a bug as it means that after we process them, they cannot reach the server. Processing duplex request bodies is problematic as the bytes can be intertwined and out of order for us to process them.

One thing that is not that obvious is the message that users see in the Chucker UI as the body processing was omitted for a different reason.

Screenshot 2021-01-30 at 10 59 40

We could add new fields to the database model and show appropriate information based on that or we can change the single message that we have. I'm open to both propositions.

## :pencil: Changes

I added two conditions that check if a request body should be processed. If they are positive the event is logged and request processing is stopped. I also added a sample that utilises a one shot request body.

## :paperclip: Related PR

#527

## :no_entry_sign: Breaking

No.

Tests were added. Unfortunately only for one shot requests. Setting up an HTTP2 server for duplex test is too problematic and there wouldn't be much gain there compared to the effort. Alternatively tests for `RequestProcessor` could be written where we check if it processed a duplex request body.

Also you can run the sample and see that one shot requests are not being processed.

## :stopwatch: Next steps

This can wait with merging after #543 is resolved.

### DIFF
--- a/library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
+++ b/library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
@@ -516,7 +516,6 @@ internal class ChuckerInterceptorTest {
         client.newCall(oneShotRequest).execute().readByteStringBody()
 
         val transaction = chuckerInterceptor.expectTransaction()
-        assertThat(transaction.isRequestBodyPlainText).isFalse()
         assertThat(transaction.requestBody).isNull()
     }
 
--- a/library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
+++ b/library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
@@ -18,6 +18,7 @@ import okhttp3.RequestBody.Companion.toRequestBody
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
 import okio.Buffer
+import okio.BufferedSink
 import okio.ByteString
 import okio.ByteString.Companion.encodeUtf8
 import okio.GzipSink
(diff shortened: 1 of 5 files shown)


library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -496,5 +498,48 @@ internal class ChuckerInterceptorTest {
...
 
+    @ParameterizedTest
+    @EnumSource(value = ClientFactory::class)
+    fun oneShotRequestBody_isNotAvailableToChucker(factory: ClientFactory) {
+        server.enqueue(MockResponse())
+        val client = factory.create(chuckerInterceptor)
+
+        val oneShotRequest = object : RequestBody() {
+            private val content = Buffer().writeUtf8("Hello, world!")
+            override fun isOneShot() = true
```
cortinico, line 509: Could you add a test also for duplex request here?

MiSikora: I mentioned it in "how to test" section. It's not really feasible at HTTP level. The best I could do is to write tests for `ResponseProcessor`.

cortinico: My bad, I overlooked that 👍 Closing this comment.

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -496,5 +498,48 @@ internal class ChuckerInterceptorTest {
...
+            override fun contentType() = "text/plain".toMediaType()
+            override fun writeTo(sink: BufferedSink) {
+                content.readAll(sink)
+            }
+        }.toServerRequest()
+
+        client.newCall(oneShotRequest).execute().readByteStringBody()
+
+        val transaction = chuckerInterceptor.expectTransaction()
+        assertThat(transaction.isRequestBodyPlainText).isFalse()
```
vbuberen, line 519: Do we need to do 2 assertions in this test? This assertion looks to me more like checking of request body type detection works fine.

vbuberen (commented):
Looks good to me.

As to options suggested in the PR description I am in for adding field to the db. But suggest to not add it right away.

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
MiSikora: Yes, it shouldn't be here. It is a leftover from splitting #527. I'll remove it.

cortinico:
Related to https://github.com/ChuckerTeam/chucker/pull/544#discussion_r567242594

> Could you add a test also for duplex request here?

I believe we should unit test the Duplex scenario on the `RequestProcessor.processBody` level (i.e. if you get a request with `isDuplex`, the fields should stay untouched and so on).
However `RequestProcessor` is not test covered, so we can address it in a separate PR.

> or we can change the single message that we have. I'm open to both propositions.

+1 for changing the message. I don't think this change is worth the edit of the model (unless we plan to show more structured work on top of those fields).

cortinico (approved):

MiSikora:
> I believe we should unit test the Duplex scenario on the `RequestProcessor.processBody` level (i.e. if you get a request with `isDuplex`, the fields should stay untouched and so on).
> However `RequestProcessor` is not test covered, so we can address it in a separate PR.

Yes, that's why I did't want to deal with it here. It is a bigger question of what should be tested via `ChuckerInterceptor` and what should be tested via `RequestProcessor` (and soon for `ResponseProcessor`). So this would mean migrating some of the tests and I'd prefer, as you suggested, to do it in a separate PR.

vbuberen (approved):

PR #546, opened 2021-01-31
Add missing fix to changelog

MiSikora:
## :page_facing_up: Context

A bug fix wasn't mentioned in the changelog.

## :pencil: Changes

I updated the changelog.

## :paperclip: Related PR

#538 

## :no_entry_sign: Breaking

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Please add your entries according to this format.
 
 ## Unreleased
 
+### Fixed
+
+* Fixed not setting request body type correctly [#538].
+
 ### Removed
 
 * Removed parametrized `ChuckerInterceptor` constructor in favour of builder pattern. Constructor that accepts only `Context` is still available.


cortinico (approved):

cortinico:
As a note for the future, we could look into setting up https://danger.systems/kotlin/ to fail PR checks if the changelog file was not updated 🤔

MiSikora:
Do you mean to run it only as a warning and not something that fails CI?

IMO public changelog should be updated only if there are changes affecting users. If Danger requires updates to changelog with every PR I think it would be kind of counterproductive as git log is better suited for this job.

cortinico:
> Do you mean to run it only as a warning and not something that fails CI?

AFAIK it's customisable (as a warn or error)

MiSikora:
Yes, it should be. I used Danger in couple of projects and I really like it. Just wanted to suggest that if we integrate it, then changelog check shouldn't be an error. :)

Issue #469, opened 2020-10-05
When adding the dependency to my app there are multiple inflation exceptions happening at run time


proto-mvp:
I added the dependency as suggested on the read-me, and added the interceptor to the OkHttpClient through my Daagger Graph.
What happens on runtime is, that I get multiple inflation exceptions  **android.view.InflateException:**
Is this a known issue?

vbuberen:
Thanks for reporting. 
We need more info from the stacktrace you have.

Curious to see, since I am currently using the latest `3.3.0` in 3 projects already and everything works great.

proto-mvp:
_Error inflating class com.google.android.material.card.MaterialCardView_
Also, Similar error on another custom view extending FrameLayout

If it helps on anything,
the dependency is added on a module not the app's module.
Please let me know if that helps.
Thank you

cortinico:
> What happens on runtime is, that I get multiple inflation exceptions **android.view.InflateException:**
> Is this a known issue?

Not is not a known issue.
Can I ask you to use the issue template ([here](https://github.com/ChuckerTeam/chucker/blob/develop/.github/ISSUE_TEMPLATE/bug_report.md)) please? Ideally if you could provide also a stacktrace, it will help us debug

proto-mvp:
---
name: multiple inflation exceptions happening at run time
about: Create a report to help us improve

---

## :writing_hand: Describe the bug

Adding the gradle deps at the project's app module, I get on some screens the following crash with an InflateException

```
com.hdfd.uitest E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.hiddenpackage.appuitest, PID: 31440
    android.view.InflateException: Binary XML file line #2 in com.hiddenpackage.appuitest:layout/list_item_next_best_action: Binary XML file line #2 in com.hiddenpackage.appuitest:layout/list_item_next_best_action: Error inflating class com.hiddenpackage.app.pega.view.NextBestActionView
    Caused by: android.view.InflateException: Binary XML file line #2 in com.hiddenpackage.appuitest:layout/list_item_next_best_action: Error inflating class com.hiddenpackage.app.pega.view.NextBestActionView
    Caused by: java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
        at android.view.LayoutInflater.createView(LayoutInflater.java:852)
[...]

PR #549, opened 2021-02-07
Update Material to 1.3.0, switch to new CircularProgressIndicator

vbuberen:
## :page_facing_up: Context
Material 1.3.0 has some [new components for progress indicators ](https://github.com/material-components/material-components-android/blob/master/docs/components/ProgressIndicator.md). This PR updates Material to 1.3.0 and switches loading indicator used on transaction payload fragment to such new component.

## :pencil: Changes
- Updated Material to 1.3.0
- Switched to `CircularProgressIndicator` component

## :no_entry_sign: Breaking
Nothing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         // Google libraries
         appCompatVersion = '1.2.0'
         constraintLayoutVersion = '2.0.4'
-        materialComponentsVersion = '1.2.1'
+        materialComponentsVersion = '1.3.0'
         roomVersion = '2.2.6'
         fragmentVersion = '1.2.5'
         lifecycleVersion = '2.2.0'
--- a/library/src/main/res/layout/chucker_fragment_transaction_payload.xml
+++ b/library/src/main/res/layout/chucker_fragment_transaction_payload.xml
@@ -7,9 +7,8 @@
     android:animateLayoutChanges="true"
     tools:context="com.chuckerteam.chucker.internal.ui.transaction.TransactionPayloadFragment">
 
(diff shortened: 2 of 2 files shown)


MiSikora (approved):

cortinico (approved):

cortinico:
_nit: This will go in Chucker 4.x so probably we won't have a problem this time._ 
But bumping Material & AppCompat in the past caused problems as we impose the transitive dependency to our consumers (like #469).

I'm unsure how we should proceed on this side in the future 🤔

MiSikora:
I know that is not that simple due to how fragile Android is, but `SemVer` is clear about it in terms of versioning. UX, on the other hand, is the other thing. But IMO the only reasonable thing we can do is to keep up with the ecosystem.

From [SemVer](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api):

    What should I do if I update my own dependencies without changing the public API?

    That would be considered compatible since it does not affect the public API. Software that explicitly
    depends on the same dependencies as your package should have their own dependency specifications
    and the author will notice any conflicts. Determining whether the change is a patch level or minor level
    modification depends on whether you updated your dependencies in order to fix a bug or introduce new
    functionality. I would usually expect additional code for the latter instance, in which case it’s obviously
    a minor level increment.

vbuberen:
Agree here about need to be careful with such updates.
However, I think that we have quite a long road to 4.0.0, so by the time we release it we can expect the community to use `1.3.0` in most projects.

PR #554, opened 2021-02-09
Response processing refactor

MiSikora:
## :page_facing_up: Context

This is the final PR that is a result of splitting #527.

## :pencil: Changes

Similarly to the `RequestProcessor` class I added the `ResponseProcessor` class that does the analogous stuff. I changed one thing, where plain text body is set to be `plainText  = true` only when there is some content. I don't think it made sense to set it to `true` when there was no body.

## :paperclip: Related PR

#527

## :no_entry_sign: Breaking

No.

Rely on CI. Manually you can simply check the sample app.

## :stopwatch: Next steps

N/A

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -4,28 +4,11 @@ import android.content.Context
 import androidx.annotation.VisibleForTesting
 import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
 import com.chuckerteam.chucker.internal.support.CacheDirectoryProvider
-import com.chuckerteam.chucker.internal.support.DepletingSource
-import com.chuckerteam.chucker.internal.support.FileFactory
-import com.chuckerteam.chucker.internal.support.Logger
-import com.chuckerteam.chucker.internal.support.ReportingSink
 import com.chuckerteam.chucker.internal.support.RequestProcessor
-import com.chuckerteam.chucker.internal.support.TeeSource
-import com.chuckerteam.chucker.internal.support.contentType
-import com.chuckerteam.chucker.internal.support.hasBody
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -46,13 +29,34 @@ public class ChuckerInterceptor private constructor(
...
+
+    private val collector: ChuckerCollector
+
+    private val requestProcessor: RequestProcessor
+
+    private val responseProcessor: ResponseProcessor
+
+    init {
+        val context = builder.context
+        val maxContentLength = builder.maxContentLength
```
cortinico, line 42: Is there a reason why you moved those initialisation inside the `init{}` block?

cortinico (approved):
LGTM, just a minor comment

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
MiSikora: They aren't used by anything else except for processors. There is no need to keep them in the interceptor.

MiSikora: I guess it could be moved out of init block and reuse the values from the builder.

MiSikora:
Bumping because I don't want to get this message lost in a resolved comment. :)

@cortinico Ok, I pushed the change in 007cd559 without the init block. Let me know which one you prefer more. I'm ok either way.

cortinico:
> @cortinico Ok, I pushed the change in [007cd55](https://github.com/ChuckerTeam/chucker/commit/007cd559f69f65a52a08bf30acb43ad9d59e5654) without the init block. Let me know which one you prefer more. I'm ok either way.

I'd go for the solution without the `init{}` block as it's more declarative and easier to follow :)

vbuberen (approved):

PR #555, opened 2021-02-10
Custom body decoding

### LINKED ISSUE #523: Improved handling of encoded/binary bodies

MiSikora:
## :camera: Screenshots

Protobuf encoded request available in preview:

Screenshot_20210210_201958

## :page_facing_up: Context

There is an issue – #523.

## :pencil: Changes

I added `BodyEncoder` interface which instances can be installed in `ChukcerInterceptor`. They are then applied when bodies are being processed. First result that is non–null `String` is used as a body in transaction.

Exporting functionality was modified to accommodate for changes and export custom payloads as well.

I also edited sample to contain some custom processing. Unfortunately I did only for requests because I cannot find any public service that will echo proto data.

## :no_entry_sign: Breaking

No

Run the sample, check postman echo request, check exporting.

## :stopwatch: Next steps

`ChuckerInterceptorTest` needs some splitting into smaller units and/or moving test to respective processors. Also, `HttpBinClient` needs some clean-up because as mentioned some time ago it is now doing more than communicating with httpbin.
Closes #523.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,11 +146,15 @@ Chucker by default handles only plain text bodies. If you use a binary format li
 object ProtoDecoder : BinaryDecoder {
     fun decodeRequest(request: Request, body: ByteString): String? = if (request.isExpectedProtoRequest) {
         decodeProtoBody(body)
-    } else null
+    } else {
+        null
+    }
 
     fun decodeResponse(request: Response, body: ByteString): String? = if (request.isExpectedProtoResponse) {
         decodeProtoBody(body)
-    } else null
+    } else {
+        null
+    }
 }
 interceptorBuilder.addBodyDecoder(ProtoDecoder).build()
 ```
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ _A fork of [Chuck](https://github.com/jgilfelt/chuck)_
   * [Multi-Window](#multi-window-)
 * [Configure](#configure-)
   * [Redact-Header️](#redact-header-️)
+  * [Decode-Body](#decode-body-)
 * [Migrating](#migrating-)
 * [Snapshots](#snapshots-)
 * [FAQ](#faq-)
@@ -65,7 +66,7 @@ android {
 
 **That's it!** 🎉 Chucker will now record all HTTP interactions made by your OkHttp client.
 
-Historically, Chucker was distributed through JitPack. 
+Historically, Chucker was distributed through JitPack.
 You can find older version of Chucker here: [![JitPack](https://jitpack.io/v/ChuckerTeam/chucker.svg)](https://jitpack.io/#ChuckerTeam/chucker).
 
 ## Features 🧰
@@ -79,6 +80,7 @@ Don't forget to check the [changelog](CHANGELOG.md) to have a look at all the ch
 * **Empty release artifact** 🧼 (no traces of Chucker in your final APK).
 * Support for body text search with **highlighting** 🕵️‍♂️
 * Support for showing **images** in HTTP Responses 🖼
+* Support for custom decoding of HTTP bodies
 
 ### Multi-Window 🚪
 
@@ -112,6 +114,9 @@ val chuckerInterceptor = ChuckerInterceptor.Builder(context)
         // This is useful in case of parsing errors or when the response body
         // is closed before being read like in Retrofit with Void and Unit types.
         .alwaysReadResponseBody(true)
+        // Use decoder when processing request and response bodies. When multiple decoders are installed they
+        // are applied in an order they were added.
+        .addBodyDecoder(decoder)
         .build()
 
 // Don't forget to plug the ChuckerInterceptor inside the OkHttpClient
@@ -128,8 +133,26 @@ It is intended for **use during development**, and not in release builds or othe
 
 You can redact headers that contain sensitive information by calling `redactHeader(String)` on the `ChuckerInterceptor`.
 
+
 ```kotlin
 interceptor.redactHeader("Auth-Token", "User-Session");
+```
+
+### Decode-Body 📖
+
+Chucker by default handles only plain text bodies. If you use a binary format like, for example, Protobuf or Thrift it won't be automatically handled by Chucker. You can, however, install a custom decoder that is capable to read data from different encodings.
+
+```kotlin
+object ProtoDecoder : BinaryDecoder {
+    fun decodeRequest(request: Request, body: ByteString): String? = if (request.isExpectedProtoRequest) {
+        decodeProtoBody(body)
+    } else null
+
+    fun decodeResponse(request: Response, body: ByteString): String? = if (request.isExpectedProtoResponse) {
+        decodeProtoBody(body)
+    } else null
+}
+interceptorBuilder.addBodyDecoder(ProtoDecoder).build()
 ```
 
 ## Migrating 🚗
--- a/library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
@@ -1,0 +1,29 @@
+package com.chuckerteam.chucker.api
+
+import okhttp3.Request
+import okhttp3.Response
+import okio.ByteString
+import okio.IOException
+
+/**
+ * Decodes HTTP request and response bodies to human–readable texts.
+ */
+public interface BodyDecoder {
+    /**
+     * Returns a text representation of [body] that will be displayed in Chucker UI transaction,
+     * or `null` if [request] cannot be handled by this decoder. [Body][body] is no longer than
+     * [max content length][ChuckerInterceptor.Builder.maxContentLength] and is guaranteed to be
+     * gunzipped even if [request] has gzip header.
+     */
+    @Throws(IOException::class)
+    public fun decodeRequest(request: Request, body: ByteString): String?
+
+    /**
+     * Returns a text representation of [body] that will be displayed in Chucker UI transaction,
+     * or `null` if [response] cannot be handled by this decoder. [Body][body] is no longer than
+     * [max content length][ChuckerInterceptor.Builder.maxContentLength] and is guaranteed to be
+     * gunzipped even if [response] has gzip header.
+     */
+    @Throws(IOException::class)
+    public fun decodeResponse(response: Response, body: ByteString): String?
+}
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/PlainTextDecoder.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/PlainTextDecoder.kt
@@ -1,0 +1,28 @@
+package com.chuckerteam.chucker.internal.support
+
+import com.chuckerteam.chucker.api.BodyDecoder
+import okhttp3.Headers
+import okhttp3.MediaType
+import okhttp3.Request
+import okhttp3.Response
+import okio.ByteString
+import kotlin.text.Charsets.UTF_8
+
+internal object PlainTextDecoder : BodyDecoder {
+    override fun decodeRequest(
+        request: Request,
+        body: ByteString,
+    ) = body.tryDecodeAsPlainText(request.headers, request.body?.contentType())
+
+    override fun decodeResponse(
+        response: Response,
+        body: ByteString,
+    ) = body.tryDecodeAsPlainText(response.headers, response.body?.contentType())
+
+    private fun ByteString.tryDecodeAsPlainText(
+        headers: Headers,
+        contentType: MediaType?,
+    ) = if (headers.hasSupportedContentEncoding && isProbablyPlainText) {
+        string(contentType?.charset() ?: UTF_8)
+    } else null
+}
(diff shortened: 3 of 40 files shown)


MiSikora:
Lint decided to fail with versions updates for some reason on this PR. 🤷‍♂️

library/src/main/java/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
```diff
@@ -57,18 +55,23 @@ internal class RequestProcessor(
...
         }
     }
+
+    private fun decodePayload(request: Request, body: ByteString) = bodyDecoders.asSequence()
+        .mapNotNull { decoder ->
+            try {
+                Logger.info("Decoding with: $decoder")
+                decoder.decodeRequest(request, body)
+            } catch (e: IOException) {
+                Logger.error("Failed to process request payload", e)
```
cortinico, line 73:
```suggestion
                Logger.warn("Decoder $decoder failed to process request payload", e)
```

library/src/main/java/com/chuckerteam/chucker/internal/support/ResponseProcessor.kt
```diff
@@ -78,34 +82,40 @@ internal class ResponseProcessor(
...
+            transaction.responseBody = decodePayload(response, payload.readByteString())
         }
     }
 
+    private fun decodePayload(response: Response, body: ByteString) = bodyDecoders.asSequence()
+        .mapNotNull { decoder ->
+            try {
+                decoder.decodeResponse(response, body)
+            } catch (e: IOException) {
+                Logger.error("Failed to process response payload", e)
```
cortinico, line 106: Same as above.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -268,16 +268,16 @@ internal class TransactionPayloadFragment :
...
                 result.add(TransactionPayloadItem.ImageItem(responseBitmap, bitmapLuminance))
-            } else if (!isBodyPlainText) {
-                requireContext().getString(R.string.chucker_body_omitted).let {
-                    result.add(TransactionPayloadItem.BodyLineItem(SpannableStringBuilder.valueOf(it)))
-                }
-            } else {
-                if (bodyString.isNotBlank()) {
-                    bodyString.lines().forEach {
+            } else if (bodyString.isBlank()) {
+                if (!isBodyPlainText) {
```
cortinico, line 272: Can you merge this into a single condition?

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -566,4 +568,179 @@ internal class ChuckerInterceptorTest {
...
+        client.newCall(request).execute().readByteStringBody()
+
+        assertThat(statefulDecoder.didDecodeRequest).isFalse()
+        assertThat(statefulDecoder.didDecodeResponse).isFalse()
+    }
+
+    private class LiteralBodyDecoder : BodyDecoder {
+        override fun decodeRequest(request: Request, body: ByteString) = "Request"
+        override fun decodeResponse(response: Response, body: ByteString) = "Response"
+    }
```
cortinico, line 714: I believe that all the `*Decoder` + all the related tests should be moved to a `ChuckerInterceptor_withDecoderTest`

sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
```diff
@@ -145,4 +149,32 @@ class HttpBinClient(
...
+
+    private class PokemonProtoBodyDecoder : BodyDecoder {
+        override fun decodeRequest(request: Request, body: ByteString): String? {
+            return if (request.url.host.contains("postman", ignoreCase = true)) {
+                Pokemon.ADAPTER.decode(body).toString()
+            } else null
+        }
+
+        override fun decodeResponse(response: okhttp3.Response, body: ByteString): String? = null
+    }
```
cortinico, line 179: Please move this to its own file. Users could take this as inspiration and it's better if it's more prominent.

sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
```diff
@@ -145,4 +149,32 @@ class HttpBinClient(
             }
         )
     }
+
+    private fun getProtoResponse() {
+        val pokemon = Pokemon("Pikachu", level = 99)
+        val body = pokemon.encodeByteString().toRequestBody("application/protobuf".toMediaType())
+        val request = Request.Builder()
+            .url("https://postman-echo.com/post")
```
cortinico, line 157: This class is getting bloated :( It's called `HttpBinClient` but we do a lot of other stuff inside:

```
        downloadSampleImage(colorHex = "fff")
        downloadSampleImage(colorHex = "000")
        getResponsePartially()
        getProtoResponse()
```

Can you please move this code into `PostmanEchoClient` and then we'll take care of refactoring the status quo?

library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+public interface BodyDecoder {
+    /**
+     * Returns a text representation of [body] that will be displayed in Chucker UI transaction,
+     * or `null` if [request] cannot be handled by this decoder. [Body][body] is no longer than
+     * [max content length][ChuckerInterceptor.Builder.maxContentLength] and is guaranteed to be
+     * gunzipped even if [request] has gzip header.
+     */
+    @Throws(IOException::class)
+    public fun decodeRequest(request: Request, body: ByteString): String?
```
cortinico, line 19: nit: This is more an API design question.

What we're modelling here is an operation that could have 3 outcomes: success (`String`), failure (`IOException`), skipped (`null`).

Have you considered different approaches (e.g. `sealed classes` or the Kotlin `Result` type)? I'm not saying that the approach here is wrong, but I was wondering if we have some benefit in modelling it differently.

cortinico (approved):
Great work @MiSikora 🚀 Really excited to see this coming!

The code looks good on my end, I left mostly minor comments and nits.

README.md
```diff
@@ -128,10 +133,28 @@ It is intended for **use during development**, and not in release builds or othe
...
 
+### Decode-Body 📖
+
+Chucker by default handles only plain text bodies. If you use a binary format like, for example, Protobuf or Thrift it won't be automatically handled by Chucker. You can, however, install a custom decoder that is capable to read data from different encodings.
+
+```kotlin
+object ProtoDecoder : BinaryDecoder {
+    fun decodeRequest(request: Request, body: ByteString): String? = if (request.isExpectedProtoRequest) {
+        decodeProtoBody(body)
+    } else null
```
vbuberen, line 149: Nitpick: suggest to use curly braces in `else` part if we use them in `if` part.

README.md
```diff
@@ -128,10 +133,28 @@ It is intended for **use during development**, and not in release builds or othe
...
+
+    fun decodeResponse(request: Response, body: ByteString): String? = if (request.isExpectedProtoResponse) {
+        decodeProtoBody(body)
+    } else null
```
vbuberen, line 153: Same nitpick as above

library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
```diff
@@ -0,0 +1,29 @@
...
+import okio.ByteString
+import okio.IOException
+
+/**
+ * Decodes HTTP request and response bodies to human–readable texts.
+ */
+public interface BodyDecoder {
+    /**
+     * Returns a text representation of [body] that will be displayed in Chucker UI transaction,
+     * or `null` if [request] cannot be handled by this decoder. [Body][body] is no longer than
```
vbuberen, line 14: Looks like a typo with 2 `body`

library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
vbuberen: Agree with this nitpick that it might be a good idea to model the result to leave less ambiguity. 
For example, we could have something like `Failure` instead of null in case of unsupported content.

library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
```diff
@@ -0,0 +1,29 @@
...
+     * [max content length][ChuckerInterceptor.Builder.maxContentLength] and is guaranteed to be
+     * gunzipped even if [request] has gzip header.
+     */
+    @Throws(IOException::class)
+    public fun decodeRequest(request: Request, body: ByteString): String?
+
+    /**
+     * Returns a text representation of [body] that will be displayed in Chucker UI transaction,
+     * or `null` if [response] cannot be handled by this decoder. [Body][body] is no longer than
```
vbuberen, line 23: Same typo with 2 `body`

library/src/main/java/com/chuckerteam/chucker/internal/support/PlainTextDecoder.kt
```diff
@@ -0,0 +1,28 @@
...
+        response: Response,
+        body: ByteString,
+    ) = body.tryDecodeAsPlainText(response.headers, response.body?.contentType())
+
+    private fun ByteString.tryDecodeAsPlainText(
+        headers: Headers,
+        contentType: MediaType?,
+    ) = if (headers.hasSupportedContentEncoding && isProbablyPlainText) {
+        string(contentType?.charset() ?: UTF_8)
+    } else null
```
vbuberen, line 27: Nitpick: Suggest to have curly braces in `else` since there are braces in `if` part.

library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
cortinico: I believe is actually valid KDoc where you can specify the display text for a link 🤔

MiSikora: Yes, it is deliberate. It allows to link to `body` property and start the sentence with a capital letter.

library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
MiSikora: Same, not a typo. :)

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
MiSikora: Not in the current format, but I think it makes more sense to change to `when`. It'll make it clearer.

sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
MiSikora: I didn't want to move focus from decoding in this PR. Like mentioned in "next steps" section I'd prefer to restructure it in a separate PR if you don't mind.

library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
MiSikora: Yes, I considered it I don't think it is the right choice. I don't see how it helps us in the processing pipeline. We are interested only in the first, successfully decoded body.

`IOException` does not model a type of respons. Methods are annotated with it because tools like adapters generally can throw this type of exception so it removes form users a burden of needing to deal with it. And in Kotlin it is easy to forget since exceptions are not checked. Also, it allows us to log exceptions for the user so they have feedback information when something goes wrong. I would be either for keeping this part as is or to remove `@Throws` and move the responsibility to the users.

Returning `null` instead of some type is a nod towards libraries like Retrofit or Moshi, where factories return it when they can't handle input. I don't mind modelling it with a sealed class but I don't know if it brings anything to the table. If you feel that it would be helpful I'll change it.

sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
cortinico: Yup let's follow up on this

library/src/main/java/com/chuckerteam/chucker/api/BodyDecoder.kt
cortinico: Thanks for your insights 👍 I think we're fine with the current modelling for the reasons you mentioned + is the one that has the smaller API surface

vbuberen: This thread started with `nit` annotation, so let's leave it as it is :)

vbuberen (approved):

PR #556, opened 2021-02-12
Switch to Activity Result API

vbuberen:
## :page_facing_up: Context
Since Activity and Fragment with the new Activity Result API to replace things like `startActivityForResult()` are stable I have blown off dust from my branch where I tried to apply this new API in Chucker and updated dependencies to use stable library versions.

## :pencil: Changes
- Updated Activity to 1.2.0
- Updated Fragment to 1.3.0
- Replaced `startActivityForResult()` used to save file with payload with new Activity Result API

## :no_entry_sign: Breaking
Nothing breaking expected and I wasn't able to find any broken things while testing, but [Fragment 1.3.0](https://developer.android.com/jetpack/androidx/releases/fragment#1.3.0) has quite a lot of changes, like new state manager.

Nothing changed from UX point of view, so the save operation works as before.

## :stopwatch: Next steps
Deal with more deprecations in Chucker codebase.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,11 +5,12 @@ buildscript {
         coroutineVersion = '1.4.2'
 
         // Google libraries
+        activityVersion = '1.2.0-alpha07'
         appCompatVersion = '1.2.0'
         constraintLayoutVersion = '2.0.4'
         materialComponentsVersion = '1.3.0'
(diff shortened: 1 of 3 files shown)


MiSikora (approved):

cortinico (approved):
Nice 👍

PR #560, opened 2021-02-14
Allow to choose interception type in the sample

### LINKED ISSUE #389: Add the ability to choose interceptor types in the sample app
As a user I need the ability to capture data either as an _OkHttp_ `Interceptor` or as a `NetworkInterceptor` without having to edit the source code of the sample app.

**This feature should:**
* Provide a way to toggle between interceptor types in the sample app UI.  
* Default to capturing data as an `Interceptor` to preserve existing functionality.
* Each run of the sample service calls (kicked off by the "do HTTP activity" button) should use the current state of the toggle at the time of the button click.

- [x] Yes
- [ ] No

MiSikora:
## :camera: Screenshots

Screenshot_20210215_185231
Screenshot_20210215_185620

## :page_facing_up: Context

There is an open issue – #389.

I wanted to split `HttpBinClient` into smaller pieces but first I wanted to add the ability to test `ChuckerInterceptor` both as an application type and as a network type interceptor without changing it in the code.

## :pencil: Changes

I added a radio group to the sample UI that allows to choose how `ChuckerInterceptor` will be applied. Based on a selected radio button the interceptor is active in one or the other pipeline.

Behaviour is slightly different than the one described in #389. Interceptor type can be changed on the fly while requests are being performed. Making it toggleable only during idle periods would be too complex as it would require either two HTTP clients or some syncing and queueing mechanisma.

## :no_entry_sign: Breaking

No.

Play with the sample and toggle radio buttons.

## :stopwatch: Next steps

Closes #389.

### DIFF
--- a/sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
+++ b/sample/src/main/java/com/chuckerteam/chucker/sample/HttpBinClient.kt
@@ -5,6 +5,8 @@ import com.chuckerteam.chucker.api.ChuckerCollector
 import com.chuckerteam.chucker.api.ChuckerInterceptor
 import com.chuckerteam.chucker.api.RetentionManager
 import com.chuckerteam.chucker.sample.HttpBinApi.Data
+import com.chuckerteam.chucker.sample.InterceptorType.APPLICATION
+import com.chuckerteam.chucker.sample.InterceptorType.NETWORK
 import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.OkHttpClient
 import okhttp3.Request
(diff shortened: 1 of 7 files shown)


vbuberen:
I believe this PR closes #389

MiSikora:
Yes, you're absolutely right!

vbuberen (commented):
Good approach to selection. Looked to complex initially, but looking carefully saw that it is actually cool :)

vbuberen:
Screenshot_1613391326
I think that we need to wrap the layout with ScrollView even after applying changes I suggested in review above.

MiSikora:
> I think that we need to wrap the layout with ScrollView even after applying changes I suggested in review above.

I tried to check a different layout for landscape mode. Let me know what you think.

Screenshot_20210215_185620

vbuberen:
Yeah, this one much better.

vbuberen (approved):

cortinico (approved):
LGTM 👍

PR #562, opened 2021-02-14
Fix UI for emtpy body in request/response

vbuberen:
## :camera: Screenshots
Before:
Screenshot 2021-02-14 at 18 31 00
After: 
Screenshot 2021-02-14 at 19 27 36

## :page_facing_up: Context
Looks like #555 introduced a bug with payload UI. All empty bodies started to show `encoded or binary body omitted` string while they should be empty. Except fixing this I added a string resources to show in empty bodies transactions, like we do when trying to share transactions.

## :pencil: Changes
- Changed conditions in payload processing in `TransactionPayloadFragment`.
- Added string resource to clear indicate that body is empty.

## :no_entry_sign: Breaking
Nothing

Run sample app and check some transactions with emtpy requests/responses 

## :stopwatch: Next steps
I thing I found one more bug introduced by #555, but want to check how Chucker behaved in `3.4.0` to understand if it is really something new.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
@@ -247,17 +247,24 @@ internal class TransactionPayloadFragment :
 
             // The body could either be an image, plain text, decoded binary or not decoded binary.
             val responseBitmap = transaction.responseImageBitmap
-            when {
-                type == PayloadType.RESPONSE && responseBitmap != null -> {
-                    val bitmapLuminance = responseBitmap.calculateLuminance()
-                    result.add(TransactionPayloadItem.ImageItem(responseBitmap, bitmapLuminance))
-                }
-                bodyString.isNotBlank() -> bodyString.lines().forEach {
-                    result.add(TransactionPayloadItem.BodyLineItem(SpannableStringBuilder.valueOf(it)))
-                }
-                !isBodyPlainText -> {
+
+            if (type == PayloadType.RESPONSE && responseBitmap != null) {
+                val bitmapLuminance = responseBitmap.calculateLuminance()
+                result.add(TransactionPayloadItem.ImageItem(responseBitmap, bitmapLuminance))
+                return@withContext result
+            }
+
+            if (bodyString.isBlank()) {
+                val text = requireContext().getString(R.string.chucker_body_empty)
+                result.add(TransactionPayloadItem.BodyLineItem(SpannableStringBuilder.valueOf(text)))
+            } else {
+                if (!isBodyPlainText) {
                     val text = requireContext().getString(R.string.chucker_body_omitted)
                     result.add(TransactionPayloadItem.BodyLineItem(SpannableStringBuilder.valueOf(text)))
+                } else {
+                    bodyString.lines().forEach {
+                        result.add(TransactionPayloadItem.BodyLineItem(SpannableStringBuilder.valueOf(it)))
+                    }
                 }
             }
 


library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -247,17 +247,24 @@ internal class TransactionPayloadFragment :
...
+                val bitmapLuminance = responseBitmap.calculateLuminance()
+                result.add(TransactionPayloadItem.ImageItem(responseBitmap, bitmapLuminance))
+                return@withContext result
+            }
+
+            if (bodyString.isBlank()) {
+                val text = requireContext().getString(R.string.chucker_body_empty)
+                result.add(TransactionPayloadItem.BodyLineItem(SpannableStringBuilder.valueOf(text)))
+            } else {
+                if (!isBodyPlainText) {
```
MiSikora, line 261: This needs to be handled in a different way. Otherwise non-blank bodies that are not plain text but decoded from a different format will not show.

MiSikora:
Whoops! Thanks for catching that. I should have been more careful when changing `if` to `when`.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
vbuberen: Yeah, you are right, just checked. Not sure how we can handle it elegantly, since we also need to understand whenever the content is decoded.

MiSikora: Ahh, I'm getting it now. There's one thing I didn't consider when refactoring in #543. `isBodyPlainText` was `true` by default before (same for responses in #554), so it breaks here.

I'm not sure how to proceed with it. IMHO setting property that body is plain text when there is no body was a bug but this complicates things because we can't do check like this.

```kotlin
when {
  isImage -> showImage()
  bodyIsNotEmpty -> showBody()
  bodyIsPlainText -> showEmptyText()
  else -> showEncodedText()
}
```

Maybe the best way would be to have `isBodyEncoded` property instead of `isBodyPlainText`? This way combination of conditions should work properly I think.

vbuberen: `isBodyEncoded` would work, I believe.

vbuberen: Maybe it is something for a follow up PR?

MiSikora: I don't mind fixing it in another PR but wouldn't this make this PR just trade one bug for another?

MiSikora (approved):

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
vbuberen: I might misunderstood you initially, because I thought that you want to deal with it.

MiSikora: I don't mind dealing with it. I just want get it fixed one way or the other. :)

vbuberen: Ok, merging this one. Let's discuss that other fix in Slack.

PR #563, opened 2021-02-14
Add support for Brotli compression

### LINKED ISSUE #202: Support brotli compression

vbuberen:
## :camera: Screenshots
Screenshot 2021-02-14 at 19 56 25
Screenshot 2021-02-14 at 19 54 20

## :page_facing_up: Context
There is a very old issue with such feature request #202. Since we are now on OkHttp 4 it is time to add Brotli support.

## :pencil: Changes
- Added new dependency [OkHttp Brotli](https://github.com/square/okhttp/tree/master/okhttp-brotli)
- Updated `OkHttpUtils` to support uncompress operations for Brotli content

## :no_entry_sign: Breaking
Nothing breaking expected

Run sample and check `brotli` transaction response.

## :stopwatch: Next steps
I would like to add tests to this feature, but not sure how to do it properly to match our test suit. Would gladly discuss it here or in Slack and appreciate help/directions.

Closes #202


library/build.gradle
```diff
@@ -69,6 +69,7 @@ dependencies {
 
     api platform("com.squareup.okhttp3:okhttp-bom:$okhttpVersion")
     api "com.squareup.okhttp3:okhttp"
+    api "com.squareup.okhttp3:okhttp-brotli"
```
MiSikora, line 72: I'm not sure if we need Brotli package from OkHttp as it only adds an interceptor for Brotli compression. I think [`org.brotli:dec`](https://mvnrepository.com/artifact/org.brotli/dec/0.1.2) should be enough.

Also, it doesn't need to be `api`. We don't expose any types.

library/src/main/java/com/chuckerteam/chucker/internal/support/OkHttpUtils.kt
```diff
@@ -51,18 +54,23 @@ private val Headers.containsGzip: Boolean
...
         ?.let { it.toLowerCase(Locale.ROOT) in supportedEncodings }
         ?: true
 
-internal fun Source.uncompress(headers: Headers) = if (headers.containsGzip) {
-    gzip()
-} else {
-    this
+internal fun Source.uncompress(headers: Headers) = when {
+    headers.containsGzip -> gzip()
+    headers.containsBrotli -> BrotliInputStream(this.buffer().inputStream()).source()
```
MiSikora, line 72: I wonder if we should add a list of supported compression types in `ChuckerInterceptor` KDoc. I'm saying that because `BinaryDecoder` mentions that it the body will be gunzipped but now we also understand Brotli. This way in `BinaryDecoder` KDoc we could just mention that body will be uncompressed and to see which compression types are supported see `ChuckerInterceptor`.

MiSikora (commented):
Great addition! I left small comments. Also, one minor thing, we should mention dependencies and the feature in the changelog.

As for the question about testing – unfortunately package that we depend on does not offer `BrotliOutputStream` so it is not that straightforward. I see two ways we can test this.
1. Depend on some other package for tests only and use it to compress the data like in gzip tests. I found [this](https://github.com/google/brotli/blob/master/java/org/brotli/wrapper/enc/BrotliOutputStream.java) with some quick Google search but there might be better options.
2. Use hardcoded Brotli bytes and use them for requests and responses in tests and verify plain–text in transactions.

library/build.gradle
vbuberen: Yeah, correct. Overlooked that `BrotliInputStream` is part of another library.

vbuberen:
> Also, one minor thing, we should mention dependencies and the feature in the changelog.

🤦🏻 Forgot about it despite planning to add a few hours ago. Switched to that other issue with bodies and forgot. Will fix it.

> 2\. Use hardcoded Brotli bytes and use them for requests and responses in tests and verify plain–text in transactions.

So far I looked at this direction due to having no classes available to create Brotli compression ourselves. 
Might take inspiration from https://github.com/square/okhttp/blob/master/okhttp-brotli/src/test/java/okhttp3/brotli/BrotliInterceptorTest.kt

library/src/main/java/com/chuckerteam/chucker/internal/support/OkHttpUtils.kt
vbuberen: Good catch. We should mention in KDoc or add something in our Readme, because now it is not clear for end users that Chucker supports compressed payloads, so even having Gzip is something unclear.

library/build.gradle
```diff
@@ -67,6 +67,8 @@ dependencies {
 
     implementation "com.google.code.gson:gson:$gsonVersion"
 
+    implementation "org.brotli:dec:$brotliVersion"
```
cortinico, line 70: Open question:

Have you tried if this works if you use `BrotliInterceptor` from  `okhttp-brotli`:
https://search.maven.org/artifact/com.squareup.okhttp3/okhttp-brotli/4.2.0/jar
https://square.github.io/okhttp/4.x/okhttp-brotli/okhttp3.brotli/-brotli-interceptor/.

I suppose if one of our consumer is using Brotli + OkHTTP they will most likely use it. 
Therefore, I'm unsure if `Headers.containsBrotli: Boolean` would even work correctly (based on the order of the interceptors).

MiSikora: I think it should work. Gzip works in the same way. But it's worth checking.

vbuberen: Tested with `BrotliInterceptor` being before `ChuckerInterceptor` and the opposite order - everything works fine and response with headers is in place.

MiSikora:
One more thing that we have to consider is sharing cURL. Should we add support for Brotli there as well?

cortinico:
> One more thing that we have to consider is sharing cURL. Should we add support for Brotli there as well?

Potentially yes, but I don't think is a hard blocker for this PR.

vbuberen:
> One more thing that we have to consider is sharing cURL. Should we add support for Brotli there as well?

Yes, we should add add there as well. Can do in a follow-up PR to not block other opened PRs.

cortinico:
Can we merge this?

vbuberen:
> Can we merge this?

Not yet. Want to do a few minor updates later today. Updates the PR since decided to revive the development a little bit)

vbuberen:
Not it can be checked and merged if everything is fine.

I decided not to test uncompression since it would mean testing `brotli` library, which is already tested quite well.

MiSikora:
> I decided not to test uncompression since it would mean testing brotli library, which is already tested quite well.

I think a sanity test that a Brotli compressed response is understood by Chucker would be helpful. It's easy to make a mistake during refactoring or some other changes and introduce a bug this way.

vbuberen:
Added such test, which replicates the same test in `Brotli` library.

MiSikora (approved):

MiSikora:
I approved but I read test code wrongly. What I meant by test is to set Brotli encoded body in `MockWebServer` and then check if it uncompressed by `ChuckerInterceptor`. Something similar to this.

https://github.com/ChuckerTeam/chucker/blob/e8d20b57356f354194a5f8271f48a6a5274a5ea7/library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt#L75-L107

vbuberen:
Ok, understood. Agree, makes sense to test in `ChuckerInterceptorTest` as well

MiSikora:
> Quite confused on how to test it right, since in case of simple replication of such Gzip case I am getting that responseBody being encoded string I used initially. Not sure it should be like that. Could you help with that?

IMO that's a good thing. Gzip is built-in into `OkHttp` so users get decoded responses. Brotli, on the other hand, is not. So this test shows that `ChuckerInterceptor` does not interfere with the user's configuration and they need to install a Brotli interceptor themselves to read responses.

vbuberen:
Understood, thanks.

So we are adding a second test which in contrary to gzip should check that the response stays untouched and still compressed with brotli, right?

MiSikora:
Correct.

MiSikora (approved):

PR #580, opened 2021-03-20
Optimize gradle config & update dependencies

Goooler:
- Update dependencies
- Optimize gradle config


cortinico:
Thanks for the PR. Generally looks good but please revert the bumps of AGP, Detekt and KtLint.
For Detekt and KtLint please do it in separate PR as you also need to update the code.

Goooler:
> Thanks for the PR. Generally looks good but please revert the bumps of AGP, Detekt and KtLint.
> For Detekt and KtLint please do it in separate PR as you also need to update the code.

Reverted.

build.gradle
```diff
@@ -1,16 +1,20 @@
...
+        coroutineVersion = '1.4.3'
 
         // Google libraries
         activityVersion = '1.2.1'
         appCompatVersion = '1.2.0'
         constraintLayoutVersion = '2.0.4'
         materialComponentsVersion = '1.3.0'
         roomVersion = '2.2.6'
-        fragmentVersion = '1.3.0'
+        fragmentVersion = '1.3.1'
```
cortinico, line 17: You need to revert also this 👍  The reason is explained here: https://github.com/ChuckerTeam/chucker/pull/577

Goooler: Done

vbuberen (changes requested):
Thanks, but I am against merging this PR, since it is just some formatting things based on someone's preference, not real contribution. As to updating dependencies we have Dependabot.

PR #585, opened 2021-03-25
Bump activity & fragment versions

Goooler:
- Bump versions for solving build failed with this PR [Bump fragment-ktx from 1.3.0 to 1.3.1](https://github.com/ChuckerTeam/chucker/pull/577)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,12 +5,12 @@ buildscript {
         coroutineVersion = '1.4.2'
 
         // Google libraries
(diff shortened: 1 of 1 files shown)


vbuberen (approved):
Thanks for your PR. Looks good to me.

For future please note that we have Dependabot enabled, so we would get these updates anyway. Manual bump is also fine, but if it is not for some feature/another contribution, there is no need in so urgent updates of dependencies.

PR #588, opened 2021-03-25
Add a shortcut when creating `ChuckerInterceptor`

Goooler:
- Add a shortcut when creating `ChuckerInterceptor` for launch `MainActivity` like `LeakCanary`

screenshot


Goooler:
It seems api test failed, how can I solve it.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -30,6 +35,22 @@ public object Chucker {
             .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
     }
 
+    public fun Context.createShortcut() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+            getSystemService(ShortcutManager::class.java)?.let {
+                val shortcut = ShortcutInfo.Builder(this, "id1")
+                    .setShortLabel("Open chucker activity")
+                    .setLongLabel("Open chucker activity")
```
MiSikora, line 43: These should come from resources.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -30,6 +35,22 @@ public object Chucker {
             .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
     }
 
+    public fun Context.createShortcut() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+            getSystemService(ShortcutManager::class.java)?.let {
+                val shortcut = ShortcutInfo.Builder(this, "id1")
```
MiSikora, line 41: Please use a value that represent this shortcut instead of `id1`. Preferably move it to a `const`.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -30,6 +35,22 @@ public object Chucker {
...
+                    .setShortLabel("Open chucker activity")
+                    .setLongLabel("Open chucker activity")
+                    .setIcon(Icon.createWithResource(this, R.mipmap.chucker_ic_launcher_round))
+                    .setIntent(getLaunchIntent(this).setAction(Intent.ACTION_VIEW))
+                    .build()
+                val allShortcuts = it.dynamicShortcuts.toMutableList()
+                allShortcuts.add(shortcut)
+                it.dynamicShortcuts = allShortcuts
```
MiSikora, line 49: This can crash. You should catch exceptions that can occur here and add a meaningful log.

Also, instead of assigning new list you should just add one shortcut using `addDynamicShortcuts()`.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
MiSikora, line 41: Before creating and adding a shortcut you should first check if it does not already exist.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -30,6 +35,22 @@ public object Chucker {
             .setFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
     }
 
+    public fun Context.createShortcut() {
```
MiSikora, line 38: This function should follow structure of other functions in this object. `@JvmStatic` and accept context as an argument.

If you make it public it should also have a documentation. Moreover leaving it public requires mocked function in `no-op` library.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -30,6 +35,22 @@ public object Chucker {
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+            getSystemService(ShortcutManager::class.java)?.let {
+                val shortcut = ShortcutInfo.Builder(this, "id1")
+                    .setShortLabel("Open chucker activity")
+                    .setLongLabel("Open chucker activity")
+                    .setIcon(Icon.createWithResource(this, R.mipmap.chucker_ic_launcher_round))
```
MiSikora, line 44: Icon is added only as `xxxhdpi` resource. Please migrate to AVD or add other resolutions as well.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -53,6 +54,10 @@ public class ChuckerInterceptor private constructor(
         decoders,
     )
 
+    init {
+        builder.context.createShortcut()
```
MiSikora, line 58: I don't really like it enforced. IMO users should be able to opt–out of it with an XML boolean resource.

MiSikora (commented):
> It seems api test failed, how can I solve it.

By making function you added internal or by executing `apiDump` Gradle task.

vbuberen:
I am curios if we need this feature before proceeding with the reviw.

MiSikora:
Personally I like it. I won't fight for it, but I see it as a useful thing.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
vbuberen: Also, please change it to `Open Chucker`. No need to specify that it is activity.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
vbuberen: AVD would be much better. Please remove that added PNG

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
vbuberen: Don't have strong preference here, but don't like to such init.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: Movd hardcode into resources.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: Made it as a const value.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: Used `addDynamicShortcuts()` replace `setDynamicShortcuts()`

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: Have not found api in `ShortcutManager` to check if it exist, if I should read and write a SP value to check?

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: I made it as a internal function.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: Added other dpi icons

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
Goooler: I added a flag in builder to enable switch

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
MiSikora: You can get all shortcuts, map to IDs and check if our ID is in there.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -12,6 +17,8 @@ import com.chuckerteam.chucker.internal.ui.MainActivity
  */
 public object Chucker {
 
+    private const val SHORTCUT_ID = "shortcutId"
```
MiSikora, line 20: Make it more specific to Chucker. Like `chuckerShortcutId` or something. I'd prefer to avoid situation where someone has defined such a value as well and our shortcut is not added.

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -140,6 +148,14 @@ public class ChuckerInterceptor private constructor(
             this.decoders += decoder
         }
 
+        /**
+         * If set to `true`, [ChuckerInterceptor] will create a shortcut for your app
+         * just like what `LeakCanary` did, you can easily to access chucker's [MainActivity]
```
MiSikora, line 153: I'm not sure if we should mention `LeakCanary` here. Don't get me wrong, it is a great tool, but there's nothing special about `LeakCanary` when it comes to this feature. Maybe link to Android docs instead and highlight that it is a dynamic shortcut?

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
MiSikora: Raster icons are OK with me. I can make a follow–up PR with an AVD later on unless you want to change it. :)

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -30,6 +37,20 @@ public object Chucker {
...
+    internal fun createShortcut(context: Context) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+            context.getSystemService(ShortcutManager::class.java)?.let {
+                val shortcut = ShortcutInfo.Builder(context, SHORTCUT_ID)
+                    .setShortLabel(context.getString(R.string.chucker_shortcut_label))
+                    .setLongLabel(context.getString(R.string.chucker_shortcut_label))
+                    .setIcon(Icon.createWithResource(context, R.mipmap.chucker_ic_launcher_round))
+                    .setIntent(getLaunchIntent(context).setAction(Intent.ACTION_VIEW))
+                    .build()
+                it.addDynamicShortcuts(listOf(shortcut))
```
MiSikora, line 49: This also can throw and should be handled here. Look at the [docs](https://developer.android.com/reference/android/content/pm/ShortcutManager#addDynamicShortcuts(java.util.List%3Candroid.content.pm.ShortcutInfo%3E)).

Goooler: But I see it already been catched.
image

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: Changed

library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
Goooler: Removed that

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: Catched

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -30,6 +37,34 @@ public object Chucker {
...
+    /**
+     * Create a shortcut to launch Chucker UI.
+     * @param context An Android [Context].
+     */
+    internal fun createShortcut(context: Context) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
+            run outside@{
+                context.getSystemService(ShortcutManager::class.java)?.let { sm ->
+                    sm.dynamicShortcuts.forEach {
+                        if (it.id == SHORTCUT_ID) return@outside
```
MiSikora, line 49: No need for `run` with an outside label. You can return here with `return@let`.

MiSikora (approved):
Left one minor comment, but other than that, LGTM. 🚀 

Thank you for this contribution! I'll wait with merging however, for a second review/opinion.

You'll also need to execute `apiDump` Gradle task because a public function was added to the builder.

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
Goooler: Your are right

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -30,6 +37,34 @@ public object Chucker {
...
+                        .setLongLabel(context.getString(R.string.chucker_shortcut_label))
+                        .setIcon(
+                            Icon.createWithResource(context, R.mipmap.chucker_ic_launcher_round)
+                        )
+                        .setIntent(getLaunchIntent(context).setAction(Intent.ACTION_VIEW))
+                        .build()
+                    try {
+                        sm.addDynamicShortcuts(listOf(shortcut))
+                    } catch (ignored: Throwable) {
+                    }
```
MiSikora: NIT: Maybe log it? We have a logger in this project for such cases.

Goooler: Logged it.

cortinico:
I think this is a nice addition 👍 
Should we add a note to the README?

library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
vbuberen: Ok, let's do that switch to AVD in a follow-up PR ourselves.

vbuberen:
> Should we add a note to the README?

I will take care of it in a follow-up PR for this feature.

vbuberen (approved):

PR #591, opened 2021-03-27
Refine code

Goooler:
## :page_facing_up: Context

Some code sections should be refined, could be replaced with properer ones.

## :pencil: Changes

- Remove redundant comma.
- Replace `?: ""` with `orEmpty()`.
- Add params with default value.
- Use `LiveData#Observe()` SAM syntax.
And more...
## :no_entry_sign: Breaking

No.

No functional changes
## :stopwatch: Next steps

N/A.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
@@ -4,11 +4,13 @@ import androidx.lifecycle.LiveData
 import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
 import com.chuckerteam.chucker.internal.data.entity.HttpTransactionTuple
 import com.chuckerteam.chucker.internal.data.room.ChuckerDatabase
+import com.chuckerteam.chucker.internal.data.room.HttpTransactionDao
 import com.chuckerteam.chucker.internal.support.distinctUntilChanged
 
-internal class HttpTransactionDatabaseRepository(private val database: ChuckerDatabase) : HttpTransactionRepository {
+internal class HttpTransactionDatabaseRepository(database: ChuckerDatabase) :
+    HttpTransactionRepository {
 
-    private val transactionDao get() = database.transactionDao()
+    private val transactionDao: HttpTransactionDao = database.transactionDao
 
     override fun getFilteredTransactionTuples(code: String, path: String): LiveData<List<HttpTransactionTuple>> {
         val pathQuery = if (path.isNotEmpty()) "%$path%" else "%"
--- a/library/src/main/java/com/chuckerteam/chucker/internal/data/room/ChuckerDatabase.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/data/room/ChuckerDatabase.kt
@@ -9,7 +9,7 @@ import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
 @Database(entities = [HttpTransaction::class], version = 6, exportSchema = false)
 internal abstract class ChuckerDatabase : RoomDatabase() {
 
-    abstract fun transactionDao(): HttpTransactionDao
+    abstract val transactionDao: HttpTransactionDao
 
     companion object {
         private const val OLD_DB_NAME = "chuck.db"
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/ContextExt.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/ContextExt.kt
@@ -6,17 +6,17 @@ import com.google.android.material.dialog.MaterialAlertDialogBuilder
 
 internal fun Context.showDialog(
     dialogData: DialogData,
-    onPositiveClick: (() -> Unit)?,
-    onNegativeClick: (() -> Unit)?
+    onPositiveClick: () -> Unit = {},
+    onNegativeClick: () -> Unit = {}
 ) {
     MaterialAlertDialogBuilder(this)
         .setTitle(dialogData.title)
         .setMessage(dialogData.message)
         .setPositiveButton(dialogData.positiveButtonText) { _, _ ->
-            onPositiveClick?.invoke()
+            onPositiveClick()
         }
         .setNegativeButton(dialogData.negativeButtonText) { _, _ ->
-            onNegativeClick?.invoke()
+            onNegativeClick()
         }
         .show()
 }
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/ContextExt.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/ContextExt.kt
@@ -6,17 +6,17 @@ import com.google.android.material.dialog.MaterialAlertDialogBuilder
 
 internal fun Context.showDialog(
     dialogData: DialogData,
-    onPositiveClick: () -> Unit = {},
-    onNegativeClick: () -> Unit = {}
+    onPositiveClick: (() -> Unit)? = null,
+    onNegativeClick: (() -> Unit)? = null
 ) {
     MaterialAlertDialogBuilder(this)
         .setTitle(dialogData.title)
         .setMessage(dialogData.message)
         .setPositiveButton(dialogData.positiveButtonText) { _, _ ->
-            onPositiveClick()
+            onPositiveClick?.invoke()
         }
         .setNegativeButton(dialogData.negativeButtonText) { _, _ ->
-            onNegativeClick()
+            onNegativeClick?.invoke()
         }
         .show()
 }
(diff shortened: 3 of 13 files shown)


MiSikora:
Please use our PR template and remove sections only if they are irrelevant to changes.

Goooler:
> Please use our PR template and remove sections only if they are irrelevant to changes.

Can you review it now ?

library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
```diff
@@ -4,11 +4,13 @@ import androidx.lifecycle.LiveData
...
 import com.chuckerteam.chucker.internal.data.room.ChuckerDatabase
+import com.chuckerteam.chucker.internal.data.room.HttpTransactionDao
 import com.chuckerteam.chucker.internal.support.distinctUntilChanged
 
-internal class HttpTransactionDatabaseRepository(private val database: ChuckerDatabase) : HttpTransactionRepository {
+internal class HttpTransactionDatabaseRepository(database: ChuckerDatabase) :
+    HttpTransactionRepository {
 
-    private val transactionDao get() = database.transactionDao()
+    private val transactionDao: HttpTransactionDao = database.transactionDao
```
MiSikora, line 13: Why this change?

library/src/main/java/com/chuckerteam/chucker/internal/data/room/ChuckerDatabase.kt
```diff
@@ -9,7 +9,7 @@ import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
 @Database(entities = [HttpTransaction::class], version = 6, exportSchema = false)
 internal abstract class ChuckerDatabase : RoomDatabase() {
 
-    abstract fun transactionDao(): HttpTransactionDao
+    abstract val transactionDao: HttpTransactionDao
```
MiSikora, line 12: Same here, why change this?

library/src/main/java/com/chuckerteam/chucker/internal/support/ContextExt.kt
```diff
@@ -6,17 +6,17 @@ import com.google.android.material.dialog.MaterialAlertDialogBuilder
 
 internal fun Context.showDialog(
     dialogData: DialogData,
-    onPositiveClick: (() -> Unit)?,
-    onNegativeClick: (() -> Unit)?
+    onPositiveClick: () -> Unit = {},
+    onNegativeClick: () -> Unit = {}
```
MiSikora, line 10: I'm ok with default arguments but why make them not nullable?

library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
Goooler: Removed val on the `database` in consctructor, transactionDao can be initiallized directly when constructing.

library/src/main/java/com/chuckerteam/chucker/internal/data/room/ChuckerDatabase.kt
Goooler: Kotlin feature, use`val` replace `fun` in`interface` or `abstract class`.

library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
MiSikora: Sorry, I don't get this response. Which constructor and which initialization do you have on mind? DAOs are lazy initialized and synchronized in Room.

library/src/main/java/com/chuckerteam/chucker/internal/data/room/ChuckerDatabase.kt
MiSikora: I don't know why would you like to change it. There is nothing wrong or unidiomatic with using a function.

library/src/main/java/com/chuckerteam/chucker/internal/support/ContextExt.kt
Goooler: If the arguments are nullable and with default value, look like `(() -> Unit)? = null`, it's chatty and not neat.

MiSikora: What's chatty and not neat and by which metric? Having nullable arguments saves object allocations.

library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
Goooler: image
Here, `database` is just a injected param only provide `dao` for `transactionDao`

library/src/main/java/com/chuckerteam/chucker/internal/data/room/ChuckerDatabase.kt
Goooler: Yeah, behaves in the same way. But I think use a field instead of function here is more graceful.

library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
MiSikora: Ah, thanks. Didn't catch that. Why would it be a problem to keep a reference to the database instead of DAO?

library/src/main/java/com/chuckerteam/chucker/internal/support/ContextExt.kt
Goooler: Ok, for needing not to allocate objects, I will change them to nullable.

library/src/main/java/com/chuckerteam/chucker/internal/data/room/ChuckerDatabase.kt
MiSikora: This is just opinion based. I don't mind the output code but I don't like to change the code just for cosmetic changes when there are no objective indicators to do so.

library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
Goooler: `database` just used to provide `dao` for `transactionDao`, we can remove the reference after constructor for min visibility.

library/src/main/java/com/chuckerteam/chucker/internal/data/room/ChuckerDatabase.kt
Goooler: Another reason, for Java code to access it, we usually use `getTransactionDao()`, but in Kotlin we are suggested to use `transactionDao`. What I modified can be more compatible with Java and Kotlin.

MiSikora: This would be true, if this class was exposed. Don't get me wrong, I'm not requesting for you to change it back to a function but it is not something that refines the code. It is just formatting it according to personal preferences.

library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
MiSikora: Visibility is still the same. `HttpTransactionDatabaseRepository` has access to database anyway. More meaningful would be to pass DAO as a constructor parameter.

vbuberen:
After looking at changes in this PR I see that most things changed are just opinion based, not something really meaningful and really that better/worse depends on someone's preference. 
While shortcut feature was a real enhancement this PR looks like some change just for the sake of change.

@Goooler I would like to ask you to stop doing such PRs, since the value of it is quite questionable (like dependencies update a few days ago or changing callbacks from nullable to not-nullable to only return it back later in the same PR.

library/src/main/java/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
Goooler: Yeah, it's only duty is as a constructor parameter, should not let other functions to access.

Goooler:
@vbuberen Thanks for your patience.

MiSikora:
For the reasons mentioned by @vbuberen with which I agree. I'm going to close this PR. These changes feel mostly opinionated without a lot of substance. Don't feel refrained from making PRs, but next time maybe try with issues from the backlog and not only cosmetic changes. Also, if you have ideas it is usually better to start with an issue.

Goooler:
> For the reasons mentioned by @vbuberen with which I agree, I'm going to close this PR. These changes feel mostly opinionated without a lot of substance. Don't feel refrained from making PRs, but next time maybe try with issues from the backlog and not only cosmetic changes. Also, if you have ideas it is usually better to start with an issue so we can talk it through.

Thanks for your patience.

PR #593, opened 2021-03-27
Add immutable flag to pending intents

MiSikora:
## :page_facing_up: Context

Android 31 [requires](https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability) to specify mutability of pending intents.

## :pencil: Changes

I added mutability flag to our pending intents.

## :no_entry_sign: Breaking

No.

Not sure, run on Android 31?

## :stopwatch: Next steps

Either cherry–pick it to `3.x` or plan release of `4.x` soonish.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
@@ -43,7 +43,7 @@ internal class NotificationHelper(val context: Context) {
             context,
             TRANSACTION_NOTIFICATION_ID,
             Chucker.getLaunchIntent(context),
-            PendingIntent.FLAG_UPDATE_CURRENT
+            PendingIntent.FLAG_UPDATE_CURRENT or immutableFlag()
         )
     }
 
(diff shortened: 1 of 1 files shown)


vbuberen (approved):

library/src/main/java/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
```diff
@@ -118,12 +118,18 @@ internal class NotificationHelper(val context: Context) {
...
+                PendingIntent.FLAG_ONE_SHOT or immutableFlag()
             )
             return NotificationCompat.Action(R.drawable.chucker_ic_delete_white, clearTitle, intent)
         }
 
     fun dismissNotifications() {
         notificationManager.cancel(TRANSACTION_NOTIFICATION_ID)
     }
+
+    private fun immutableFlag() = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
koral--, line 130: Do we really need this check?
Flags are generally forwards compatible.
During compilation constant will be inlined and in runtime on pre-M those bits won't be read.
Warning will be generated about using constants from higher API levels than min SDK but in this case such usage is legitimate IMHO.

MiSikora: I'm not sure what are you proposing. I don't want IDE to highlight this and Lint to fail. Do you suggest to add `@SuppressLint("InlinedApi")`? R8 can remove this condition anyway with `-assumevalues`. Then there is dex2oat and finally ART itself.

koral--: > Do you suggest to add @SuppressLint("InlinedApi")

Exactly.

> R8 can remove this condition

True, but not everyone use R8.

MiSikora: > True, but not everyone use R8.

I would say that if someone does not use R8 then they don't care about optimizations anyway. :)

cortinico: I'd be more for having the explicit if-then-else. It's more aligned with the rest of the codebase and makes the code more explicit.

PR #600, opened 2021-04-04
Replace IntentService with JobIntentService

vbuberen:
## :page_facing_up: Context
Chucker used `IntentService` for notification action to clear transactions database, which is currently deprecated.
To address this deprecation this PR replaces `IntentService` with `JobIntentService`. 

Since `JobIntentService` requires call to `enqueueWork()` function I had to use `Broadcast` and declare `BroadcastReceiver` which would call `enqueueWork()` when required.
Additionally, I had to add `WAKE_LOCK` permission for pre-Oreo devices or Chucker would crash on devices with API < 26.

## :pencil: Changes
- Switched from `IntentService` to `JobIntentService`
- Declared a `BroadcastReceiver` to receive `Clear` action events.
- Added `WAKE_LOCK` permission for pre-Oreo devices.

## :no_entry_sign: Breaking
Nothing breaking expected

No special requirements. Just try `Clear` action from some Chucker notification and everything should work.

### DIFF
--- a/library/src/main/AndroidManifest.xml
+++ b/library/src/main/AndroidManifest.xml
@@ -9,6 +9,8 @@
         </intent>
     </queries>
 
+    <uses-permission android:name="android.permission.WAKE_LOCK"/>
+
     <application>
         <activity
             android:name="com.chuckerteam.chucker.internal.ui.MainActivity"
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/support/ClearDatabaseService.kt
```diff
@@ -1,15 +1,20 @@
...
 import com.chuckerteam.chucker.internal.data.repository.RepositoryProvider
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
 
-internal class ClearDatabaseService : IntentService(CLEAN_DATABASE_SERVICE_NAME) {
+internal class ClearDatabaseService : JobIntentService() {
     private val scope = MainScope()
 
-    override fun onHandleIntent(intent: Intent?) {
+    fun enqueueWork(context: Context, work: Intent) {
```
MiSikora, line 13: I think this should be a method on the companion object. Otherwise it is awkward to use with instantiating the service in order to enqueue work on it.

vbuberen: Agree

MiSikora (approved):

library/src/main/AndroidManifest.xml
```diff
@@ -9,6 +9,8 @@
         </intent>
     </queries>
 
+    <uses-permission android:name="android.permission.WAKE_LOCK"/>
```
koral--, line 12: Shouldn't it have android:maxSdkVersion set? PR description suggests that it is only needed for API < 26.

vbuberen: TIL - didn't know about this property.

vbuberen:
Limited max SDK for permission. Tested on different versions, including latest versions - looks like are good.

cortinico:
Sorry I arrived late to the party, the code looks great on my end 👌 
Little small note @vbuberen, the CHANGELOG needs to include this change :) Are you going to do it or should I do it?

vbuberen:
@cortinico I wanted to add it to changelogs, but later understood that this change is something that happened internally and doesn't affect our users. This is why I skipped adding it.

PR #601, opened 2021-04-04
Limit Jcenter usage

vbuberen:
## :page_facing_up: Context
Future-proofing Chucker by limiting usage of Jcenter. Most dependencies now resolved through `mavenCentral` instead. 

~~However, there was one issue with `org.jetbrains.trove4j:trove4j`, which is used by `android-lint`, so I had to add `gradlePluginPortal()` to `allprojects` section or the build process would fail for me. Not sure how it helps, considering that I couldn't find `trove4j` on Gradle portal.~~

Jcenter left only for dependencies, which can't be resolved via `mavenCentral`.

## :pencil: Changes
- Removed `jcenter` from repositories in `buildscript`.
- Added `mavenCentral` to `allprojects` section.
- (UPD: 08.04.2021) Limited usage of Jcenter only to required dependencies in `allprojects` section: `trove4j` and `kotlinx-html-jvm`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,6 @@ buildscript {
     }
 
     repositories {
-        jcenter()
         google()
         gradlePluginPortal()
     }
@@ -71,8 +70,9 @@ allprojects {
     group = GROUP
 
     repositories {
-        jcenter()
         google()
+        mavenCentral()
+        gradlePluginPortal()
     }
 
     tasks.withType(Test) {
--- a/build.gradle
+++ b/build.gradle
@@ -72,7 +72,12 @@ allprojects {
     repositories {
         google()
         mavenCentral()
-        gradlePluginPortal()
+        jcenter {
+            content {
+                includeModule("org.jetbrains.trove4j", "trove4j")
+                includeModule("org.jetbrains.kotlinx", "kotlinx-html-jvm")
+            }
+        }
     }
 
     tasks.withType(Test) {


koral--:
> However, there was one issue with `org.jetbrains.trove4j:trove4j`, which is used by `android-lint`, so I had to add `gradlePluginPortal()` to `allprojects` section or the build process would fail for me. Not sure how it helps, considering that I couldn't find `trove4j` on Gradle portal.
> 

According to mvnrepository it can be found in spring plugins repo https://mvnrepository.com/artifact/org.jetbrains.trove4j/trove4j

vbuberen:
> According to mvnrepository it can be found in spring plugins repo https://mvnrepository.com/artifact/org.jetbrains.trove4j/trove4j

Yes, I saw that. But I wasn't able to find this thing on https://plugins.gradle.org/ (which I suppose is the place for gradlePluginPortal() ). And this is why I am curios why adding `gradlePluginPortal()` helped.

build.gradle
```diff
@@ -71,8 +70,9 @@ allprojects {
     group = GROUP
 
     repositories {
-        jcenter()
         google()
+        mavenCentral()
+        gradlePluginPortal()
```
cortinico, line 75: I'm a bit against this change, mostly because we're _polluting_ the runtime/compile dependencies with a repository that should not offer such dependencies (but only build plugins).

The reason why it worked is most likely because Gradle Portal is mirroring other repos.
I think we can add a:

```
// TODO Remove me once trove4j migrates away from JCenter.
jcenter {
    content {
        includeModule("org.jetbrains.trove4j", "trove4j")
    }
}
```

To restrict usage of `jcenter()` to only `trove4j`

MiSikora: I agree. I don't think there is a need to rush this.

vbuberen:
One more addition - not only `trove4j` is missing, but since we are using `detekt` we are losing `kotlinx-html-jvm`. 
Added it as well in the latest commit.

cortinico (approved):

cortinico:
> but since we are using `detekt` we are losing `kotlinx-html-jvm`.
> Added it as well in the latest commit.

Yup that is tracked here: https://github.com/detekt/detekt/pull/3455

vbuberen:
> Yup that is tracked here: [detekt/detekt#3455](https://github.com/detekt/detekt/pull/3455)

Yeah, checked that out already 😄

PR #602, opened 2021-04-06
Add TargetApi annotation

Goooler:
## :pencil: Changes

Add `@TargetApi` for `createShortcut()`

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/Chucker.kt
@@ -1,5 +1,6 @@
 package com.chuckerteam.chucker.api
 
+import android.annotation.TargetApi
 import android.content.Context
 import android.content.Intent
 import android.content.pm.ShortcutInfo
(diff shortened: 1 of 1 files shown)


vbuberen:
Thanks for another contribution.

However, I am not sure what is the advantage of having this annotation? I haven't seen any lint warnings so far about this function.

cortinico:
Agree to what @vbuberen said

> I haven't seen any lint warnings so far about this function.

I believe that's because the method contains an if-then-else with an API Level check, so that's why Lint is not complaining.

Goooler:
> I believe that's because the method contains an if-then-else with an API Level check, so that's why Lint is not complaining.


`@RequiresApi` will make lint warning, `@TargetApi` will not, add this annotation just for attention.

MiSikora:
[Documentation](https://developer.android.com/reference/android/annotation/TargetApi) does not indicate that usage of this annotation is for attention.

It's for making Lint quiet.

> Indicates that Lint should treat this type as targeting a given API level, no matter what the project target is.

vbuberen:
This brings up to my initial question - why we need this PR? It looks like it won't bring any benefits, so I am closing it.

PR #609, opened 2021-04-13
Feature/clickable url

bilgehankalkan:
## :page_facing_up: Context
Response body sometimes can contain URL's of feature network requests, with this development user can navigate between requests. Also other URL's will be opened in browser( via Intent.VIEW action).

## :pencil: Changes
- Add new method to query previous request by url.
- Add BodyLineItemCreator to assign ClickableSpan to response body field values.

## :no_entry_sign: Breaking
- This implementation does not changes current behaviour of API.

Sample response fields are enough for testing JSON fields. To test XML fields, custom response or implementing to another library that can provide XML response is required.

### DIFF
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
@@ -1,0 +1,43 @@
+package com.chuckerteam.chucker.internal.ui.transaction
+
+import android.text.Spannable
+import android.text.SpannableStringBuilder
+import android.text.style.ClickableSpan
+import android.view.View
+import com.chuckerteam.chucker.internal.support.Logger
+
+internal class BodyLineItemCreator(private val bodyLine: String, private val onUrlClicked: (String) -> Unit) {
+
+    fun create(): SpannableStringBuilder =
+        SpannableStringBuilder().apply {
+            append(bodyLine)
+            checkUrl { url, range ->
+                val clickableSpan = object : ClickableSpan() {
+                    override fun onClick(widget: View) {
+                        onUrlClicked(url)
+                    }
+                }
+
+                setSpan(clickableSpan, range.first, range.last + 1, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
+            }
+        }
+
+    private fun checkUrl(link: (String, IntRange) -> Unit) {
+        val matchGroup = getValue() ?: return
+        try {
+            if (URL_MATCHER.toRegex().matches(matchGroup.value)) {
+                link.invoke(matchGroup.value, matchGroup.range)
+            }
+        } catch (ignored: IllegalStateException) {
+            Logger.info("regex not matched for bodyLine $bodyLine")
+        }
+    }
+
+    private fun getValue(): MatchGroup? = KEY_VALUE_SEPARATOR.toRegex().matchEntire(bodyLine)?.groups?.get(2)
+
+    companion object {
+
+        private const val KEY_VALUE_SEPARATOR = "(.*)?\".*\":\\s\"(.*)\"(.*)?"
+        private const val URL_MATCHER = "^(https?|ftp|file)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]"
+    }
+}
--- a/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
@@ -40,7 +40,7 @@ internal class BodyLineItemCreator(private val bodyLine: String, private val onU
 
     companion object {
 
-        private const val KEY_VALUE_SEPARATOR = "\\s*\".*\":\\s\"(.*)\"|<.*>(.*)</.*?"
+        private const val KEY_VALUE_SEPARATOR = "\\s*\".*\":\\s\"(.*)\",?|<.*>(.*)</.*?"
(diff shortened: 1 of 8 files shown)


library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
```diff
@@ -0,0 +1,46 @@
...
+import android.text.Spannable
+import android.text.SpannableStringBuilder
+import android.text.style.ClickableSpan
+import android.view.View
+import com.chuckerteam.chucker.internal.support.Logger
+
+internal class BodyLineItemCreator(private val bodyLine: String, private val onUrlClicked: (String) -> Unit) {
+
+    fun create(): SpannableStringBuilder =
+        SpannableStringBuilder().apply {
```
MiSikora, line 12: You can use `buildSpannedString()` instead.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
```diff
@@ -0,0 +1,46 @@
...
+                }
+
+                setSpan(clickableSpan, range.first, range.last + 1, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
+            }
+        }
+
+    private fun checkUrl(link: (String, IntRange) -> Unit) {
+        val matchGroup = getValue() ?: return
+        try {
+            if (URL_MATCHER.toRegex().matches(matchGroup.value)) {
```
MiSikora, line 28: Instead of matching URL by regex use something like `String.toHttpUrlOrNull()`.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
```diff
@@ -0,0 +1,46 @@
...
+                link.invoke(matchGroup.value, matchGroup.range)
+            }
+        } catch (ignored: IllegalStateException) {
+            Logger.info("regex not matched for bodyLine $bodyLine")
+        }
+    }
+
+    private fun getValue(): MatchGroup? {
+        val groups = KEY_VALUE_SEPARATOR.toRegex().matchEntire(bodyLine)?.groups.orEmpty().toList()
```
MiSikora, line 37: Create regex only once in the companion object.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
```diff
@@ -0,0 +1,46 @@
...
+        return groups.getOrNull(2) ?: groups.getOrNull(1)
+    }
+
+    companion object {
+
+        private const val KEY_VALUE_SEPARATOR = "\\s*\".*\":\\s\"(.*)\",?|<.*>(.*)</.*?"
```
MiSikora, line 43: You can avoid escaping characters by using triple quote to build the string.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -297,6 +299,13 @@ internal class TransactionPayloadFragment :
         }
     }
 
+    private fun onBodyLineItemClicked(url: String) = lifecycleScope.launch {
+        when (val action = viewModel.onBodyLineItemClicked(url)) {
+            is Action.OpenTransaction -> TransactionActivity.start(requireContext(), action.transactionId)
+            is Action.OpenUrl -> startActivity(Intent(Intent.ACTION_VIEW).apply { data = Uri.parse(url) })
```
MiSikora, line 305: No need for `apply`. `Intent.setData()` returns `Intent`.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/TransactionViewModel.kt
```diff
@@ -49,14 +49,32 @@ internal class TransactionViewModel(transactionId: Long) : ViewModel() {
     fun encodeUrl(encode: Boolean) {
         mutableEncodeUrl.value = encode
     }
+
+    suspend fun onBodyLineItemClicked(url: String): Action {
+        val request = RepositoryProvider.transaction().getByUrl(url)
```
MiSikora, line 54: I'm rather hesitant about this functionality. URL by themselves are not unique. `POST`, `PUT`, `GET` and so on can use the same path. Moreover, one endpoint can be invoked multiple times and there will be many instances of it in our database and this just navigates to whichever happens to be latest.

MiSikora (commented):
Thanks for the PR but I'm rather hesitant towards the feature. I left some comments about the code but I see one major issue with this functionality. It works for some simple cases but most APIs are hidden behind authentication layer so opening it in the browser does not give anything.

library/src/main/java/com/chuckerteam/chucker/internal/ui/transaction/BodyLineItemCreator.kt
bilgehankalkan: That returns `SpannedString` and `BodyLineItem` accepts `SpannableStringBuilder`. (`TransactionBodyAdapter` uses `SpannableStringBuilder#clearSpans`)

MiSikora: Ah ok, Never mind then. :)

vbuberen:
I am also against this feature for the reason mentioned already about links being mostly hidden behind some auth or being a signed urls when we are talking about links to some content (images, videos, etc.)

MiSikora:
Thank you for the contribution but after talking about this internally we're going to close this PR due to the reasons already mentioned here.

PR #626, opened 2021-05-19
Gradle to 7.0.2

cortinico:
## :page_facing_up: Context
There has been two point releases for Gradle 7. Let's bump to the latest stable.

## :pencil: Changes
https://docs.gradle.org/7.0.2/release-notes.html


cortinico:
Seems like lint is failing as appcompat is not up-to-date 🤔 Weird that dependabot hasn't suggested it on Monday.

vbuberen (approved):

PR #637, opened 2021-06-26
Update 3.5.0

vbuberen:
## :page_facing_up: Context
This PR is a set of changes required to release a new version of Chucker with support Android 12.
It has not all changes from `develop`, but all non-breaking ones and changes required to support Android 12.

## :pencil: Changes
- #593 PR for compatibility with Android 12
- All dependencies updates
- A few other non-breaking fixes

## :no_entry_sign: Breaking
Nothing breaking expected as I tried to take only parts we need.

No special testing required. Tried to run both on an emulator with API 19 and on one with Android S

## :stopwatch: Next steps
I will create a follow up PRs with updates to changelogs and readme files.

### DIFF
--- a/gradle/gradle-mvn-push.gradle
+++ b/gradle/gradle-mvn-push.gradle
@@ -15,7 +15,7 @@ dokkaJavadoc {
             skipEmptyPackages.set(true)
 
             perPackageOption {
-                prefix.set("com.chuckerteam.chucker.internal")
+                matchingRegex.set(".*\\.internal.*")
                 suppress.set(true)
             }
         }
--- a/gradle/gradle-mvn-push.gradle
+++ b/gradle/gradle-mvn-push.gradle
@@ -95,6 +95,10 @@ afterEvaluate {
                             name = 'Olivier Perez'
                             email = 'olivier@olivierperez.fr'
                         }
+                        developer {
+                            id = 'michalsikora90'
+                            name = 'Michal Sikora'
+                        }
                     }
                 }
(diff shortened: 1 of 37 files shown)


library/build.gradle
```diff
@@ -23,11 +23,16 @@ android {
 
     buildFeatures {
         viewBinding = true
+        buildConfig = false
```
MiSikora, line 26: This is in theory a breaking change but I think we can make it slide. I don't suppose anyone uses `BuildConfig`.

MiSikora (approved):

library/build.gradle
vbuberen: Good catch. I would like to play safe and be sure that we don't introduce anything breaking for 3.x users, so reverted this change.

gradle/gradle-mvn-push.gradle
```diff
@@ -95,6 +95,10 @@ afterEvaluate {
                             name = 'Olivier Perez'
                             email = 'olivier@olivierperez.fr'
                         }
+                        developer {
```
vbuberen, line 98: @MiSikora Found out that we had you missing for quite some time here in POM. Decided to fix it here. Will do a PR for `develop` as well later.
Let me know if this data is fine (took it from one of your repos 😄 ) and if you would like to add your email as well here.

MiSikora: Yup, looks good. Skipping email is fine :).

PR #640, opened 2021-06-28
Update developers list in POM file

vbuberen:
## :page_facing_up: Context
We forgot to add MiSikora as a developer in POM file. This change happened on `3.x` branch already to have him mentioned in 3.5.0 release. Now adding here, so we have him in future releases.

### DIFF
--- a/gradle/gradle-mvn-push.gradle
+++ b/gradle/gradle-mvn-push.gradle
@@ -97,7 +97,7 @@ afterEvaluate {
                         }
                         developer {
                             id = 'michalsikora90'
-                            name = 'Michal Sikora'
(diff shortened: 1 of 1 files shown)


gradle/gradle-mvn-push.gradle
```diff
@@ -95,6 +95,10 @@ afterEvaluate {
                             name = 'Olivier Perez'
                             email = 'olivier@olivierperez.fr'
                         }
+                        developer {
+                            id = 'michalsikora90'
+                            name = 'Michal Sikora'
```
koral--, line 100:
```suggestion
                            name = 'Michał Sikora'
```

PR #641, opened 2021-06-29
Setup automatic repo closure

cortinico:
## :page_facing_up: Context
I've setup a manual GH Actions workflow to trigger the closing/publishing of the sonatype plugin.

## :pencil: Changes
I've used the `io.codearte.nexus-staging` to achieve this. I wanted to use https://github.com/gradle-nexus/publish-plugin/ but just closing the repo was not supported due to https://github.com/gradle-nexus/publish-plugin/issues/19

I've also added the `--no-parallel` flag on the publishing tasks as I noticed that two Staging Repositories gets created whenver a new tag is published. This should mitigate the issue.

I've tested the plugin locally when releasing `3.5.0`. We can try the GH workflow end to end on the next release.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,6 +36,9 @@ buildscript {
         robolectricVersion = '4.5.1'
         truthVersion = '1.1.3'
         vintageJunitVersion = '4.13'
+
+        // Publishing
+        nexusStagingPlugin = '0.30.0'
     }
 
     repositories {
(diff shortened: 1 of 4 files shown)


MiSikora (approved):

build.gradle
```diff
@@ -102,3 +107,9 @@ ext {
     targetSdkVersion = 30
     compileSdkVersion = 30
 }
+
+nexusStaging {
+    username = findProperty("NEXUS_USERNAME")
+    password = findProperty("NEXUS_PASSWORD")
+    stagingProfileId = "ea09119de9f4"
```
vbuberen, line 114: Maybe it would be better if we put this value into env variable as well?

cortinico: We can do that 👍 However the `stagingProfileId` is not a secret (I follow the same approach we use for detekt: https://github.com/detekt/detekt/blob/7f2a4787a9d78ce32b6f2b4a4df2bd2d571ed5ac/buildSrc/src/main/kotlin/packaging.gradle.kts#L17)

Issue #634, opened 2021-06-09
Exception on Android 12 / S


harry248:
## :writing_hand: Describe the bug
On Android 12 chucker causes a crash due to a missing intent flag.

## :bomb: Steps to reproduce
Bump compile/target sdk to Android S and try any app using chucker.

## :wrench: Expected behavior
Chucker doesn't crash the app.

## :iphone: Tech info
 - Device: Pixel 4a
 - OS: Android 12 Beta 1
 - Chucker version: 3.4.0

## :page_facing_up: Additional context
java.lang.IllegalArgumentException: de.ocm.app.dev: Targeting S+ (version 10000 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
    Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
        at android.app.PendingIntent.checkFlags(PendingIntent.java:386)
        at android.app.PendingIntent.getActivityAsUser(PendingIntent.java:470)
        at android.app.PendingIntent.getActivity(PendingIntent.java:456)
        at android.app.PendingIntent.getActivity(PendingIntent.java:420)
        at com.chuckerteam.chucker.internal.support.NotificationHelper$transactionsScreenIntent$2.invoke(SourceFile:47)
        at com.chuckerteam.chucker.internal.support.NotificationHelper$transactionsScreenIntent$2.invoke(SourceFile:19)
        at kotlin.SynchronizedLazyImpl.getValue(SourceFile:74)
        at com.chuckerteam.chucker.internal.support.NotificationHelper.getTransactionsScreenIntent(Unknown Source:2)
        at com.chuckerteam.chucker.internal.support.NotificationHelper.show(SourceFile:101)
        at com.chuckerteam.chucker.api.ChuckerCollector.onRequestSent$com_github_ChuckerTeam_Chucker_library(SourceFile:65)
        at com.chuckerteam.chucker.api.ChuckerInterceptor.intercept(SourceFile:111)

vbuberen:
Hey @harry248 
Thanks for reporting this issue. We already have it resolved in our snapshot version: https://github.com/ChuckerTeam/chucker#snapshots-
Please give it a try and let me know if it works fine. 
We will try to cut a new `3.4.1` stable release for Android 12 users in nearest time. Need just to pick non-breaking commits there.

[...]

PR #653, opened 2021-07-16
Correctly set `FLAG_IMMUTABLE` for error screen `PendingIntent`

cortinico:
## :page_facing_up: Context
Originally reported here https://github.com/ChuckerTeam/chucker/issues/634#issuecomment-880196269
When we backported this PR #593, the `PendingIntent` for the error screen slipped through. Therefore users

## :pencil: Changes
Just adding the `FLAG_IMMUTABLE` to the only `PendingIntent` that was missing it.

## :paperclip: Related PR
#593 and #637

I haven't tested this, it's a best guess fix.

## :stopwatch: Next steps
We should potentially release `3.5.1` with this fix included. Wdyt @vbuberen @MiSikora


vbuberen:
Right, we forgot to add flag for this intent because in develop we already removed errors tab and I missed it during cherrypicking. 
We would need a hotfix release.

vbuberen (approved):

MiSikora (approved):

PR #665, opened 2021-07-26
Add groups

israNascimento:
## :camera: Screenshots

https://user-images.githubusercontent.com/9087318/126991171-99b6a1d4-e04a-4401-9729-e7e9ef76e6e5.mov

## :page_facing_up: Context
Adding possibility to group requests by something useful for the context of an app. The users could do something like group the requests used by some feature.

## :pencil: Changes
Update DAO to filter the requests by multiples URLS.
Update collector to accept a list of Groups

## :no_entry_sign: Breaking
No

Follow the steps on the video. I added to the sample app the Groups "Delay" and "Cookies"

## :stopwatch: Next steps
I don't know, maybe a possibility to also group by the request method

I need some help to think about the comment on the ViewModel, you'll know about what I'm talking about once you read :P


MiSikora:
This is such a big and invasive PR that I don't really know how to approach it. I think it should be discussed first whether we want this feature before merging any code.

I won't make a full review now, but from my point of view the code cannot be merged as is. There are issues like global state, random UUIDs, strigly–typed queries or mock based test that don't make me feel confident that it would be maintainable part of code.

PR #678, opened 2021-08-13
Enhancement - Repeat a request from chucker's TransactionActivity

chaits98:
## :camera: Screenshots

Screenshot_20210813_130454
Screenshot_20210813_130503
Screenshot_20210813_130528
Screenshot_20210813_130522

## :page_facing_up: Context

This feature was an enhancement which I thought would help developers debug specific APIs for intricate flows which require multiple API calls
Updated CHANGELOG.md

## :pencil: Changes

I have changed Transaction detail view and added a menu item for repeat request

## :paperclip: Related PR

NA
## :no_entry_sign: Breaking

NA

No special case

## :stopwatch: Next steps

No next steps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Please add your entries according to this format.
 ### Added
 * Decoding of request and response bodies can now be customized. In order to do this a `BodyDecoder` interface needs to be implemented and installed in the `ChuckerInterceptor` via `ChuckerInterceptor.addBinaryDecoder(decoder)` method. Decoded bodies are then displayed in the Chucker UI.
 * Create dynamic shortcut when `ChuckerInterceptor` added. Users can opt out of this feature using `createShortcut(false)` in `ChuckerInterceptor.Builder`
+* Repeat request action menu item which let's you repeat any API call made previously from the application to which chucker is attached.
 
 ### Fixed
 
--- a/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/java/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -3,6 +3,7 @@ package com.chuckerteam.chucker.api
 import android.content.Context
 import androidx.annotation.VisibleForTesting
 import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
+import com.chuckerteam.chucker.internal.support.*
 import com.chuckerteam.chucker.internal.support.CacheDirectoryProvider
 import com.chuckerteam.chucker.internal.support.PlainTextDecoder
 import com.chuckerteam.chucker.internal.support.RequestProcessor
@@ -36,19 +37,21 @@ public class ChuckerInterceptor private constructor(
 
     private val collector = builder.collector ?: ChuckerCollector(builder.context)
 
+    // Empty set is added as headers to redact to save headers in db for data repeating API calls
     private val requestProcessor = RequestProcessor(
         builder.context,
         collector,
         builder.maxContentLength,
-        headersToRedact,
+        emptySet(),
         decoders,
     )
 
+    // Empty set is added as headers to redact to save headers in db for data repeating API calls
     private val responseProcessor = ResponseProcessor(
         collector,
         builder.cacheDirectoryProvider ?: CacheDirectoryProvider { builder.context.filesDir },
         builder.maxContentLength,
-        headersToRedact,
+        emptySet(),
         builder.alwaysReadResponseBody,
         decoders,
     )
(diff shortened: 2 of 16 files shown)


MiSikora:
While I like the idea, the code has issues that would need to be resolved first:
- No tests.
- Passing context to the presentation layer.
- Efficiency issues where every retry creates a new OkHttp instance, and it does it on the main thread.
- Interceptor used for the repeated requests does not respect user's settings like redacted headers or retention period.
- Possibility to repeat redirect requests.
- Requests with arbitrary bodies can't be handled properly. Anything that is binary–encoded is basically out of our reach. We are also limited by the body payload size.
- Some headers might be missing depending on the configuration of user's OkHttp client or due to redacting.

chaits98:
@MiSikora I appreciate your inputs on my approach.

I have a few things I want to discuss.

1. Interceptor used for the repeated requests does not respect user's settings like **_redacted headers_** or **_retention period_**.

    1. **_redacted headers_** - Elucidated along point 3
    2. **_retention period_** - I do not see a way around this point unless we store user's preference of retention period for the running session.

2. Requests with **_arbitrary bodies_** can't be handled properly. Compression can be problematic. We are also limited by the **_body payload size_**.
    1. **_arbitrary bodies_** - The purpose of this feature is to be able to resend the exact same request again to a user's server. Any change in the body should come from the application to which the chucker is attached.
    2. **_body payload size_** - What is the body size limit, how is large sized request body stored / shown to user in case such a request is made.

3. Some headers might be missing depending on the configuration of user's OkHttp client or due to redacting.
    1. Since any redacted headers are not available at the time of viewing the transactions, the contract of redacting headers is maintained in repetition of request made.
    2. Assuming any redacted headers are essential to the user's API call, I see two options available as follows:

        - The user of chucker library should not redact such headers in case this feature is to be utilised.
        - The user's choice of redacting any headers is maintained only on UI and not in the stored data which would essentoially break the contract between the library and it's user.


Please provide you input on this and correct me if I am wrong here.

MiSikora:
> @MiSikora I appreciate your inputs on my approach.
>    2. **_retention period_** - I do not see a way around this point unless we store user's preference of retention period for the running session.

Yes that's the point. Since interception creation has a single entry point in the builder it could be done there.

>    1. **_arbitrary bodies_** - The purpose of this feature is to be able to resend the exact same request again to a user's server. Any change in the body should come from the application to which the chucker is attached.

We might not know how the bodies are shaped. Depending on the order of interceptors and whether it is an application or a network interceptor we might not get the full picture of what is going on with the body. For example compression other than gzip which we don't know how to reapply or any type of customisation.

>    2. **_body payload size_** - What is the body size limit, how is large sized request body stored / shown to user in case such a request is made.

It is customisable in the interceptor. If the body reaches acceptable limit then it is truncated and we don't have all bytes. To simplify this, I think in the first implementation it would be the best to allow only for repetition of `GET` type requests.

>    1. Since any redacted headers are not available at the time of viewing the transactions, the contract of redacting headers is maintained in repetition of request made.

Yes, redacted headers still should be redacted. But it should be handled by reusing user's configuration.

>    2. Assuming any redacted headers are essential to the user's API call, I see two options available as follows:
>       
>       * The user of chucker library should not redact such headers in case this feature is to be utilised.
>       * The user's choice of redacting any headers is maintained only on UI and not in the stored data which would essentoially break the contract between the library and it's user.

IMO the second choice is preferable. Redacting is important only on the UI layer and during the export anyway. If some malicious agent can inspect app's storage then it has to be capable of breaking OS sandbox at which point they can do whatever they want including inspecting HTTP traffic on their own.

However, there are still issues where we might not have all of the information to successfully retry requests. The most basic case are authorisation headers. I often do it with an additional interceptor that is applied as a network interceptor. We should advise the users that if they want to properly utilise repetition feature then need to apply `ChuckerInterceptor` as the last interceptor in the chain.

Before you proceed with anything I think it would be best to wait for a feedback on this feature from other maintainers.

chaits98:
@MiSikora sure.
I'll halt any further development until other maintainers provide a feedback

cortinico:
> I'll halt any further development until other maintainers provide a feedback

Sorry for the late reply.

Thanks for taking the time to send this PR @chaits98 🙏 

I'm a bit unsure about the cost/benefit of adding this feature. I believe it could be valuable for some of our users (as you mentioned, those following intricate flows). At the same time, I believe those users are not a significant portion of our user base.

If you need to re-run a request, the likelihood of you having to edit some fields as well (a nonce or a header) is quite high. Chucker is not offering anything like that. So those users are most likely using Charles or other more advanced proxies.

The rationale behind Chucker is to give developers a quick snapshot of which network requests happened as the app runs. We never intended to fully replace standalone products like Charles or similar, and I believe we should refrain from adding features that we can't be sure are valuable for the majority of our users.

vbuberen:
I don't think Chucker needs this functionality at this point, since it removes focus from the primary idea of Chucker being an inspector.

chaits98:
@vbuberen @MiSikora @cortinico @technoir42 

Thank you so much for taking time to review this PR.
Since it is decided that this feature will not be added to chucker, I will close this PR.

I wanted to use this feature for myself, so I will go ahead and do that and it is available for anyone to use on my fork of chucker.

Keep up the good work guys, your project really helps android developers. Kudos.

PR #696, opened 2021-09-24
Add support for HTTP Archive (HAR)

### LINKED ISSUE #401: Add support for exporting requests to HTTP Archive format (.har)
I'm not fixed on `.har`, but it's a relatively simple plain text format, that's well supported by many applications (including Charles Proxy).

We've run into many issues where people running our QA app are either not allowed, or not capable of running Charles Proxy and having the ability to view these requests there after would be a huge help to our debugging efforts.

My goal is to dump the entire `transactions` table into a single file that I can view in Charles Proxy.

I've done some initial digging, and I think the database contains all the information I would need in order to pull this off.

- [x] Yes
- [ ] No

I'm willing to do the work for this, but it likely won't happen right away.

I'm also not convinced it's necessary to do this in Chucker core. I could fairly easily just look at the database in application code or a separate library that knows about Chuckers internal database and write the export code there. That being said, the reason I'm posting this here; I think this would be a feature that might be useful to other Churcker users as well.

gusdantas:
## :camera: Screenshots
WhatsApp Image 2021-09-24 at 13 09 12 WhatsApp Image 2021-09-24 at 13 09 39 WhatsApp Image 2021-09-24 at 13 10 06
WhatsApp Image 2021-09-24 at 13 11 13 WhatsApp Image 2021-09-24 at 13 11 38

## :page_facing_up: Context
This updates @JayNewstrom 's [solution](https://github.com/ChuckerTeam/chucker/pull/417) which allows Chucker to share/export files with desktop http viewers such as Charles Proxy, Chrome Network Debugger, etc.

Closes #401

## :pencil: Changes
Most of the code is new, but some changes were made on toolbars export buttons to allow har export.

## :no_entry_sign: Breaking
No breaking changes.

@JayNewstrom 's [solution](https://github.com/ChuckerTeam/chucker/pull/417) has some automated tests. You can also test via the UI (see screenshots). After you have the file ready to import into another application, it should be able to be imported.

## :stopwatch: Next steps
As told by @JayNewstrom:
```
This handles the simple case, and allows you to view request/response in Charles. The generated file is likely not 100% accurate/to spec yet. See comments in #401 for some reasons why (ie. missing data).

I'd like to ensure this is the right direction, and get bugs/examples that fail to import and fix them as we find them.

I'd also like to add support for binary request/response bodies via base 64 encoding.
```


MiSikora:
Sorry for the radio silence. I'm currently occupied with life stuff. I should have more time soonish.

gusdantas:
@MiSikora I added a test on `ChuckerInterceptorTest` but I couldn't find a way to get the request when the `ClientFactory` is `NETWORK`. Is there a correct way to do it?

MiSikora:
> Is there a correct way to do it?

You should be able to take it from the server. Something like `server.takeRequest().headers.byteCount()`.

gusdantas:
> You should be able to take it from the server. Something like `server.takeRequest().headers.byteCount()`. Or `response.networkResponse!!.request.headers.byteCount()`.

@MiSikora thank you, now the test passes on both `ClientFactory`!

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -562,6 +562,34 @@ internal class ChuckerInterceptorTest {
...
+        val response = call.execute()
+
+        val transaction = chuckerInterceptor.expectTransaction()
+        if (factory == ClientFactory.APPLICATION) {
+            assertThat(transaction.requestHeadersSize).isEqualTo(response.request.headers.byteCount())
+            assertThat(transaction.responseHeadersSize).isEqualTo(response.headers.byteCount())
+        } else {
+            assertThat(transaction.requestHeadersSize).isEqualTo(server.takeRequest().headers.byteCount())
+            assertThat(transaction.responseHeadersSize).isEqualTo(response.networkResponse!!.headers.byteCount())
+        }
```
MiSikora, line 590:
```suggestion
        val expectedResponse = if (factory == ClientFactory.APPLICATION) response else response.networkResponse!!
        assertThat(transaction.requestHeadersSize).isEqualTo(expectedResponse.request.headers.byteCount())
        assertThat(transaction.responseHeadersSize).isEqualTo(expectedResponse.headers.byteCount())
```

MiSikora:
Left 1 NIT. I'll do tests tomorrow to see how it behaves, and hopefully, we're good to go. 👍

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -562,6 +562,38 @@ internal class ChuckerInterceptorTest {
...
+        val response = call.execute()
+
+        val (originalRequestHeadersSize, originalResponseHeadersSize) = when (factory) {
+            ClientFactory.APPLICATION -> {
+                Pair(response.request.headers.byteCount(), response.headers.byteCount())
+            }
+            ClientFactory.NETWORK -> {
+                Pair(server.takeRequest().headers.byteCount(), response.networkResponse!!.headers.byteCount())
+            }
+        }
```
MiSikora, line 590: I do insist that you use the suggested snippet for this. However, using `when` instead of `if/else` is probably better. This change introduces unnecessary duplication by extracting requests, headers and taking data from different places, blurring the whole picture.

MiSikora (approved):
Ok, I did some tests using Chucker both as an application interceptor and a network interceptor. For reading HAR files, I used this application - https://toolbox.googleapps.com/apps/har_analyzer.

Thank you for this contribution! I didn't find any issues, and everything worked as expected. After changing one last NIT and after @cortinico gives their approval, I think we can merge it.

library/src/test/java/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
gusdantas: I changed a little but the problem of taking data from different sources is mantained... both request and response are changed depending of the `ClientFactory`...

MiSikora: Code below passes the test. What is the problem?

```kotlin
val chuckerInterceptor = ChuckerInterceptorDelegate(
    maxContentLength = SEGMENT_SIZE,
    headersToRedact = setOf("Header-To-Redact"),
    cacheDirectoryProvider = { tempDir },
)
val client = factory.create(chuckerInterceptor)

val request = Request.Builder().url(serverUrl).header("Header-To-Redact", "Hello").build()
val call = client.newCall(request)

server.enqueue(
    MockResponse().addHeader("Header-To-Redact", "Goodbye").setResponseCode(HTTP_NO_CONTENT)
)
val response = call.execute()

val transaction = chuckerInterceptor.expectTransaction()
val expectedResponse = when (factory) {
    ClientFactory.APPLICATION -> response
    ClientFactory.NETWORK -> response.networkResponse!!
}
assertThat(transaction.requestHeadersSize).isEqualTo(expectedResponse.request.headers.byteCount())
assertThat(transaction.responseHeadersSize).isEqualTo(expectedResponse.headers.byteCount())
```

gusdantas: Oh, now it worked! I probably did something wrong when I first tried it. Thanks!

cortinico (approved):

cortinico:
Great work 🚀 Thanks for tackling this

gusdantas:
@cortinico @MiSikora @vbuberen I do appreciate all the compliments, but the main effort was made by @JayNewstrom !
Thank you!

PR #733, opened 2021-12-04
Fix RTL issue in payload

mreram:
All TextView(s) must be LTR because in RTL language the payload will be affected and we have some problems

## :camera: Screenshots
Screen Shot 2021-12-04 at 11 02 34 AM

## :page_facing_up: Context
When our app is RTL mode we have some problems with payload texts.

## :pencil: Changes
I added force LTR text direction in payload items.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Please add your entries according to this format.
 * Fixed request headers not being redacted in case of failures [#545].
 * Fixed wrongful processing of one shot and duplex requests [#544].
 * Fixed writing to database on the main thread [#487].
+* Fixed RTL issue in payload view
 
 ### Removed
 
--- a/library/src/main/res/layout/chucker_transaction_item_body_line.xml
+++ b/library/src/main/res/layout/chucker_transaction_item_body_line.xml
@@ -5,6 +5,7 @@
     android:layout_height="wrap_content"
     android:textIsSelectable="true"
     android:minLines="1"
+    android:textDirection="ltr"
     android:gravity="start"
     android:layout_marginHorizontal="@dimen/chucker_doub_grid"
     android:typeface="monospace" />


vbuberen:
Thanks for you contribution!
The screenshot you provided is how the payload is supposed to work for RTL locales?

mreram:
Thanks for your consideration.
This screenshot is before fixing the issue.
After this change:
Screen Shot 2021-12-04 at 5 46 44 PM

Drjacky:
The line `"title": برای شما` is correct but, it'd be better to keep the whole alignment the same as before(LTR); Cause now it's weird, hard to read and RTL developers have used to read from LTR.

MiSikora:
I think JSON preview is for searching, inspection, and visual validation, not for reading a text like news or messages. I prefer the layout after changes in this PR. But I'm not a native RTL user. Maybe we could make a poll to get feedback from our community?

vbuberen:
@mreram Thanks for the `after` screenshot.

@Drjacky Thanks for your input here.

@MiSikora Had the same idea yesterday, so tweeted a question about the change: https://twitter.com/vbuberen/status/1467238705081503749?t=nT30nBL-UkyWXwOKUSUD9A&s=09
So far received just 2 replies.

Drjacky:
@mreram Would you please try to set `textAlignment="gravity"` or `textAlignment="viewStart"` to check if you could have the whole text LTR and have the fix on line `"title":برای شما`.

mreram:
I assume in json format or other programming languages it's rare to use other languages as a keyword.
Also, in key-value json every Farsi characters are 2bytes and it's not good to use other languages in json format. Thanks for your suggestion @Drjacky I will test it.

mreram:
@Drjacky with textAlignment="gravity"
Screen Shot 2021-12-05 at 10 00 52 PM

And textAlignment="viewStart" is same as before.
I think the only one that works is textAlignment="ltr" and it will be fixed in most cases. Anyway, I've created a layout file with the same name and I'm using LTR in my app to fix this issue.

cortinico:
> I prefer the layout after changes in this PR

Same here. Also not a native RTL so hard to make a call. I believe some RTL users are coding ltr (at least the one I know personally) so I'm assuming ltr is also fine for visualizing payloads like JSON/XML and others. +1 for merging this

vbuberen:
Ok, if everyone agree here let's merge the fix.
Let's see if we get feedback afterwards. 

@mreram Thanks again for your contribution and making Chucker a better tool for everyone 🚀

vbuberen (approved):

PR #751, opened 2021-12-25
Switch source directory from java to kotlin

vbuberen:
## :page_facing_up: Context
We don't have Java in the library for quite some time and since nowadays we can more or less safely use `kotlin` folder as a source set I did the change to get rid of `java` folder.

## :pencil: Changes
- Change source sets to use `kotlin` folder instead of `java`
- Fixed lint warning about ViewBinding property

## :no_entry_sign: Breaking
Nothing breaking expected. `apiCheck` task worked fine locally for me after the change.

Chucker and its tests should work as before without any issues.


library-no-op/build.gradle
```diff
@@ -11,6 +11,10 @@ android {
         ]
     }
 
+    sourceSets.all {
+        kotlin.srcDir("src/$name/kotlin")
+    }
```
MiSikora, line 16: Is it necessary? If I recall correctly they recently changed AGP to automatically include `kotlin` source sets.

MiSikora (approved):

library-no-op/build.gradle
vbuberen: I also remember that AGP should handle that, but didn't check. Was keeping in mind that back in spring had to add such construction.

vbuberen: Will check if it works without and update PR, thanks.

vbuberen: Yeah, looks like everything works without adding `sourceSets` lines in `build.gradle` files. Updated.

Issue #783, opened 2022-03-11
Add function to generate a file of transactions (without immediately exporting)


jd565:
## :warning: Is your feature request related to a problem? Please describe
We currently use chucker in our app and find it is a really helpful tool for debugging issues. We are currently working on automating our internal bug reporting tooling for our QA team. The basic flow is they can shake the device and write the bug report and it will appear in asana/jira/whatever.

What we would like to do is when they submit a bug report automatically get some logs from chucker and upload that file with the bug report, so we can look through the network calls if we think that will help debugging the issue.

## :bulb: Describe the solution you'd like
Right now chucker has all the logic available to write a file with all this information in. What I would like is if there was a function that we could call to create the file - either then telling us where this file is, or allowing us to pass in a file/path that we want chucker to write to.

## :bar_chart: Describe alternatives you've considered
Without this we would resort to writing a manually SQL query and copying code out of chucker to implement the same thing.

## :page_facing_up: Additional context
I think it would be useful to provide some parameters to allow you to specify a certain timeframe or number of transactions (as getting all the chucker transactions can take an awful long time), so something like

`suspend fun writeChuckerLogs(maxTransactions: Long?, timeframe: Duration?, outputFile: File?): File`

## :raising_hand: Do you want to develop this feature yourself?
- [x] Yes
- [ ] No
Would be happy to develop feature, but also happy if anyone else wants to give it a go

cortinico:
> What I would like is if there was a function that we could call to create the file - either then telling us where this file is, or allowing us to pass in a file/path that we want chucker to write to.

It sounds reasonable to me to add this functionality. Feel free to send a PR 👍

jd565:
@cortinico PR above 😄

cortinico:
> @cortinico PR above 😄

Thanks for doing it 🙏 I'll try to review it in the next days or within the weekend

jd565:
This has been implemented and merged now

PR #784, opened 2022-03-12
Add export function to API

jd565:
Add an export function that writes transaction data to a file and
returns a URI to access that file. The output format is the same as when
exporting from the chucker UI, but can be triggered from code.

## :camera: Screenshots
All internal changes - nothing different in UI

## :page_facing_up: Context
Adds functionality to export transactions to a file on the API. [Related issue](https://github.com/ChuckerTeam/chucker/issues/783)

## :pencil: Changes
Updated the Chucker API object to add in the new functionality. I've not made this a suspending function to maintain interop with java, so have just left a comment that this needs to be called on a background thread.

I added a button to the sample app (I had to comment out the strict mode checking otherwise the sample app closed immediately) which calls the new method - I'm not sure if this should be left in the sample long term so happy to remove it if it doesn't make sense.

## :no_entry_sign: Breaking

No

Added a button to the sample app to export to a file then pulled the file and verified the contents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Please add your entries according to this format.
 * Brotli compression support
 * Added ability to export list of transactions as .har file.
 * Added ability to save single transaction as .har file.
+* Added ability to export transactions to a file triggered in code.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Please add your entries according to this format.
 * Brotli compression support
 * Added ability to export list of transactions as .har file.
 * Added ability to save single transaction as .har file.
-* Added ability to export transactions to a file triggered in code.
+* Added ability to export transactions to a file programmatically.
 
 ### Fixed
 
--- a/library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
@@ -2,6 +2,9 @@ package com.chuckerteam.chucker.api
 
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
+import android.widget.Toast
+import com.chuckerteam.chucker.R
 
 /**
  * No-op implementation.
@@ -19,4 +22,11 @@ public object Chucker {
     public fun dismissNotifications(context: Context) {
         // Empty method for the library-no-op artifact
     }
+
+    @JvmStatic
+    public fun writeTransactions(
+        context: Context,
+        maxTransactions: Long?,
+        startTimestamp: Long?,
+    ): Uri? = null
 }
--- a/library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
@@ -2,9 +2,6 @@ package com.chuckerteam.chucker.api
 
 import android.content.Context
 import android.content.Intent
-import android.net.Uri
-import android.widget.Toast
-import com.chuckerteam.chucker.R
 
 /**
  * No-op implementation.
--- a/library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
@@ -22,11 +22,4 @@ public object Chucker {
     public fun dismissNotifications(context: Context) {
         // Empty method for the library-no-op artifact
     }
-
-    @JvmStatic
-    public fun writeTransactions(
-        context: Context,
-        maxTransactions: Long?,
-        startTimestamp: Long?,
-    ): Uri? = null
 }
--- a/library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
@@ -81,43 +81,6 @@ public object Chucker {
         NotificationHelper(context).dismissNotifications()
     }
 
-    /**
-     * Export the chucker transactions to a file.
-     *
-     * This must be called on a background thread.
-     *
-     * @param context Application context
-     * @param maxTransactions Maximum number of transactions to return in the file. Passing null
-     * means no limit on the number of transactions
-     * @param startTimestamp The timestamp to read transactions from. Passing null means
-     * transactions will not be limited by timestamp
-     * @return The content uri of a file with the transactions in
-     */
-    @JvmStatic
-    public fun writeTransactions(
-        context: Context,
-        maxTransactions: Long?,
-        startTimestamp: Long?,
-    ): Uri? {
-        RepositoryProvider.initialize(context)
-        val transactions =
-            RepositoryProvider.transaction().getTransactions(maxTransactions, startTimestamp)
-        if (transactions.isEmpty()) {
-            Toast
-                .makeText(context, R.string.chucker_export_empty_text, Toast.LENGTH_SHORT)
-                .show()
-            return null
-        }
-
-        val sharableTransactions = TransactionListDetailsSharable(transactions, encodeUrls = false)
-        return runBlocking {
-            sharableTransactions.writeToFile(
-                context = context,
-                fileName = "api_transactions.txt",
-            )
-        }
-    }
-
     internal var logger: Logger = object : Logger {
         val TAG = "Chucker"
 
--- a/library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
@@ -5,13 +5,19 @@ import android.content.Intent
 import android.content.pm.ShortcutInfo
 import android.content.pm.ShortcutManager
 import android.graphics.drawable.Icon
+import android.net.Uri
 import android.os.Build
 import android.util.Log
+import android.widget.Toast
 import androidx.core.content.getSystemService
 import com.chuckerteam.chucker.R
+import com.chuckerteam.chucker.internal.data.repository.RepositoryProvider
 import com.chuckerteam.chucker.internal.support.Logger
 import com.chuckerteam.chucker.internal.support.NotificationHelper
+import com.chuckerteam.chucker.internal.support.TransactionListDetailsSharable
+import com.chuckerteam.chucker.internal.support.writeToFile
 import com.chuckerteam.chucker.internal.ui.MainActivity
+import kotlinx.coroutines.runBlocking
 
 /**
  * Chucker methods and utilities to interact with the library.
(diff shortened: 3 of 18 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Please add your entries according to this format.
 * Brotli compression support
 * Added ability to export list of transactions as .har file.
 * Added ability to save single transaction as .har file.
+* Added ability to export transactions to a file triggered in code.
```
cortinico, line 14:
```suggestion
* Added ability to export transactions to a file programmatically.
```

library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -19,4 +22,11 @@ public object Chucker {
     public fun dismissNotifications(context: Context) {
         // Empty method for the library-no-op artifact
     }
+
+    @JvmStatic
+    public fun writeTransactions(
+        context: Context,
+        maxTransactions: Long?,
+        startTimestamp: Long?,
+    ): Uri? = null
```
cortinico, line 31: I think this should be moved to `ChuckerCollector`

library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -75,6 +81,43 @@ public object Chucker {
...
+     * transactions will not be limited by timestamp
+     * @return The content uri of a file with the transactions in
+     */
+    @JvmStatic
+    public fun writeTransactions(
+        context: Context,
+        maxTransactions: Long?,
+        startTimestamp: Long?,
+    ): Uri? {
+        RepositoryProvider.initialize(context)
```
cortinico, line 102: If you move this inside the Collector, this won't be needed.

library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -75,6 +81,43 @@ public object Chucker {
...
+        val transactions =
+            RepositoryProvider.transaction().getTransactions(maxTransactions, startTimestamp)
+        if (transactions.isEmpty()) {
+            Toast
+                .makeText(context, R.string.chucker_export_empty_text, Toast.LENGTH_SHORT)
+                .show()
+            return null
+        }
```
cortinico, line 110: This method should not show UI operations. It's up to the consumer of this method to do so if they wish (say if the URI returned is null).

library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -75,6 +81,43 @@ public object Chucker {
...
+
+        val sharableTransactions = TransactionListDetailsSharable(transactions, encodeUrls = false)
+        return runBlocking {
```
cortinico, line 113: This is not great as it might break the UI thread. Potentially returning a suspending function here would be better

sample/src/main/kotlin/com/chuckerteam/chucker/sample/MainActivity.kt
```diff
@@ -38,6 +38,13 @@ class MainActivity : AppCompatActivity() {
             launchChuckerDirectly.visibility = if (Chucker.isOp) View.VISIBLE else View.GONE
             launchChuckerDirectly.setOnClickListener { launchChuckerDirectly() }
 
+            exportToFile?.visibility = if (Chucker.isOp) View.VISIBLE else View.GONE
+            exportToFile?.setOnClickListener {
+                Thread {
+                    Chucker.writeTransactions(this@MainActivity, 1, null)
+                }.start()
```
cortinico, line 45: Please don't start a thread explicity here. We use coroutines to do any async work.

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
```diff
@@ -42,4 +42,13 @@ internal class HttpTransactionDatabaseRepository(private val database: ChuckerDa
     }
 
     override suspend fun getAllTransactions(): List<HttpTransaction> = transactionDao.getAll()
+
+    override fun getTransactions(
```
cortinico, line 46:
```suggestion
    override fun getTransactionsInTimeRange(
```

or somethign more meaningful

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/repository/HttpTransactionRepository.kt
```diff
@@ -26,4 +26,6 @@ internal interface HttpTransactionRepository {
     fun getTransaction(transactionId: Long): LiveData<HttpTransaction?>
 
     suspend fun getAllTransactions(): List<HttpTransaction>
+
+    fun getTransactions(maxTransactions: Long?, minTimestamp: Long?): List<HttpTransaction>
```
cortinico, line 30: Same as above `getTransactionsInTimeRange`

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/room/HttpTransactionDao.kt
```diff
@@ -44,4 +44,7 @@ internal interface HttpTransactionDao {
 
     @Query("SELECT * FROM transactions")
     suspend fun getAll(): List<HttpTransaction>
+
+    @Query("SELECT * FROM transactions WHERE requestDate >= :timestamp LIMIT :limit")
+    fun getTransactions(limit: Long, timestamp: Long): List<HttpTransaction>
```
cortinico, line 49: Same as above: `getTransactionsInTimeRange`

library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -75,6 +81,43 @@ public object Chucker {
...
     }
 
+    /**
+     * Export the chucker transactions to a file.
+     *
+     * This must be called on a background thread.
+     *
+     * @param context Application context
+     * @param maxTransactions Maximum number of transactions to return in the file. Passing null
+     * means no limit on the number of transactions
```
cortinico, line 91: Why would want to limit this? I don't fully understand the use case.

cortinico (commented):
Thanks for the PR. Left several comments that needs to be addressed before we can proceed

cortinico:
> I've not made this a suspending function to maintain interop with java, so have just left a comment that this needs to be called on a background thread.

Just noticed this. I would say that we can safely make this a suspending function at this stage.

Moreover, tests are missing. Could you add some of them?

library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
jd565: I have updated this to be inside ChuckerCollector, but it seems slightly odd to me - (as a consumer) why do I have to instantiate a class that is about handling sending/receiving http messages in order to dump the saved transactions to a file?

sample/src/main/kotlin/com/chuckerteam/chucker/sample/MainActivity.kt
jd565: Have updated to use coroutines, but have had to add coroutines and activity-ktx dependencies to the sample gradle project as they weren't there before

library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
jd565: I'm not sure I have a specific use case but it seems like a good idea to have a limit call on an API like this.

Dumping the whole list of transactions from the database can take considerable time, so having ways to limit it and make this API a bit quicker would be good imo.

With the API like this users can specify either a time limit (e.g. calculate the last 1 hour) or can set a count limit (e.g. show the last 100 HTTP messages)

jd565:
> 

I think I have covered all your comments. Could you give some more information about what functions you are expecting to be tested?

The old `Sharable.shareAsFile` function has no tests on it, so I am not sure if you want that to be tested? The overall function aswell in ChuckerCollector I am not sure if you want a test on but I am not 100% sure how to achieve this without mocking everything and leaving a not very useful test.

library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
cortinico: The documentation says:

```
 * The collector responsible of collecting data from a [ChuckerInterceptor] and
 * storing it/displaying push notification. 
```

While the `Chucker` class you used originally is just the entry point of our library. As you want to do an advanced operation, I think it's reasonable to ask the user to instanciate and extra object. This will not pollute the `Chucker` interface making it a god object in the long run.

sample/src/main/kotlin/com/chuckerteam/chucker/sample/MainActivity.kt
cortinico: Yeah I realized this is a side effect 🤔 I'll be fine with it. I'm unsure what's the other maintainers take on this (specifically @MiSikora )

library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
cortinico: I would not add them, till we have a valid use case. We can still add them afterwards if needed. Having the method as suspending allows it to run on a separate thread, so having it as a long operation is not a major deal imho.

Between the two parameters, the timestamp seems more reasonable. The limit instead, will be useless if not paired with a proper pagination support (as you'll be essentially losing some transactions).

jd565:
@cortinico Thanks for looking over it. I think I have addressed all the comments above, but I'm not 100% sure about the testing side of it. Could you clarify what parts you think should have unit test coverage and I can look into it?

jd565:
@cortinico Can you give any advice about the above?

cortinico (approved):

library-no-op/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -2,6 +2,9 @@ package com.chuckerteam.chucker.api
 
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
+import android.widget.Toast
+import com.chuckerteam.chucker.R
```
vbuberen, line 7: Do we need these new imports here?

library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -5,13 +5,19 @@ import android.content.Intent
 import android.content.pm.ShortcutInfo
 import android.content.pm.ShortcutManager
 import android.graphics.drawable.Icon
+import android.net.Uri
 import android.os.Build
 import android.util.Log
+import android.widget.Toast
 import androidx.core.content.getSystemService
 import com.chuckerteam.chucker.R
+import com.chuckerteam.chucker.internal.data.repository.RepositoryProvider
```
vbuberen, line 14: I don't think we need these imports as well. Feels like some leftovers from experimenting, etc.

vbuberen (commented):
The approach looks fine to me.

Have left a few comments about strange new imports.

Also, I feel that we need to mention this feature in our Readme file as well.

BluestormDNA:
@jd565 @cortinico cound it be possible to add HAR file export too?

cortinico:
> @cortinico I also forgot about this :/ Your changes look fine and if you are happy to merge then I'm happy to.

Yup I'm merging this one 👍 Thank you again @jd565 

> @jd565 @cortinico cound it be possible to add HAR file export too?

@BluestormDNA please open a separate issue

PR #750, opened 2021-12-25
Fix strict mode threading issues

vbuberen:
## :page_facing_up: Context
This is a fix for #737. Seems like the issue was in accessing `SharedPreferences` from the main thread. We were not catching it earlier due to having not all `StrictMode` options enabled.
Added additional `StrictMode` checks to the sample app. Additionally saw that we still have a piece of code for removing Chuck database file. I suppose we don't need it anymore, so removed it.

## :pencil: Changes
- Added more `StrictMode` checks to the sample app.
- Add dispatcher switching for `doMaintanence()` function to move from the main thread
- Made prefs initialisation lazy, since it was also making StrictMode not happy before
- Removed the code which tried to remove Chuck database
- Updated Gradle wrapper to 7.3.2

## :no_entry_sign: Breaking
No breaking changes introduced

## :hammer_and_wrench: How to test
Just run the sample app and see that no issues found with more `StrictMode` options enabled.

## :stopwatch: Next steps
Closes #737

PR #800, opened 2022-04-18
[feature/chucker-gql] GQL support against latest develop 

alkber:
There was already an old implementation done [here ](https://github.com/ChuckerTeam/chucker/pull/70). This PR seems to be old one (2019) and way out of sync with develop branch. I have refactored it to meet the latest changes. 

This almost getting ready and not the final PR yet, need to get valuable input on the code quality and optimization if required. 

Sample Activity has also been provided for the demo

**Known Issues**  
Couldn't successfully run the sample until i commented this part of the code 

```
//        StrictMode.setVmPolicy(
//            StrictMode.VmPolicy.Builder()
//                .detectLeakedClosableObjects()
//                .penaltyLog()
//                .penaltyDeath()
//                .build()
//        )
//
//        StrictMode.setThreadPolicy(
//            StrictMode.ThreadPolicy.Builder()
//                .detectDiskReads()
//                .detectDiskWrites()
//                .penaltyLog()
//                .penaltyDeath()
//                .build()
//        )
```


sample/src/main/kotlin/com/chuckerteam/chucker/sample/GqlHttpTask.kt
```diff
@@ -0,0 +1,56 @@
...
+import retrofit2.converter.gson.GsonConverterFactory
+import retrofit2.http.Body
+import retrofit2.http.Headers
+import retrofit2.http.POST
+
+class GqlHttpTask(
+    client: OkHttpClient,
+) : HttpTask {
+
+    private val api = Retrofit.Builder()
```
MiSikora, line 14: I think that the vast majority of people will use GraphQL with something like Apollo. I would prefer to see a real-life scenario integration within our sample.

library/src/main/kotlin/com/chuckerteam/chucker/internal/support/RequestProcessor.kt
```diff
@@ -75,4 +82,15 @@ internal class RequestProcessor(
                 null
             }
         }.firstOrNull()
+
+    private fun inferGqlRequest(body: String): String? {
```
MiSikora, line 86: Inferring GraphQL from a body won't work for GET requests as there is only URL query parameter for those.

It's a bit hard to have a full-proof solution as anyone could have query parameters or bodies that conform to GraphQL format. I have an idea of how to deal with it but let's just first implement handling of GET requests as well.

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -134,10 +136,13 @@ internal class HttpTransaction(
             return when (status) {
                 Status.Failed -> " ! ! !  $method $path"
                 Status.Requested -> " . . .  $method $path"
-                else -> "$responseCode $method $path"
+                else -> "$responseCode $method $path ${isGraphqlRequest.takeIf { true } ?: ""}"
             }
         }
 
+    val isGraphqlRequest: Boolean
+        get() = !gqlOperationName.isNullOrEmpty()
```
MiSikora, line 144: This doesn't seem right. The `operationName` is an optional parameter in [GraphQL](https://graphql.org/learn/serving-over-http/#post-request). I think we can rely solely on `query`. Plus there is an additional issue for GET  requests, which I mentioned here.

MiSikora (changes requested):
For a PR that introduces integration with a new specification, there are no tests. Please cover necessary scenarios such as GET and POST requests with `ChuckerInterceptor` as both application and network interceptor.

It would be also helpful if you could follow our PR template. Screenshots especially would be appreciated.

MiSikora:
> **Known Issues** Couldn't successfully run the sample until i commented this part of the code
> 
> ```
> //        StrictMode.setVmPolicy(
> //            StrictMode.VmPolicy.Builder()
> //                .detectLeakedClosableObjects()
> //                .penaltyLog()
> //                .penaltyDeath()
> //                .build()
> //        )
> //
> //        StrictMode.setThreadPolicy(
> //            StrictMode.ThreadPolicy.Builder()
> //                .detectDiskReads()
> //                .detectDiskWrites()
> //                .penaltyLog()
> //                .penaltyDeath()
> //                .build()
> //        )
> ```

I experienced it in the past as well but I thought that #750 fixed it.

cortinico:
Sorry for the late reply here.
I've noticed we do have another similar PR here: #805
I'd like to understand which of the two we prefer to move forward.

alkber:
Please close this PR in respect of a better implementation at [https://github.com/ChuckerTeam/chucker/pull/805](url)

cortinico:
> Please close this PR in respect of a better implementation at [https://github.com/ChuckerTeam/chucker/pull/805](url)

👍 We'll keep the disucssion going on #805

Issue #579, opened 2021-03-19
Expose data model


speekha:
## :warning: Is your feature request related to a problem? Please describe
We recently added a feature allowing users to export all sorts of logs in some builds of our app. Since all network logs are already store in chucker's database, it would have been very helpful to be able to query that database through Chucker's API, but it wasn't possible because most of Chucker's code is declared as internal and thus not accessible from another module.

## :bulb: Describe the solution you'd like
Repository and model classes could be exposed as public.

## :bar_chart: Describe alternatives you've considered
A current workaround is for devs to query the database themselves (basically duplicating Chucker's code), which implies running the risk of their code being broken when Chucker's design evolves without the safety of compile time verifications.

## :raising_hand: Do you want to develop this feature yourself?
Could offer help by pushing a PR if needed

MiSikora:
Generally speaking I would like to do this but not right now. The problem is that current data model falls a bit under the concept of "bag of not necessarily related properties" and is not extensible without breaking changes which would limit our release cycles.

However, as I mentioned it is definietly that I personally have in mind but from my perspective 3 things should happen first.
* Resolve #259.
* Make Chucker multi-model project.
* Support more clients than OkHttp.

This would make me more confident that the model is open to modifications without introducing breaking changes or that the breaking changes are really needed to provide new functionalities.

vbuberen:
Thanks for suggestion. From my side I don't see this feature as something I would like to see in Chucker. I don't think that we are ready to commit that much to expose even data modules. It makes us much less flexible as any change would require tons of back and forth communicating the change, etc. Would love to avoid agreeing every single internal change with users.

MiSikora:
Clarification from my side – I don't see it as something useful on its own. What I would like to achive in the future is support for other HTTP clients. This imposes modularization and mapping from client specific type to our types thus making them public. But this is not something that I see doing anytime soon.

cortinico:
> Generally speaking I would like to do this but not right now

Same for me.

To give some historical context: Chuck used to have a `ContentProvider` to expose the data to other apps. It was removed once we forked the project as it was unused and yet another feature to maintain.

Recently I was thinking about the "export" capabilities of Chucker, and ideally we could provide an extension point where you can bring your own _Exporter_. So we don't have to support multiple formats like Json, XML, HAR (#417) and so on but those can be plugged in by our users.

[...]

PR #805, opened 2022-04-21
Improve transaction list for GraphQL requests

### LINKED ISSUE #69: Add support for GraphQL

### LINKED ISSUE #116: Additional GraphQL Support
PR #70 (relating to issue #69) adds support for GraphQL by detecting an _operation name_ field in the request.  I believe that there are additional nuances to GraphQL we should address before calling our support "complete"

* Reading through [Serving over HTTP](https://graphql.org/learn/serving-over-http/) it appears as though we would be missing GraphQL icons on valid outgoing requests as "operation name" is optional.  This was the case when I ran the code against my team's _soon-to-be-released_ app.
* We have the opportunity to parse the "variables" collection to give users a more full-featured summary of the request; perhaps we could tweak the "request" tab to be more GraphQL specific (including operation name, variables and query) in the cases where we detect it's needed.
* Reading through [The Anatomy of a GraphQL Query](https://blog.apollographql.com/the-anatomy-of-a-graphql-query-6dffa9e9e747) we could probably detect the operation type (query, mutation, or subscription) without too much effort.

**Additional context**
PR #70 attempts to parse plain text requests to detect a `operationName` field and if present, the GUI receives the GraphQL icon.  It looks like there are several cases where an `operationName` will be absent, but we would want to include the icon.
* a *GET* request like `http://myapi/graphql?query={me{name}}` is valid GraphQL
* the _operation name_ would only appear in a *GET* when there are multiple operations
    > If the query contains several named operations, an operationName query parameter can be used to control which one should be executed
* for a *POST*
    > If the "query" query string parameter is present (as in the GET example above), it should be parsed and handled in the same way as the HTTP GET case.
* also, we should be sensitive to the content type and only use GSON if it is "application/json"
    > If the "application/graphql" Content-Type header is present, treat the HTTP POST body contents as the GraphQL query string.

GraphQL (like SOAP before it) exposes a single HTTP endpoint, perhaps we could register that with the `ChuckerInterceptor` and avoid introspecting the request looking for an operation name?

Also, we parse the request with GSON for all outgoing network calls, in-line with the actual network call.  Knowing that a particular URL is a non-GraphQL endpoint would allow us to skip this step.  Perhaps we could defer this work to the `HttpTransactionDatabaseRepository` as we will be on a background thread away from the actual network call?

**Do you want to develop this feature yourself?**
I could easily see several _user stories_ to extend the GraphQL support, each of which would add value and could be independently worked by people.  I would be happy to work with you on coming up with several new issues, and would also be glad to work on any of them.

Canato:
Close #69 
Close #116 

## 🎥 Video
https://user-images.githubusercontent.com/15942449/164345053-ff8ea0f5-41f5-4b58-a0d1-575b1e1f88b0.mp4

## 📷 Image
_after iterative improvement_

## :page_facing_up: Context
Based on #69, #116 and a little on #579 we have a desire for a more clean list when using GraphQL.
This helps to debug who uses Graphql without changing normal okhttp requests

## :pencil: Changes
Retrieve request header for transaction list and display it if is a GraphQl request and have the `operationName`

On the Sample code, there is a new button `Do GraphQl Activity` that can be pressed for testing.

## :stopwatch: Next steps
Wanna be sure the code I add is following the library standards, open to suggestions!


Canato:
Great libraries people, hope this is useful and we can release it soon
@cortinico @MiSikora @olivierperez @vbuberen

cortinico:
Hey @Canato,
Have you seen the PR at #800?
I'd like to coordinate between the two to make sure we merge only one of the two approach.

Canato:
> Hey @Canato, Have you seen the PR at #800? I'd like to coordinate between the two to make sure we merge only one of the two approach.

Ohh @cortinico Thanks! I saw #70 because I come from the issue, but didn't spot #800 when I start to do these changes!

I'm happy to move with 800 if @alkber wanna fix the issues open there. I'm happy to close this one.

From another point of view, my implementation is a simple one and since the tests passed(if the code is valid) we could merge and use the 800 as an iterative improvement. 
I was already planning to return with more requests/tests/icons, but I believe we should have small PRs and do them in small batches.

I see that my sample implementation differs from the other PR too because I'm using a real schema from the guides, would be interesting to know what will be the best solution.

That said, any of the solutions work for me if we can put them to work =D. Let me know @alkber

alkber:
> > Hey @Canato, Have you seen the PR at #800? I'd like to coordinate between the two to make sure we merge only one of the two approach.
> 
> Ohh @cortinico Thanks! I saw #70 because I come from the issue, but didn't spot #800 when I start to do these changes!
> 
> I'm happy to move with 800 if @alkber wanna fix the issues open there. I'm happy to close this one.
> 
> From another point of view, my implementation is a simple one and since the tests passed(if the code is valid) we could merge and use the 800 as an iterative improvement. I was already planning to return with more requests/tests/icons, but I believe we should have small PRs and do them in small batches.
> 
> I see that my sample implementation differs from the other PR too because I'm using a real schema from the guides, would be interesting to know what will be the best solution.
> 
> That said, any of the solutions work for me if we can put them to work =D. Let me know @alkber

Hi @cortinico  , isorry i was really busy with personal work and couldn't address the comments in PR #800. I believe @Canato approach is more generic and neat than mine as i read the code, and he has decorated the PR with more test cases too.. You may close PR #800  and use this one. However I have a small suggestion, to @Canato PR. If the UI item can be updated to the way PR #800 does then it would be much appreciated and visually appealing.  @Canato can you take it up? you can reuse my work in your code.

Canato:
If @cortinico is ok I can update the UI to use the graphql icon and add a new line ^^. Thanks @alkber

Canato:
@cortinico news on this? =D

Canato:
Add the GraphQL icon and new field for it.

Screenshot_20220601_165247_com chuckerteam chucker sample

downloadApolloSchema.sh
```diff
@@ -0,0 +1,9 @@
+./gradlew downloadApolloSchema --endpoint="https://rickandmortyapi.com/graphql" --schema="sample/src/main/graphql/com/chuckerteam/chucker/sample/schema.json"
```
cortinico, line 1: Please add a `#!/bin/bash` on top

downloadApolloSchema.sh
```diff
@@ -0,0 +1,9 @@
...
+
+# GraphQL Schema Update
```
cortinico, line 3: Move those before the command please

downloadApolloSchema.sh
```diff
@@ -0,0 +1,9 @@
...
+# Open Android Studio terminal
+# Run command `./downloadApolloSchema.sh`
+# Format the file (`ctrl+option+i` && `option+cmd+L`), if need
+# Replace `"null"` to `null`, if need
```
cortinico, line 7: Why this is needed? How can a user know if they need to do it?

downloadApolloSchema.sh
```diff
@@ -0,0 +1,9 @@
...
+# Go to the top of the file anc click the run icon "Generate GraphQL SDL schema file"
```
cortinico, line 8:
```suggestion
# Go to the top of the file and click the run icon "Generate GraphQL SDL schema file"
```

cortinico: Which Icon? Inside which IDE?

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransactionTuple.kt
```diff
@@ -17,6 +19,7 @@ internal data class HttpTransactionTuple(
     @ColumnInfo(name = "protocol") var protocol: String?,
     @ColumnInfo(name = "method") var method: String?,
     @ColumnInfo(name = "host") var host: String?,
+    @ColumnInfo(name = "requestHeaders") var requestHeaders: String?,
```
cortinico, line 22: Commenting here but valid for the whole PR:

1. I would not expose `requestHeaders` at all. It can be quite big and contains several unnecessary fields
2. I would instead create a new field called `graphQlOperationName` or `graphQlPath` and add it to both `HttpTransactionTuple` and `HttpTransaction`
3. You can then populate this inside the `RequestProcessor` so you'll just have a string that you can simply visualize. No need to inspect the whole Header set for what you need.

Also please add tests when you touch `RequestProcessor`

library/src/main/res/layout/chucker_list_item_transaction.xml
```diff
@@ -40,27 +41,51 @@
...
 
     <TextView
-        android:id="@+id/host"
+        android:id="@+id/graphqlPath"
         android:layout_width="0dp"
         android:layout_height="wrap_content"
         android:layout_marginStart="@dimen/chucker_half_grid"
         android:ellipsize="end"
         android:maxLines="2"
+        android:textAppearance="@style/Chucker.TextAppearance.ListItem"
```
cortinico, line 50: I'm unsure we want to use this style as it will make it bold. Ideally having it styled as the `host` would be better 👍

cortinico (changes requested):
Great stuff! Thanks for doing this. Sorry for the late review 🙏 
Let me know if you need some help integrating the changes I suggested

downloadApolloSchema.sh
Canato: the `if need` here was related to if we have these fields. Indeed could make confusing. Will remove.

downloadApolloSchema.sh
Canato: Is the play icon (will update the name and icon)
▶️

Add the information about IDE supported.

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransactionTuple.kt
Canato: I did an implementation first where I called `fun setGraphQlOperationName(headers: List<HttpHeader>)` method inside `fun setRequestHeaders(headers: List<HttpHeader>) {...}` but that felt like some hide knowledge and harder to test.

To make the methods with a single responsibility and clean structure I decided to split the calls and direct called on `RequestProcessor`

This created a repetition of the method `toHttpHeaderList(headers)` but I believe it will bring more safety for maintenance, let me know what you think 

(plus I didn't add any room version migration, but look ok)

library/src/main/res/layout/chucker_list_item_transaction.xml
Canato: Done! @cortinico

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -156,6 +158,11 @@ internal class HttpTransaction(
         requestHeaders = JsonConverter.instance.toJson(headers)
     }
 
+    fun setGraphQlOperationName(headers: Headers) {
+        graphQlOperationName = toHttpHeaderList(headers)
+            .find { it.name.lowercase().contains("operation-name") }?.value
+    }
+
```
ArjanSM, line 165: @cortinico I may have missed something but still curious. Wasn't the plan to let the developers register a GraphQL endpoint in the `ChuckerInterceptor` so that we could avoid identifying GraphQL operations on the basis of certain headers?
I may be wrong but it appears we still may not be able to support GraphQL HTTP GET requests using this change.
Any thoughts?

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -53,7 +53,8 @@ internal class HttpTransaction(
     @ColumnInfo(name = "responseHeadersSize") var responseHeadersSize: Long?,
     @ColumnInfo(name = "responseBody") var responseBody: String?,
     @ColumnInfo(name = "isResponseBodyEncoded") var isResponseBodyEncoded: Boolean = false,
-    @ColumnInfo(name = "responseImageData") var responseImageData: ByteArray?
+    @ColumnInfo(name = "responseImageData") var responseImageData: ByteArray?,
+    @ColumnInfo(name = "graphQlOperationName") var graphQlOperationName: String?,
```
ArjanSM, line 57: Just curious, but wouldn't we need to include a check for this parameter in the `fun hasTheSameContent()`?

Canato:
Any news one this? Worth to solve the conflicts ?

vbuberen:
> Any news on this? Worth to solve the conflicts ?

Yes, worth solving. Let's make it happen.

library/src/main/res/layout/chucker_list_item_transaction.xml
cortinico: Nope this is good to go 👍

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
cortinico: Good call, I've added it 👍

cortinico (approved):
I've rebased your changes @Canato and fixed the detekt failures :+1: This is good to merge on my end.

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
cortinico: > Though majority of clients use Apollo-Kotlin to execute GraphQL requests but it appears we still may not be able to support GraphQL HTTP GET requests.
Any thoughts?

That's correct. For the time being we're assuming you're having a "operation-name" header. We can expand on this condition in the future to find the Graph QL Operation Name.

Canato:
Thanks @cortinico how do we know when will be the next release or if this will be available in a snapshot?

cortinico:
No sorry we don't have an ETA at the moment. I'm looking into wrapping up the work for a 4.x.
We probably want to include the #259 work which will take some time to do (as it hasn't started yet). In the meanwhile you can use the SNAPSHOT versions for 4.x

Canato:
No issues, Thanks! =D

alifernandez45 (changes requested):
- get-pdf-tc-icomm.pdf

PR #815, opened 2022-05-03
Adding chucker to diff. build types

gsrathoreniks:
## :page_facing_up: Context

The current method of adding dependency doesn't work for build types apart from `debug` and `release`

## :pencil: Changes

Updated the implementation guide so that it can be implemented across android apps with different `buildTypes` as well.

- Create new buildType with any name apart from `debug` and `release`
- Build the project and it'll fail.
- Now follow the implementation guide added with this PR.
- This will enable/disable the chucker for that particular `buildType`

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,24 @@ dependencies {
   releaseImplementation "com.github.chuckerteam.chucker:library-no-op:3.5.2"
 }
 ```
+If you're using different build types for your project for example `releaseDebug` like below
+```
+releaseDebug {
+      minifyEnabled false
+      debuggable true
+      shrinkResources false
+      versionNameSuffix "-releaseDebug"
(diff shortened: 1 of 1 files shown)


vbuberen:
I don't see any value in this addition. You just specified your particular case here and it is not universal for other setups other users might have. You also add details, like `debuggable...`, etc. which are irrelevant to Chucker setup. Can't accept this PR.

cortinico:
> The current method of adding dependency doesn't work for build types apart from `debug` and `release`

Also this is not true. If you have a `staging` build type, you can totally add Chucker as `stagingImplementation`, `stagingReleaseImplementation`, `stagingDebugImplementation` and so on. There si no need of extra configuration

PR #832, opened 2022-06-03
Fix various build warnings

cortinico:
## :page_facing_up: Context
Just noticed we have a variety of build warnings I'd like to resolve as they're noisy and not really valuable.

## :pencil: Changes
- Remove jcenter()
- Set maxIssues for detekt to 0, and suppressed `LongMethod` on tests
- Suppressed UNUSED_PARAMETER on `library-no-op`
- Fixed several type inference issues in our tests

Essentially wait for a green CI.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,12 +77,6 @@ allprojects {
     repositories {
         google()
         mavenCentral()
-        jcenter {
-            content {
-                includeModule("org.jetbrains.trove4j", "trove4j")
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -77,12 +77,6 @@ allprojects {
     repositories {
         google()
         mavenCentral()
-        jcenter {
```
vbuberen, line 80: Great that we can finally get rid of it.

vbuberen (approved):

PR #853, opened 2022-07-20
Support for manual transaction logging

aliab:
## :page_facing_up: Context
Add Support for manual transaction logging in the chucker. For example, developers can populate `MQTT` transactions into their chucker log.

## :pencil: Changes
-  I've added  added `saveTransaction(transaction: ManualHttpTransaction)` method to `ChuckerCollector` to save manual transaction data into DB.
- I've added a new `ManualHttpTransaction` entity to the API package for the manual type of logging. It is wrapped over the internal `HttpTransaction` entity and ensures that previous visibility is not violated.

## :no_entry_sign: Breaking
No breaking changes introduced

Just run the sample app. The new ManualTransactionTask is added to the sample app `MainActivity` httpTasks to test if this feature is working OK.


library/src/main/kotlin/com/chuckerteam/chucker/api/entity/ManualHttpTransaction.kt
```diff
@@ -0,0 +1,74 @@
+package com.chuckerteam.chucker.api.entity
+
+import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
+
+/**
+ * Represent Manual Http transaction that developer want to populate as an custom http transaction
+ */
+public data class ManualHttpTransaction(
```
cortinico, line 8: Why do we need another class here?

cortinico:
Thanks for sending this @aliab 🙏 I think we can simplify this a bit

library/src/main/kotlin/com/chuckerteam/chucker/api/entity/ManualHttpTransaction.kt
aliab: I didn't want to move `HttpTransaction` class to the API package or expose it to the developer. I could use a builder pattern to build transaction objects or create a method with a bunch of args in ChuckerCollector class.
but both of these mentioned approaches has it downsides regarding logging the data. In this implemented approach, to be able to log the start and end information of a transaction, the developer no longer needs to store the data in a separate data store and can create the model and fill it in during transactions. at the end send it to be stored in the chucker.

aliab:
> Thanks for sending this @aliab 🙏 I think we can simplify this a bit

I tried to make as few changes as possible. btw, I want this feature in my app and if you give me a hint about your idea, maybe I could deliver it sooner.

aliab:
@cortinico kindly reminder

cortinico:
> @cortinico kindly reminder

Hey @aliab 
I've took a closer look at this change and I think we can't merge it now. Specifically as I'd like to focus on:
- https://github.com/ChuckerTeam/chucker/issues/259
- https://github.com/ChuckerTeam/chucker/issues/579

Exposing the Manual class you suggested exposes us to having to maintain this in the long run. As we're planning on potentially separating requests and responses, having a class in the public API which doesn't follow this design would be harder to maintain the future.

Let's get back to this once we're done with the redesign.

PR #879, opened 2022-09-10
Add support for Android 13

cortinico:
## :camera: Screenshots
chucker-notification-permission

## :page_facing_up: Context

This commit adds support for Android 13 and add supports for requesting the notification permission.

## :pencil: Changes

- Bump of compileSdk to 33
- Bump of appcompat to 1.5.0
- Adding support for a `Chucker.requestNotificationPermission` util method to help users request the Notification permission.

Apps that are already requesting such permission, should not use the method and Chucker will just work.

## :paperclip: Related PR

Fixes #854
Closes #869
Closes #881

## :hammer_and_wrench: How to test
Tested on an API 33 emulator. On Api <= 32 Chucker works correctly.

## :stopwatch: Next steps
We might want to backport this to 3.x?

PR #887, opened 2022-09-18
Updates for Android 13

### LINKED ISSUE #854: Chucker library not working for Android 13
Chucker library not working for Android 13

## :bomb: Steps to reproduce
Reproduce the behavior:
1. Update the compileSDK and targetSDK version to 33 
2. Add chucker dependency in build.gradle
`implementation "com.github.chuckerteam.chucker:library:3.5.2"`
3. Add below code in your OkHttpClient builder
`ChuckerCollector chuckerCollector = new ChuckerCollector(context,
                    true, RetentionManager.Period.ONE_HOUR);
            ChuckerInterceptor chuckerInterceptor = new ChuckerInterceptor.Builder(context)
                    .collector(chuckerCollector)
                    .maxContentLength(250_000L)
                    .alwaysReadResponseBody(true)
                    .build();`

## :wrench: Expected behavior
When the app is launched, the network logs should get displayed on the notification panel

## :Android: Tech info
 - Device: Pixel 4 XL
 - OS: Android 13
 - Chucker version: 3.5.2

Chucker library was working fine till Android 12. After upgrading the app to Android 13, I am not getting the http logs.

Is it supporting Android 13 or not ?

If it's supporting then how to integrate it.

vbuberen:
## :camera: Videos
[notifications_success_scenario.webm](https://user-images.githubusercontent.com/13467769/190923556-cbffb2ea-54eb-4b59-bf63-3fdfabad6e77.webm)
[notifications_disable_scenario.webm](https://user-images.githubusercontent.com/13467769/190923551-702943a6-b310-4645-8b41-d8c0d0a2cdee.webm)

## :page_facing_up: Context
An alternative approach to #879 . I wanted to avoid forcing users to modify their code somehow for Android 13, so went with the way where notification permission is added into Chucker's AndroidManifest. From a user side the only required action is to open Chucker after adding (ideally, via shortcut which we provide by default) and grant the permission. The flow feels like not the most convenient one, but I don't see other easier ways at the moment to handle this new permission request.

## :pencil: Changes
- Bumped `targetSDK` and `compileSDK` to 33
- Added notifications permission handling on devices with Android 13

## :no_entry_sign: Breaking
Nothing breaking is expected

Run Chucker sample on a device/emulator with Android 13 and try to open Chucker directly.

## :stopwatch: Next steps
As I suggest to use the shortcut which we add, we might want to remove the possibility of disabling shortcut creation from our API to not confuse users.

Closes #854


build.gradle
```diff
@@ -101,8 +101,8 @@ task clean(type: Delete) {
 
 ext {
     minSdkVersion = 21
-    targetSdkVersion = 31
-    compileSdkVersion = 31
+    targetSdkVersion = 33
```
cortinico, line 104: might be worth also bumping app compat to 1.5.x in this same PR?

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
```diff
@@ -39,6 +50,15 @@ internal class MainActivity :
     private val applicationName: CharSequence
         get() = applicationInfo.loadLabel(packageManager)
 
+    private val permissionRequest = registerForActivityResult(
+        ActivityResultContracts.RequestPermission()
+    ) { isPermissionGranted: Boolean ->
+        if (!isPermissionGranted) {
+            showToast(applicationContext.getString(R.string.chucker_notifications_permission_not_granted))
+            Logger.error("Notification permission denied. Can`t show transactions info")
```
cortinico, line 58:
```suggestion
            Logger.error("Notification permission denied. Can't show transactions info")
```

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
```diff
@@ -39,6 +50,15 @@ internal class MainActivity :
...
+        }
```
cortinico, line 59: Might be worth to also log here a success message?

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
```diff
@@ -39,6 +50,15 @@ internal class MainActivity :
     private val applicationName: CharSequence
         get() = applicationInfo.loadLabel(packageManager)
 
+    private val permissionRequest = registerForActivityResult(
+        ActivityResultContracts.RequestPermission()
+    ) { isPermissionGranted: Boolean ->
+        if (!isPermissionGranted) {
+            showToast(applicationContext.getString(R.string.chucker_notifications_permission_not_granted))
```
cortinico, line 57: I would make this a SnackBar as well with Long Duration.

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
```diff
@@ -55,18 +75,54 @@ internal class MainActivity :
...
+                    applicationContext.getString(R.string.chucker_notifications_permission_not_granted),
+                    Snackbar.LENGTH_LONG
+                ).setAction(applicationContext.getString(R.string.chucker_change)) {
+                    Intent(Settings.ACTION_APPLICATION_DETAILS_SETTINGS).apply {
+                        addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                        data = Uri.fromParts("package", packageName, null)
+                    }.also { intent ->
+                        startActivity(intent)
+                    }
+                }.show()
```
cortinico, line 120: Here we're breaking a bit from the Google's guideline where the rationale should just explain why we need the permission and we should still allow the user to pick yes/no. If yes -> we should fire the `permissionRequest`.

Letting the user jump to the Settings screen sort of assumes that the user selected No on the past.

cortinico (commented):
Great stuff, I prefer this approach than mine 👍 

The only thing is that I would save the:

```kotlin
    private fun shouldShowNotification() =
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
            !BaseChuckerActivity.isInForeground && notificationManager.areNotificationsEnabled()
        } else {
            !BaseChuckerActivity.isInForeground
        }
```

to be added inside `NotificatioHelper.show` as querying for `areNotificationsEnabled` seems to be the recommended way to verify if notifications are enabled 👍

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
vbuberen: I don't think it is worth it.

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
vbuberen: Yes, I did it on purpose as I assume that Chucker is used by devs, not regular users, so showing one more dialog to somebody, who disabled permission on purpose on first launch is going to be annoying. Devs know how and why permission is needed, especially, since showing notifications is what Chucker does. 
Instead, you have a snackbar and can change settings if you wish or just mind your own business with Chucker without dismissing the dialog.

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
vbuberen: We show toasts for all other fault/errors states. The reason I used Snackbar for redirection to Settings is to have an action button, so it is convenient to just click on it and end up in settings. Here we don't have any action and to be consistent with other failed operations (like exporting stuff, etc.) I use Toast.

build.gradle
vbuberen: Yes, can do that as well.

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
cortinico: Ok but you log a:

```
Logger.info("Notification permission granted")
```

Just below. Shoudl we remove that as well then?

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
cortinico: Mmm I understand.

For me it was a bit odd as, when playing with the app I saw both the Toast and the SnackBar with the same error message, so I thought it was a bug.

I'm fine having it as a toast, but I would still have Duration Long as I barely had the time to read it.

cortinico (approved):
LGTM 👍 Minor comments that needs to follow up, but this is good to merge on my end. Thanks for doing it!

yogiseralia:
Eagerly waiting for this update, when can we expect a new release with these changes??

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
vbuberen: Good catch. Didn't like empty braces initially. Removed.

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
vbuberen: Made it long.

PR #908, opened 2022-10-16
Minor cleanups

yagmurerdogan:
## :pencil: Changes

- No any UI related change
- Reformat xml file with option + command + L 
- Create dimen resource for zero dp 
- Use orEmpty() instead of ?: ""
- Delete + char from constraints view id in xml files
- Please add "hacktoberfest-accepted" label, thank you!

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/ResponseProcessor.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/ResponseProcessor.kt
@@ -141,7 +141,6 @@ internal class ResponseProcessor(
 
     private companion object {
         const val MAX_BLOB_SIZE = 1_000_000L
-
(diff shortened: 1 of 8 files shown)


cortinico (approved):

vbuberen:
> Reformat xml file with option + command + L

There are Actions on our CI that take care of checking the formatting of code and fail if somebody's PR has bad formatting. This, the project code is formatted fine even without this action.

> Please add "hacktoberfest-accepted" label, thank you!

I don't think it is appropriate to ask such thing. For me it is just an additional confirmation that the value of the PR is really low and all changes are mainly to get one more PR counted towards Hacktoberfest. Don't want to encourage other people do the same, so against merging this PR.

If you would like to have something counted towards Hacktoberfest pleasepixk something among opened issues in Chucker. Fix/changes to resolve one of the issues will definitely be valuable and I would gladly accept it in case the quality is good.

PR #918, opened 2022-11-01
Support EventListener to decorate HttpTransaction recordings

carlonzo:
## :camera: Screenshots

image

## :page_facing_up: Context
Added EventListener to decorate recordings for HttoTransaction.

This will allow to record duration for:
- Call
- Connect
- Secure Connect
- Dns

## todo 
- create tests

I am not sure if this feature is interesting for a Chucker user, but wanted to send a PR to start a discussion of it

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerEventListener.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerEventListener.kt
@@ -1,0 +1,75 @@
+package com.chuckerteam.chucker.api
+
+import com.chuckerteam.chucker.internal.data.entity.HttpTransaction
+import okhttp3.Call
+import okhttp3.EventListener
+import okhttp3.Handshake
+import okhttp3.Protocol
+import java.io.IOException
+import java.net.InetAddress
+import java.net.InetSocketAddress
+import java.net.Proxy
+
+internal class ChuckerEventListener constructor(
+    private val onCallEnded: ((Call) -> Unit)
+) : EventListener() {
+
+    val httpTransaction = HttpTransaction()
+
+    override fun callStart(call: Call) {
+        super.callStart(call)
+        httpTransaction.callStartDate = System.currentTimeMillis()
+    }
+
+    override fun callEnd(call: Call) {
+        super.callEnd(call)
+        onCallEnded(call)
+        httpTransaction.callEndDate = System.currentTimeMillis()
+    }
+
+    override fun callFailed(call: Call, ioe: IOException) {
+        super.callFailed(call, ioe)
+        onCallEnded(call)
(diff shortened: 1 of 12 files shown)


carlonzo:
The reason behind this PR is to understand how long a network request actually takes. Today any interceptor added after Chucker that runs expensive operations, could increase the time required to for a request to actually return to the user.

Also will add infos about the time it was required to acquire a connection and the time required by the dns

cortinico:
> I am not sure if this feature is interesting for a Chucker user, but wanted to send a PR to start a discussion of it

Thanks for sending this PR @carlonzo 
I believe we're not looking into adding this degree of granularity around events for Chucker. If you need this type of information (e.g. DNS start/end), you can create your own OkHTTP interceptor and get those information and create your own callbacks. I don't think Chucker should be doing this.

carlonzo:
Thanks for reviewing my PR.

The reason I created it was because this level of granularity may be interesting to be added to get information regarding the health of the network as well as the the cost of the app network stack. On my own project this was important to gather the cost in duration of the connection as well as the the time spent by all the interceptors added to okhttp.
I believe that adding the EventListener (an interceptor can't give this level of details) as opt-in feature may help developers understand the full picture of every single connection and not just the single request-response data.

Thanks for considering and for maintaining this project.

cortinico:
> I believe that adding the EventListener (an interceptor can't give this level of details) as opt-in feature may help developers understand the full picture of every single connection and not just the single request-response data.

True but also quite invasive from the API point of view. How do you feel about having a custom interceptor that does this for you? Why do you feel Chucker should do it instead?

cortinico:
> What part of the API you feel is invasive?

The fact that you're adding 8 new fields in the database + the fact that you added an `internal class ChuckerEventListener` inside the `.api` package. It's also unclear to me how do you expect the users to interact with this.

carlonzo:
Sorry for the delay, but somehow I didn't see your reply.

Thanks for it. For API here I don't understand what you mean. The public API has changed with a new public method available from the Interceptor where I've added `public fun useEventListener(): EventListener.Factory`.
The 8 new fields in the DB are not part of the API, but are a fields used by the private implementation. I can easily drop the 8 to 4 if the number if that is an issue. Is the DB supposed to be accessible to the user or external parties?

The `ChuckerEventListener` in the `api` package is a mistake. I will move it the internal one. 

I can see it my change invasive in 2 points:
- new api method in the interceptor
- the `HttpTransactionFactory` I needed to add the same instance of HttpTransaction per Call, to allow the EventListener to decorate the transaction with the extra infos required.

cortinico:
> Is the DB supposed to be accessible to the user or external parties?

Nope is not. At least not in the current stage as it's not properly encapsulated.

PR #970, opened 2023-02-24
Add `skipPaths` to selectively skip requests from Chucker

### LINKED ISSUE #266: Allow to selectively skip Chucker interception

ArjanSM:
## :page_facing_up: Context

Fixes #266 

## :pencil: Changes

I added a public API, `skipPaths(paths:List<String>)`, to the `ChuckerInterceptor.Builer`. 
The ChuckerInterceptor decides to process the transaction based on the outcome of the lambda in `skipEndpoints`.

## :no_entry_sign: Breaking

Adding `skipPaths(..)` to the `ChuckerInterceptor.Build` will be a breaking change. 
- Added tests to `ChuckerInterceptorTest`
- I've also added another endpoint `/anything` to the [HttpBinHttpTask.kt](sample/src/main/kotlin/com/chuckerteam/chucker/sample/HttpBinHttpTask.kt) and added `/anything` to the skip paths list in the sample app.

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -77,10 +77,10 @@ public class ChuckerInterceptor private constructor(
         }
         val response = try {
             chain.proceed(request)
-        } catch (exception: IOException) {
-            transaction.error = exception.toString()
+        } catch (e: IOException) {
+            transaction.error = e.toString()
             collector.onResponseReceived(transaction)
-            throw exception
+            throw e
         }
         return if(shouldProcessTheRequest)
             responseProcessor.process(response,transaction)
--- a/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -7,8 +7,8 @@ import com.chuckerteam.chucker.internal.support.CacheDirectoryProvider
 import com.chuckerteam.chucker.internal.support.PlainTextDecoder
 import com.chuckerteam.chucker.internal.support.RequestProcessor
 import com.chuckerteam.chucker.internal.support.ResponseProcessor
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.Interceptor
-import okhttp3.Request
 import okhttp3.Response
 import java.io.IOException
 
@@ -54,7 +54,7 @@ public class ChuckerInterceptor private constructor(
         decoders,
     )
 
-    private val skipEndpoints = builder.skipEndpoints.toSet()
+    private val skipPaths = builder.skipPaths.toSet()
 
     init {
         if (builder.createShortcut) {
@@ -71,8 +71,7 @@ public class ChuckerInterceptor private constructor(
     override fun intercept(chain: Interceptor.Chain): Response {
         val transaction = HttpTransaction()
         val request = chain.request()
-
-        return if(skipEndpoints.any { it(request) } ) {
+        return if(skipPaths.any { it == request.url.encodedPath }) {
              chain.proceed(request)
         } else {
             requestProcessor.process(request, transaction)
@@ -92,6 +91,7 @@ public class ChuckerInterceptor private constructor(
      *
      * @param context An Android [Context].
      */
+    @Suppress("TooManyFunctions")
     public class Builder(internal var context: Context) {
         internal var collector: ChuckerCollector? = null
         internal var maxContentLength = MAX_CONTENT_LENGTH
@@ -100,7 +100,7 @@ public class ChuckerInterceptor private constructor(
         internal var headersToRedact = emptySet<String>()
         internal var decoders = emptyList<BodyDecoder>()
         internal var createShortcut = true
-        internal var skipEndpoints = mutableSetOf<(Request) -> Boolean>()
+        internal var skipPaths = mutableSetOf<String>()
 
         /**
          * Sets the [ChuckerCollector] to customize data retention.
@@ -170,8 +170,20 @@ public class ChuckerInterceptor private constructor(
             this.cacheDirectoryProvider = provider
         }
 
-        internal fun skipEndpoints(comparisonLogics: List<(Request) -> Boolean>): Builder = apply {
-            this.skipEndpoints.addAll(comparisonLogics)
+        public fun skipPaths(skipPaths: List<String>): Builder = apply {
+            skipPaths.forEach { candidatePath ->
+                processSkipPaths(candidatePath)
+            }
+        }
+
+        private fun processSkipPaths(candidatePath: String) {
+            if (candidatePath.isNotBlank()) {
+                candidatePath.toHttpUrlOrNull()
+                    ?: "http://localhost:8080$candidatePath".toHttpUrlOrNull()
+                        ?.let { validHttpUrl ->
+                            this@Builder.skipPaths.add(validHttpUrl.encodedPath)
+                        }
+            }
         }
 
         /**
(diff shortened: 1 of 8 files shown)


cortinico:
I'd like to test this out @ArjanSM on the sample app but I've just noticed it instacrash due to a StrictMode violation :|

library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -164,6 +170,10 @@ public class ChuckerInterceptor private constructor(
             this.cacheDirectoryProvider = provider
         }
 
+        internal fun skipEndpoints(comparisonLogics: List<(Request) -> Boolean>): Builder = apply {
```
cortinico, line 173: (this should be `public` if you want to expose it)

cortinico (changes requested):
Thanks for sending this over @ArjanSM 

I don't think we want to evolve the API in this direction as this will expose the `Request` object from OkHTTP to the user.
A user can manipulate the request (i.e. Read the body) in way we cannot prevent.

I think the right way to approach this feature is either:
- Provide a list of paths to use for the skip (a `List<String>`)
- Provide a list of header to strip and use for the skip (again a `List<String>`)

library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -164,6 +170,23 @@ public class ChuckerInterceptor private constructor(
             this.cacheDirectoryProvider = provider
         }
 
+        public fun skipPaths(skipPaths: List<String>): Builder = apply {
```
cortinico, line 173: Maybe this should be a `vararg` of String?

library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -164,6 +170,23 @@ public class ChuckerInterceptor private constructor(
...
+            skipPaths.forEach { candidatePath ->
+                processSkipPaths(candidatePath)
+            }
+        }
+
+        private fun processSkipPaths(candidatePath: String) {
+            if (candidatePath.isNotBlank()) {
+                val nullableHttpUrl = candidatePath.toHttpUrlOrNull()
+                    ?: "http://localhost:8080$candidatePath".toHttpUrlOrNull()
```
cortinico, line 182: Not a super big fan of this. You should do something like:

```
HttpUrl.Builder().scheme("https").host("example.com").addPathSegment(candidatePath).build()
```

cortinico (commented):
I think the new approach goes in the right direction 👍 
Left a couple of comments.

The CI is still red though, can you look into it @ArjanSM

library/src/test/kotlin/com/chuckerteam/chucker/util/ChuckerInterceptorDelegate.kt
```diff
@@ -42,6 +43,7 @@ internal class ChuckerInterceptorDelegate(
         .redactHeaders(headersToRedact)
         .alwaysReadResponseBody(alwaysReadResponseBody)
         .cacheDirectorProvider(cacheDirectoryProvider)
+        .skipPaths(skipPaths = skipPaths.map{it}.toTypedArray())
```
cortinico, line 46: is `.map{it}.` necessary?

library/src/test/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptorTest.kt
```diff
@@ -630,4 +632,80 @@ internal class ChuckerInterceptorTest {
...
+        chuckerInterceptorWithoutSkipping.expectNoTransactions()
+
+        executeRequestForPath(client,"www.bla.c/skip/path","Hello, world!")
+        chuckerInterceptorWithoutSkipping.expectNoTransactions()
+
+        executeRequestForPath(client,"bla.c/skip/path","Hello, world!")
+        chuckerInterceptorWithoutSkipping.expectNoTransactions()
+
+        executeRequestForPath(client, "90","Hello, world!")
+        chuckerInterceptorWithoutSkipping.expectNoTransactions()
```
cortinico, line 685: This code looks like not formatted right (look at the space before/after commas)

library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -68,25 +71,27 @@ public class ChuckerInterceptor private constructor(
...
-
-        val response = try {
-            chain.proceed(request)
-        } catch (e: IOException) {
-            transaction.error = e.toString()
-            collector.onResponseReceived(transaction)
-            throw e
+        return if(skipPaths.any { it == request.url.encodedPath }) {
+             chain.proceed(request)
+        } else {
```
cortinico, line 76: you won't be catching `IOException` here.
Please update this logic so that:

1. You check if the path has to be skipped it, and save it in a boolean variable.
2. If true, call `requestProcessor.process`, `collector.onResponseReceived` (for the error)` and `responseProcessor.process`

cortinico (changes requested):

library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
ArjanSM: A little confused about this one.

Did you mean something like the following:-
```
@Throws(IOException::class)
    override fun intercept(chain: Interceptor.Chain): Response {
        val transaction = HttpTransaction()
        val request = chain.request()
        val skipProcessing = skipPaths.any { it == request.url.encodedPath } 
        return if(skipProcessing) {
             chain.proceed(request)
        } else {
            requestProcessor.process(request, transaction)
            val response = proceedRequest(chain, request, transaction)
            responseProcessor.process(response, transaction)
        }
    }

    private fun proceedRequest(
        chain: Interceptor.Chain,
        request: Request,
        transaction: HttpTransaction
    ): Response {
        val response = try {
            chain.proceed(request)
        } catch (e: IOException) {
            transaction.error = e.toString()
            collector.onResponseReceived(transaction)
            throw e
        }
        return response
    }

Comment (cortinico) [reply library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt]: I meant something like:

```kotlin
		val shouldProcess = !(skipPaths.any { it == request.url.encodedPath })
		if (shouldProcess) {
		    requestProcessor.process(request, transaction)
		}

        val response = try {
            chain.proceed(request)
        } catch (e: IOException) {
            transaction.error = e.toString()
			if (shouldProcess) {
                collector.onResponseReceived(transaction)
            }
            throw e
        }

        return if (shouldProcess) responseProcessor.process(response, transaction) else response
```

ArjanSM: Ah! I see what you meant. This probably saves us from the redundant callsite of  `chain.proceed(..)` as well.

cortinico: Awesome I'll check this out later today 👍

library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
```diff
@@ -68,25 +71,28 @@ public class ChuckerInterceptor private constructor(
...
         } catch (e: IOException) {
             transaction.error = e.toString()
             collector.onResponseReceived(transaction)
             throw e
         }
-
-        return responseProcessor.process(response, transaction)
+        return if(shouldProcessTheRequest)
+            responseProcessor.process(response,transaction)
+        else response
```
cortinico, line 87: Let's merge it as it is, but I believe KtLint is broken here as this is not properly formatted

cortinico (approved):

PR #975, opened 2023-02-25
Fix `DiskReadViolation` on ChuckerCollector

cortinico:
## :page_facing_up: Context
The sample app is insta-crashing when doing http activity.
That happens because we're calling `RepositoryProvider.transaction()....` on the UI thread.

## 🔴 Video
| Before | After |
| -- | -- |
| before | after |

## :no_entry_sign: Breaking
nope

See the atteched video

## :stopwatch: Next steps
n/a

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerCollector.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerCollector.kt
@@ -47,7 +47,9 @@ public class ChuckerCollector @JvmOverloads constructor(
      */
     internal fun onRequestSent(transaction: HttpTransaction) {
         scope.launch {
-            RepositoryProvider.transaction().insertTransaction(transaction)
+            withContext(Dispatchers.IO) {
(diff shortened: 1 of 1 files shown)


vbuberen:
It is strange, because Room handles switching to a different thread automatically for quite some time already: 
https://medium.com/androiddevelopers/room-coroutines-422b786dc4c5

cortinico:
> It is strange, because Room handles switching to a different thread automatically for quite some time already: [medium.com/androiddevelopers/room-coroutines-422b786dc4c5](https://medium.com/androiddevelopers/room-coroutines-422b786dc4c5)

You're right 👍 

Seems like there is a bug in Room 2.5.0: https://issuetracker.google.com/issues/270841445

How do we proceed? Either we merge this as it is, or we suppress the penalityDeath for this scenario as well?

vbuberen:
Thanks for finding an open issue.
Let's merge as it is instead of suppressing `penalityDeath`.

vbuberen (approved):

PR #979, opened 2023-03-01
Fix android 13 issue

eshraqibrahimshaker:


vbuberen:
Low quality PR without proper description and with a change which is already implemented on `develop`.

PR #992, opened 2023-03-28
merge debug resources failed android studio

Gnanendra18:
Execution failed for task ':neointcore:mergeFMCLiveDebugResources'.
> A failure occurred while executing com.android.build.gradle.internal.res.ResourceCompilerRunnable
   > Resource compilation failed (Failed to compile values resource file

## :camera: Screenshots
image


cortinico:
Thanks for sending this over @Gnanendra18 

I don't believe this is the proper fix. If HTML inside a string resource is not valid, we should replace it with a `SpannableString` or so, you can't just remove the link 👍

cortinico (changes requested):
See my previous comment

Issue #1017, opened 2023-04-26
spanJson method does not complete for large JSON strings


kmayoral:
https://github.com/ChuckerTeam/chucker/blob/a4f33c037b2c098a45c6574020f8acebfa334886/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/SpanTextUtil.kt#L28

Thanks for all the changes to everyone involved! I noticed when checking this out locally that this method goes on for either a very long time or indefinitely for moderate and large JSON strings that previously displayed fine in earlier versions of the library.

I took a stab at implementing a non recursive solution and arrived at this implementation for `simpleSpanJson()` which should be a drop in replacement for the most part.

```
    private enum class TokenType(val delimiters: Set<Char>) {
        STRING(setOf('"')),
        ARRAY(setOf('[', ']')),
        OBJECT(setOf('{', '}')),
        KEY_SEPARATOR(setOf(':')),
        VALUE_SEPARATOR(setOf(',')),
        NONE(setOf());

        companion object {
            val allPossibleTokens = values().map { it.delimiters }.flatten().toSet().toTypedArray().toCharArray()
        }
    }

    private fun CharSequence.indexOfNextToken(
        startIndex: Int = 0
    ): Pair<Int, TokenType> {
        var index = indexOfAny(TokenType.allPossibleTokens, startIndex)
        val tokenType = when (this[index]) {
            in TokenType.ARRAY.delimiters -> TokenType.ARRAY
            in TokenType.OBJECT.delimiters -> TokenType.OBJECT
            in TokenType.KEY_SEPARATOR.delimiters -> TokenType.KEY_SEPARATOR
            in TokenType.VALUE_SEPARATOR.delimiters -> TokenType.VALUE_SEPARATOR
            in TokenType.STRING.delimiters -> TokenType.STRING
            else -> null
        }
        tokenType?.let {
            return index to it
        }
        return -1 to TokenType.NONE
    }

    private fun CharSequence.indexOfNextUnescapedQuote(startIndex: Int = 0): Int {
        var index = indexOf('"', startIndex)
        while (index < length) {
            if (this[index] == '"' && (index == 0 || this[index - 1] != '\\')) {
                return index
            }
            index = indexOf('"', index + 1)
        }
        return -1
    }
    public fun simpleSpanJson(input: CharSequence): SpannableStringBuilder {
        // First handle the pretty printing step via gson built-in support
[...]

PR #1024, opened 2023-04-26
Fixed JSON pretty color printed for long text

kmayoral:
## :camera: Screenshots
Screenshot_20230426_110559

Updated with boolean value color formatting:

Screenshot_20230426_121454

## :page_facing_up: Context
Per this [issue](https://github.com/ChuckerTeam/chucker/issues/1017), I noticed that the current implementation of the JSON color formatting logic would fail to ever complete when presented with a moderately large sized JSON string to format with colors.

## :pencil: Changes
I updated the SpanTextUtil implementation to implement `fun simpleSpanJson(input: CharSequence): SpannableStringBuilder` to match the existing signature of `fun spanJson(input: CharSequence): SpannableStringBuilder`. The difference is that instead of relying on a recursive solution to format the string, we simply process the string token by token which in my tests have allowed for the large json strings to be processed quickly.

## :paperclip: Related PR
N/A

## :no_entry_sign: Breaking
N/A - same method signature

I had included a unit test with a large string that was failing to complete the test with the old implementation and completing quickly with the new implementation but I removed it due to linting errors for now.

## :stopwatch: Next steps
N/A


kmayoral:
Here are some screenshots of some local testing I performed when supplying the following long json string to the tests:
https://pokeapi.co/api/v2/pokemon/1/

Screenshot 2023-04-26 at 11 23 03 AM
Screenshot 2023-04-26 at 11 22 59 AM

kmayoral:
I'd also be happy to refactor the changes to remove the old implementation and keep the new one as `spanJson()` if we decide to move forward with this.

kmayoral:
I went ahead a made a few more changes:
- added support for boolean value color formatting
- remove legacy `spanJson()` and renamed this impl to match
- added a single test containing a long json string that failed to complete in the legacy implementation but completes in several seconds with this impl

cortinico (approved):
Change looks great to me, thanks for adding this @kmayoral

kmayoral:
Happy to help, thanks for a great library!

PR #1031, opened 2023-05-03
Fix HAR output issue

kmayoral:
## :camera: Screenshots
Screenshot 2023-05-03 at 10 34 22 AM

## :page_facing_up: Context
The existing implementation for the HAR output of transactions had a small bug in that transaction entries with an empty response body were being converted to HAR entries with a missing `content` object, violating the HAR spec. 

## :pencil: Changes
Updated the HAR output logic to render an empty content object in the output JSON for transactions with empty transaction bodies. 

## :paperclip: Related PR
N/A

## :no_entry_sign: Breaking
N/A

I dev tested this by generating a HAR file from the list of transactions within my app before/after these changes were made. When testing before and attempting to open the generated HAR file within the Chrome Network Tools debugger, the file was unable to be parsed, and when I tested after, it parsed and navigated successfully.

Screenshot 2023-05-03 at 11 05 34 AM

## :stopwatch: Next steps
Potential future improvements could be to use a non pretty-printed Gson serializer object to output the HAR contents to save on whitespace inflating the output file size.


cortinico (approved):
Thank you 🙏

PR #1047, opened 2023-05-26
Advanced Filter - I

ArjanSM:
## :camera: Screenshots
@cortinico Not sure if this is what you had in mind

## :page_facing_up: Context

Refer #286 
## :pencil: Changes

I'm not super proud of it but I'm using Regex to identify the custom search query pattern 🤦 
I can elaborate it to enable search on the basis of protocol and other factors but thought of getting this reviewed.
## :no_entry_sign: Breaking

None

Unit test included

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
@@ -9,9 +9,20 @@ import com.chuckerteam.chucker.internal.support.distinctUntilChanged
 internal class HttpTransactionDatabaseRepository(private val database: ChuckerDatabase) : HttpTransactionRepository {
 
     private val transactionDao get() = database.transactionDao()
-
+    private val customQueryRegex = """method *: *(get|put|post|patch|delete) *"""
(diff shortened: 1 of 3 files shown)


vbuberen:
Would like to share some feedback as well.
It feels a bit weird to type method search. As a user I would prefer to have some filtering view with checkboxes, like the one I described in the issue you referenced. This way users won't be confused by what they can type in search view and have a visible set of filters right away. With current implementation you showed I feel like most users won't even notice that there is such functionality.

ArjanSM:
> 
@vbuberen 
I agree that discovery of the method search in this pr is something that would require developer education. 
Maybe, I misunderstood what @cortinico meant in this [comment](https://github.com/ChuckerTeam/chucker/issues/286#issuecomment-1518737198). 
I also understand that you'd wanted to expose the filters using a `BottomSheet`.
This is what I'd worked before I thought of implementing the changes in this PR.

ArjanSM:
> > 
> 
> @vbuberen I agree that discovery of the method search in this pr is something that would require developer education. Maybe, I misunderstood what @cortinico meant in this [comment](https://github.com/ChuckerTeam/chucker/issues/286#issuecomment-1518737198). I also understand that you'd wanted to expose the filters using a `BottomSheet`. This is what I'd worked on before I thought of implementing the changes in this PR. 

Still a Work In Progress. But this is where I'm at.

PR #1075, opened 2023-07-10
Use isVisible to simplify code

filograno:
## :page_facing_up: Context
This PR simplify some parts of the code by using `isVisible` instead of `visibility` View property.

## :pencil: Changes
This is a very tiny refactoring so no logic or UI change has been done. Anyway part of the code touched are:
- the visibility logic of the buttons inside the sample apps
- the visibility of the toolbar search summary
- the visibility of GraphQL icon and GraphQL path TextView


cortinico (approved):
Thank you very much

PR #988, opened 2023-03-18
Automatic Scroll on Search added

4shutosh:
## :camera: Screenshots

Screenshot 2023-03-18 at 4 43 24 PM

## :page_facing_up: Context

Scroll to Search content added, fixes #388 
Untitled

## :pencil: Changes

Scroll to search item in response screen added

## :no_entry_sign: Breaking

No

## :hammer_and_wrench: How to test

Open Response Screen, ideally where the whole body covers more than the screen size, and search for something that is not visible in the current scroll state to see the change. 

After searching if found, the recycler view is automatically scrolled to the position, making it easy to test/evaluate the response. Earlier, just the text was being highlighted making it hard to find in a lengthy response.

PR #1076, opened 2023-07-13
Bug fix - Scroll to Search Navigation Bar Visibility Fix

4shutosh:
## :page_facing_up: Context

Issue - Search for something on the response page, the search results are visible along with the nav bar, but when we append something to the query, that leads to zero search results, highlighting is reset, but the nav bar with old search results is still visible.

## :paperclip: Related PR

#988 

- Search for a word available on the response page, and append something making it a query that won't have any results. the nav bar and the highlighted results should be reset.


cortinico:
> but the nav bar with old search results is still visible.

Do you have a gif or a video of this bug?

4shutosh:
Yes, this is the gif for the bug behaviour
ezgif com-resize

cortinico (approved):
LGTM 👍

Issue #1164, opened 2024-01-24
With notifications disabled, Chucker will still ask for notification permission when launching the activity


joakimtall:
After adding Chucker like this

```
val chuckerInterceptor = ChuckerInterceptor.Builder(context)
    .collector(ChuckerCollector(context, showNotification = false))
    .build()
OkHttpClient.Builder()
    .addInterceptor(chuckerInterceptor)
    ...
```

And then launching it by

```
context.startActivity(Chucker.getLaunchIntent(context))
```

Chucker will then ask for notification permission, and then send no notifications

joakimtall:
I could probably make a PR for this if I got pointers on how to implement it

cortinico:
Yup so the problem is here:
https://github.com/ChuckerTeam/chucker/blob/74b2a857567b92ba7e2f59e136f3dbcce38f5765/library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt#L99-L131

We'll have to check from the `ChuckerCollector` if `showNotification == true` before requesting permission

joakimtall:
How can we get a reference to the ChuckerCollector instance? If you want me to take a stab at it

cortinico:
I guess we'll have to create a new field inside the `Chucker` object to store this information

joakimtall:
I have made a PR according to this idea

PR #1165, opened 2024-01-24
Disable notification permission request when showNotification is false

joakimtall:
Proposal to fix https://github.com/ChuckerTeam/chucker/issues/1164

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
@@ -19,6 +19,8 @@ import com.chuckerteam.chucker.internal.ui.MainActivity
 public object Chucker {
     private const val SHORTCUT_ID = "chuckerShortcutId"
 
+    internal var showNotifications: Boolean = true
+
     /**
      * Check if this instance is the operation one or no-op.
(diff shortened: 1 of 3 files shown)


cortinico:
@joakimtall can you update the CHANGELOG.md file as well?

cortinico (approved):

library/src/main/kotlin/com/chuckerteam/chucker/api/Chucker.kt
```diff
@@ -19,6 +19,8 @@ import com.chuckerteam.chucker.internal.ui.MainActivity
 public object Chucker {
     private const val SHORTCUT_ID = "chuckerShortcutId"
 
+    internal var showNotifications: Boolean = true
```
cortinico, line 22: Can you add a short comment here also?

joakimtall:
> @joakimtall can you update the CHANGELOG.md file as well?

Done!

Thanks for early approval! But please take a look if the comment in Chucker.kt is what you had in mind

cortinico:
> > @joakimtall can you update the CHANGELOG.md file as well?
> 
> Done!
> 
> Thanks for early approval! But please take a look if the comment in Chucker.kt is what you had in mind

Yup thank you!

PR #1172, opened 2024-02-03
Update README.md to recommend usage as network interceptor

ColtonIdle:
## :page_facing_up: Context
Seems like people want chucker to work as a charles replacement. Even something "simple" such as cache hits won't show as cache hits in chucker while they will in charles. I will admittedly say that I don't have a deep understanding of network vs app interceptor, so I guess someone with more info can argue for an app interceptor vs network. but I think networkInterceptor is probably a better default?

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ To start using Chucker, just plug in a new `ChuckerInterceptor` to your OkHttp C
 
 ```kotlin
 val client = OkHttpClient.Builder()
-                .addInterceptor(ChuckerInterceptor(context))
+                .addNetworkInterceptor(ChuckerInterceptor(context)) // You can choose to use Chucker as either an application or network interceptor, depending on your requirements.
                 .build()
 ```
 


cortinico (commented):
Isn't this already explained in the FAQs here?
https://github.com/ChuckerTeam/chucker/?tab=readme-ov-file#faq-

ColtonIdle:
Yeah, don't disagree. Just feel like people should be told a bit earlier that they should make a decision between network and app interceptor.

cortinico:
> Yeah, don't disagree. Just feel like people should be told a bit earlier that they should make a decision between network and app interceptor.

I agree but I won't change the default snippet as we probably want users to default to use Chucker as application interceptor

ColtonIdle:
OOoh. Okay... any reason why you think it should stay as an app level interceptor by default? Maybe that would help me connect the dots here. Thanks

cortinico:
> OOoh. Okay... any reason why you think it should stay as an app level interceptor by default? Maybe that would help me connect the dots here. Thanks

It really depends on a case by case basis. I don't have an extremely strong opinion.
Perhaps we can just have both examples one below the other in the top snippet?

ColtonIdle:
Yeah. I think that's my point. I don't think one way is explicitly better than the other, but even for me (someone that has used chucker a ton) just comes here, and copies pastes the first example. Would be just nice to be reminded that there's a decision to be made there IMO.

cortinico:
> and copies pastes the first example

True. Yet having them one on top of the other should make things really clear

ColtonIdle:
But anyway, thanks for hearing me out @cortinico 

I'll close this for now, and maybe we decide on the readme later

PR #1211, opened 2024-04-10
Escape quotes in curl share

madlymad:
## :camera: Screenshots
Share the curl and paste it into postman
Before my change you get that one:
Screenshot 2024-04-10 at 16 35 20

After my change you get that one
Screenshot 2024-04-10 at 16 35 04

## :page_facing_up: Context

In our project/server we have a header that contain a JSON formatted text and every time someone try to use the curl from chucker the output curl is not working and result to errors.

## :pencil: Changes

Will update the PR with that...

## :paperclip: Related PR

No relations

## :no_entry_sign: Breaking

No breaking changes

Ideally with a curl that will have a header with quotes `"`

## :stopwatch: Next steps

No


christiancastaldi (approved):
thanks for the fix!

PR #1214, opened 2024-04-21
Add save all transactions functionality

irakliy01:
## :camera: Screenshots
image

## :page_facing_up: Context
I usually test my apps on an emulator that isn't connected to Google services. This means I can't directly share transactions with others, as I don't have suitable apps for that purpose. Instead, it's much easier for me to save transactions locally and then share the file from my computer using Device Explorer.
Chucker only supported sharing transaction details as text or HAR but lacked direct file saving capabilities (you could save only a one request/response body).
This PR adds a feature to save all transactions on the local storage. New save buttons are located within share menu separated from share buttons group by divider.

## :pencil: Changes

- Introduced "Save as Text" and "Save as HAR" buttons to the share menu, grouped together and visually separated from existing share options.
- Implemented save-to-file logic, allowing users to save transactions data locally in either plain text or HAR format.
- Developed a FileSaver object to streamline the file-saving process within the library.
- Included a JUnit test to ensure the FileSaver.saveFile method writes files correctly to the given URI and with the expected content.

Click on the share button in the app bar and then click on _save as text_ / _save as .har file_. Check created files in a directory you specified.

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/FileSaver.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/FileSaver.kt
@@ -8,7 +8,18 @@ import okio.Source
 import okio.buffer
 import okio.sink
 
+/**
+ * Utility class to save a file from a [Source] to a [Uri].
+ */
 public object FileSaver {
+    /**
+     * Saves the data from the [source] to the file at the [uri] using the [contentResolver].
+     *
+     * @param source The source of the data to save.
+     * @param uri The URI of the file to save the data to.
+     * @param contentResolver The content resolver to use to save the data.
+     * @return `true` if the data was saved successfully, `false` otherwise.
+     */
     public suspend fun saveFile(
         source: Source,
         uri: Uri,
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/FileSaver.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/FileSaver.kt
@@ -4,7 +4,6 @@ import android.content.ContentResolver
 import android.net.Uri
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
-import okio.BufferedSink
 import okio.Source
 import okio.buffer
 import okio.sink
@@ -19,7 +18,7 @@ public object FileSaver {
             runCatching {
                 contentResolver.openOutputStream(uri)?.use { outputStream ->
                     outputStream.sink().buffer().use { sink ->
-                        saveToFile(sink, source)
+                        sink.writeAll(source)
                     }
                 }
             }.onFailure {
@@ -28,12 +27,5 @@ public object FileSaver {
             }
             return@withContext true
         }
-    }
-
-    private fun saveToFile(
-        sink: BufferedSink,
-        source: Source,
-    ) {
-        sink.writeAll(source)
     }
 }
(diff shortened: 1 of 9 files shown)


library/src/test/kotlin/com/chuckerteam/chucker/internal/support/FileSaverTest.kt
```diff
@@ -0,0 +1,48 @@
...
+            val file = uri.toFile()
+
+            file.createNewFile()
+
+            FileSaver.saveFile(source, uri, contentResolver)
+
+            assertThat(file.length()).isEqualTo(TEST_FILE_CONTENT.length)
+            assertThat(file.readText()).isEqualTo(TEST_FILE_CONTENT)
+
+            file.delete()
```
cortinico, line 40: Why do you delete the file here? YOu should instead use a JUnit temp folder if you're afraid this will pollute other test runs

library/src/main/kotlin/com/chuckerteam/chucker/internal/support/FileSaver.kt
```diff
@@ -0,0 +1,31 @@
...
+import okio.buffer
+import okio.sink
+
+public object FileSaver {
+    public suspend fun saveFile(
+        source: Source,
+        uri: Uri,
+        contentResolver: ContentResolver,
+    ): Boolean {
+        return withContext(Dispatchers.IO) {
```
cortinico, line 17:
```suggestion
    ): Boolean = withContext(Dispatchers.IO) {
```

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -415,33 +424,20 @@ internal class TransactionPayloadFragment :
...
-                                    ?: throw IOException(TRANSACTION_EXCEPTION)
-                            }
-                        }
-                    }
-                }
-            } catch (e: IOException) {
-                Logger.error("Failed to save transaction to a file", e)
-                return@withContext false
+    ): Source {
+        return when (type) {
```
cortinico, line 431:
```suggestion
    ): Source = when (type) {
```

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
```diff
@@ -248,6 +281,93 @@ internal class MainActivity :
...
+                    HAR -> R.string.chucker_save_har_http_confirmation
+                },
+            ),
+            onPositiveClick = {
+                when (exportType) {
+                    TEXT -> saveTextToFile.launch(EXPORT_TXT_FILE_NAME)
+                    HAR -> saveHarToFile.launch(EXPORT_HAR_FILE_NAME)
+                }
+            },
+            onNegativeClick = null,
```
cortinico, line 308: this is unneceesary?

cortinico (approved):
Thanks for sending this over @irakliy01 
I believe we can include this 👍

library/src/main/kotlin/com/chuckerteam/chucker/internal/support/FileSaver.kt
irakliy01: Corrected

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
irakliy01: Corrected

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/MainActivity.kt
irakliy01: There is no action on dismiss (old actions also set this listener as null)

irakliy01:
@cortinico updated PR

cortinico (approved):

Issue #1237, opened 2024-06-26
Method skipPaths on builder does not work as expected


mikegr:
## :writing_hand: Describe the bug
Using `skipPaths` does not work as expected. When using the builder the path is added by using the `addPathSegment` method on the `HttpUrl.Builder`, which will escape any slash to `%F2`. You should use the `addPathSegments` to allow a full path. Because later the full "encodedPath" of the request url is used to compare. 

See line 
https://github.com/ChuckerTeam/chucker/blob/main/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt#L193
and 
https://github.com/ChuckerTeam/chucker/blob/main/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt#L75

## :bomb: Steps to reproduce
chucker = "4.0.0"

```
                ChuckerInterceptor.Builder(app)
                    .skipPaths("/f1/proxy/ggw/posthog-proxy/public/capture")
                    .build()
```

## Result: 
<img width="859" alt="Bildschirmfoto 2024-06-26 um 12 08 55" src="https://github.com/ChuckerTeam/chucker/assets/47275/8bacfe77-2840-46cd-ae13-9607696ebb70">

VenomVendor:
The path is added as single `segment` instead of `segments`

https://github.com/ChuckerTeam/chucker/blob/9a2487605d052d20a124e794560716c61e1e1d0e/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt#L193

faraz152:
I'd like to work on this. The fix is straightforward — `addPathSegment` encodes slashes as `%2F`, so multi-segment paths like "/f1/proxy/ggw/...` never match the request's `encodedPath`. Changing to `addPathSegments` preserves the path structure. Will have a PR up shortly.

faraz152:
Looks like this was already fixed by PR #1290 (merged Oct 2024) which introduced `addNonBlankPathSegments`. The issue can probably be closed. 🙂

cortinico:
Oh yeah, thanks for finding htis out @faraz152

PR #1290, opened 2024-10-03
Solution: method skipPaths on builder does not work as expected, issue #1237

Ridje:
## :page_facing_up: Context
The [Issue](https://github.com/ChuckerTeam/chucker/issues/1237) occurs when we add path segments delimited by '/' as part of the filtering mechanism, and HttpUrl treats them as a single, large path segment.

## :pencil: Changes
I added methods to safely add skips_path.

### Why not use HttpUrl::addPathSegments?
The problem with addPathSegments is that it replaces a leading '/' with an empty path segment, turning '/' into '//'. This causes issues for HttpUrl in general.

#### For example
Both MockWebServer::url and HttpUrl.Builder::resolve don't handle '//' segments properly.
```kotlin
private fun executeRequestForPath(
        okHttpClient: OkHttpClient,
        path: String,
        responseBody: String,
    ) {
        val httpUrl =
            HttpUrl.Builder()
                .scheme("https")
                .host("testexample.com")
                .addPathSegments(path)
                .build()

        val request = Request.Builder().url(server.url(httpUrl.encodedPath)).build()
        server.enqueue(MockResponse().setBody(responseBody))
        okHttpClient.newCall(request).execute().readByteStringBody()
  }
```

| Path                            |  request.url                             | 
| -- |--|
| "" | http://localhost:52347/ |
| "     "| http://localhost:52347/%20%20%20%20|
| "/skip/path"| http://skip/path |
| "//skip//path" | http://skip//path |

I consider these results problematic for our tests as the logic is changed here.

Even if we found a workaround, user inputs would still be transformed from '/path/to/test' to '//path/to/test', which leads to unexpected behavior.

### Comparison table for user input conversion using HttpUrl method and custom extension

| User filter                             | with HttpUrl::addPathSegments                               | with addNonBlankPathSegments    |
|--------------------------------------------|----------------------------------------------|----------------------------------------------|
| ""                                         | /                                            | /                                            |
| "    "                                     | /%20%20%20%20                                |                                              |
| "example"                                  | /example                                     | /example                                     |
| "www.example.com/skip/path"                | /www.example.com/skip/path                   | /www.example.com/skip/path                   |
| "example.com/skip/path"                    | /example.com/skip/path                       | /example.com/skip/path                       |
| "90"                                       | /90                                          | /90                                          |
| "https://example/"                         | /https://example/                            | /https:/example                              |
| "/skip/path"                               | //skip/path                                  | /skip/path                                   |
| "/skip//"                                  | //skip//                                     | /skip                                        |
| "http://localhost:8080/skip/path/ext"      | /http://localhost:8080/skip/path/ext          | /http:/localhost:8080/skip/path/ext          |
|"//skip//path" | ///skip//path// | /skip/path |

## :no_entry_sign: Breaking
Duplicated and empty path segments will be ignored once this change is applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Please add your entries according to this format.
 * Kotlin to 2.0.20
 * AGP to 8.5.2
 * Fix Toolbar is not accessible on api level 35 [#1280]
+* Fixed the `skipPaths` method unexpectedly modified the passed arguments [#1237]
 
 ### Added
 * Added _save as text_ and _save as .har file_ options to save all transactions [#1214]
--- a/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/api/ChuckerInterceptor.kt
@@ -200,7 +200,9 @@ public class ChuckerInterceptor private constructor(
         /**
          * Sets a list of [String] to skip paths. When any of the [String] matches
          * a request path, the request will be skipped.
-         * NOTE: empty path segments like '/', '//' and so on are not supported and will be skipped
+         *
+         * **Note:** An empty path will be treated as the '/'.
+         * '//' will also be treated as the '/' path.
          */
         public fun skipPaths(vararg paths: String): Builder =
             apply {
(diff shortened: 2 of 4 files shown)


cortinico (approved):
Great work! Thank you 🙏 Can you also update the CHANGELOG.md file @Ridje

Ridje:
Updated:
1. Modified commentary.
2. Changed listen in CHANGELOG.MD.

Issue #1334, opened 2024-12-30
java.lang.IndexOutOfBoundsException


iamarjun:
## :writing_hand: Describe the bug
Chucker crashes with `java.lang.IndexOutOfBoundsException` when we try to search for something in the response tab.

## :bomb: Steps to reproduce
1. Go to the Response tab in chucker
2. Click on the search button and start searching for anything in the response
3. Chucker crashes with `java.lang.IndexOutOfBoundsException`

```
Process: com.cheq.retail.uat, PID: 6919
                                                                                                    java.lang.IndexOutOfBoundsException: setSpan (26 ... 33) ends beyond length 26
                                                                                                    	at android.text.SpannableStringBuilder.checkRange(SpannableStringBuilder.java:1326)
                                                                                                    	at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:685)
                                                                                                    	at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:677)
                                                                                                    	at com.chuckerteam.chucker.internal.support.SearchHighlightUtilKt.applyColoredSpannableSubstring(SearchHighlightUtil.kt:47)
                                                                                                    	at com.chuckerteam.chucker.internal.support.SearchHighlightUtilKt.highlightWithDefinedColorsSubstring(SearchHighlightUtil.kt:36)
                                                                                                    	at com.chuckerteam.chucker.internal.ui.transaction.TransactionBodyAdapter.highlightItemWithColorOnPosition$com_github_ChuckerTeam_Chucker_library(TransactionPayloadAdapter.kt:139)
                                                                                                    	at com.chuckerteam.chucker.internal.ui.transaction.TransactionPayloadFragment.scrollToSearchedItemPosition(TransactionPayloadFragment.kt:378)
                                                                                                    	at com.chuckerteam.chucker.internal.ui.transaction.TransactionPayloadFragment.access$scrollToSearchedItemPosition(TransactionPayloadFragment.kt:47)
                                                                                                    	at com.chuckerteam.chucker.internal.ui.transaction.TransactionPayloadFragment$onQueryTextChange$1.invokeSuspend(TransactionPayloadFragment.kt:338)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTaskKt.resume(DispatchedTask.kt:221)
[...]

PR #1335, opened 2025-01-03
Fix: Search functionality in the body tab

iamarjun:
This commit addresses and fixes the search functionality in the body ([issue reference](https://github.com/ChuckerTeam/chucker/issues/1334)) tab by:

- Including all items in the search instead of filtering only body line items.
- Correcting the index used for search and highlighting.
- Clearing spans when the query string is not found.
- Resetting highlight on all items.

These changes ensure that the search works correctly and highlights the correct lines in the body tab.

## :camera: Screenshots

https://github.com/user-attachments/assets/e039c24f-cbd2-4a66-bd32-bb74629267d2

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
@@ -90,16 +90,16 @@ internal class TransactionBodyAdapter(private val onCopyBodyListener: () -> Unit
         foregroundColor: Int,
     ): List<SearchItemBodyLine> {
         val listOfSearchItems = arrayListOf<SearchItemBodyLine>()
-        items.filterIsInstance<TransactionPayloadItem.BodyLineItem>()
-            .withIndex()
+        items.withIndex()
             .forEach { (index, item) ->
+                if (item !is TransactionPayloadItem.BodyLineItem) return@forEach
                 val listOfOccurrences = item.line.indicesOf(newText)
                 if (listOfOccurrences.isNotEmpty()) {
                     // storing the occurrences and their positions
                     listOfOccurrences.forEach {
                         listOfSearchItems.add(
                             SearchItemBodyLine(
-                                indexBodyLine = index + 1,
+                                indexBodyLine = index,
                                 indexStartOfQuerySubString = it,
                             ),
                         )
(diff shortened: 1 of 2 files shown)


iamarjun:
[@cortinico](https://github.com/cortinico)
[@olivierperez](https://github.com/olivierperez)
[@vbuberen](https://github.com/vbuberen)

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -320,6 +320,7 @@ internal class TransactionPayloadFragment :
                     backgroundSpanColor,
                     foregroundSpanColor,
                 )
+            println("listOfSearchQuery: $listOfSearchQuery")
```
abhishekgupta-mset, line 323: @iamarjun please remove println

A7ak, line 323: @iamarjun, please remove println

iamarjun: done

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
```diff
@@ -90,16 +90,16 @@ internal class TransactionBodyAdapter(private val onCopyBodyListener: () -> Unit
         foregroundColor: Int,
     ): List<SearchItemBodyLine> {
         val listOfSearchItems = arrayListOf<SearchItemBodyLine>()
-        items.filterIsInstance<TransactionPayloadItem.BodyLineItem>()
```
cortinico, line 93: Why have you removed this?

cortinico (commented):
Thanks for sending this over

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadAdapter.kt
iamarjun: since `TransactionPayloadItem` have mutliple types, filtering the `BodyLineItem ` subtype and then indexing them does really give us the actual index of the item in the whole list. As a result wrong line with wrong search query gets highlighted, also the primary reason for the crash as well

cortinico: Ahh I understand. Thanks for clarifying 👍

cortinico (approved):

Issue #1339, opened 2025-01-27
Chucker API transaction display name mapper/interceptor


NickMelnykIronvest:
## :warning: Is your feature request related to a problem? Please describe
Adding simple mapper (predicated to ChuckerInterceptor/Collector) to change transaction (request) displayed name. For instance  have long legacy/proxyURL that contain real url at the end. But it is truncated (max line 4 in chucker list). So for display I would like to get HttpTransaction and provide CharSequence, i.e. display name. 

## :bulb: Describe the solution you'd like
Simple mapper for display name of API request. Default impl would match current display name.

## :bar_chart: Describe alternatives you've considered
 I have tries to implement on my own, but ChuckerCollector cannot be inherited/extended. So I do not need to change transaction itself, I only need to remap display name, which I do not found possible at the moment out of the box.

## :raising_hand: Do you want to develop this feature yourself?
- [ ] Yes
- [x] No

cortinico:
This is not easily achievable today. I think the easiest thing we could add is a `Redact-URL` feature similar to the `Redact-Header` feature we have explained here: https://github.com/ChuckerTeam/chucker#redact-header-%EF%B8%8F

NickMelnykIronvest:
@cortinico please see this PR
https://github.com/ChuckerTeam/chucker/pull/1340

PR #1340, opened 2025-01-28
### Display Name Formatter ✍️

NickMelnykIronvest:
Allows to customize transaction display name on the screen with transactions list. Could be useful to map long paths into short display names for reading convenience. Providing null values will fallback to default display name. Please note, that new formatter will take place only after transactions list screen start/restart.

```kotlin
Chucker.setHttpTransactionNameFormatter { transaction ->
    // provide CharSequence that acts as custom display name of transaction
}
```

## :page_facing_up: Context
Optional Transaction Display Name Formatter
[Display Name Formatter](https://github.com/ChuckerTeam/chucker/issues/1339)

## :pencil: Changes
-   Added transaction display name formatter [#1339]

## :no_entry_sign: Breaking
Non breaking changes.

Set any custom formatter as shown below and test null and non null values. 
Null values fallback to default display name.
```kotlin
Chucker.setHttpTransactionNameFormatter { transaction ->
    // provide CharSequence that acts as custom display name of transaction
}
```

## :stopwatch: Next steps
No plans. Potentially interface could be extended to also support custom display name in push-notifications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Please add your entries according to this format.
 
 ## Unreleased
 
+-   Added transaction display name formatter [#1339]
+
 ### Added
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ _A fork of [Chuck](https://github.com/jgilfelt/chuck)_
     - [Redact-Header 👮‍♂️](#redact-header-️)
     - [Decode-Body 📖](#decode-body-)
     - [Notification Permission 🔔](#notification-permission-)
+    - [Display Name Formatter ✍️](#display-name-formatter-)
   - [Migrating 🚗](#migrating-)
   - [Snapshots 📦](#snapshots-)
   - [FAQ ❓](#faq-)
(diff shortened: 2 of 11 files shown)


cortinico:
Thanks for sending this over @NickMelnykIronvest 

Not sure if you saw my message here: https://github.com/ChuckerTeam/chucker/issues/1339#issuecomment-2616628396

This is not doing what I've suggested. Specifically I don't think we should add this capability to the `Chucker` object

NickMelnykIronvest:
> Thanks for sending this over @NickMelnykIronvest
> 
> Not sure if you saw my message here: [#1339 (comment)](https://github.com/ChuckerTeam/chucker/issues/1339#issuecomment-2616628396)
> 
> This is not doing what I've suggested. Specifically I don't think we should add this capability to the `Chucker` object


> This is not doing what I've suggested. 

@cortinico 
Yeah, I saw your comment. 
Redact-header (or redact-url for this matter) has completely different purpose. 
This formatter instead designed to modify display name of transaction without alterating transaction/data as such.

 > Specifically I don't think we should add this capability to the `Chucker` object
 
@cortinico I do not see why. Assume you indeed agree to add ability to transaction display name, what better place you would choose set/store this interface. 
I choose Chucker because it is object and perfectly suited to store formatter. Amongst other classes we have `ChuckerInterceptor`. But modifying transaction name (path) has nothing in common with interceptor. The same applies to the `ChuckerCollector`. No other suitable publicly available interfaces are available. 


Just tell me straight, whether I could expect it merged and published anytime soon or request will be discarded (so I have to figure out other way how to use this feature in my project). If you need me to fix anyting before PR is accepted, let me know.

cortinico:
> Redact-header (or redact-url for this matter) has completely different purpose.
This formatter instead designed to modify display name of transaction without alterating transaction/data as such.

Oh I understand your use case. You want to alter the visualization of the URL **only** on the transaction list.

That's something we shouldn't be merging then as I think it's a bit too niche and specific to your use case.

As a side note, you can always override the layout of the transaction list by providing new XML resource files that override Chuckers so you could have a multiline string in the transaction list.

cortinico:
> That's something we shouldn't be merging then as I think it's a bit too niche and specific to your use case.

Closing because of this

PR #1357, opened 2025-04-27
Forcing LTR in response payload Recycler View

Younes-Charfaoui:
### After Change
image

### Before Change
image
 

## :page_facing_up: Context
Suppose you have an app that supports right-to-left (RTL) languages, such as Arabic. In that case, the chucker app will also be RTL, which can sometimes hinder the readability of the response, especially if it is a large JSON. You depend on indentation to locate certain values for debugging purposes or to confirm that specific values have been received. For that reason, I suggest this change.

## :pencil: Changes
I have forced the layout direction of the RecyclerView to be LTR always.

## :no_entry_sign: Breaking
Nothing.

Change the Sysmte language of the test device to arabic, then In the sample App, update the `AndroidManifest.xml` to include `android:supportsRtl="true"` in `application` tag, then try to launch some HTTP requests and verify the results that have some json, try this before and after applying the change in the PR.

## :stopwatch: Next steps
Not sure, but the same should happen of the overview tab, but it is less critical than this one.

### DIFF
--- a/library/src/main/res/layout/chucker_fragment_transaction_payload.xml
+++ b/library/src/main/res/layout/chucker_fragment_transaction_payload.xml
@@ -110,6 +110,7 @@
         android:paddingVertical="@dimen/chucker_doub_grid"
         android:scrollbars="vertical"
         android:visibility="invisible"
+        android:layoutDirection="ltr"
         app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"


cortinico:
> Suppose you have an app that supports right-to-left (RTL) languages, such as Arabic. In that case, the chucker app will also be RTL, which can sometimes hinder the readability of the response, especially if it is a large JSON. You depend on indentation to locate certain values for debugging purposes or to confirm that specific values have been received. For that reason, I suggest this change.

That's a good point, but maybe the header section on top should still respect the RTL? I'm not a RTL native so I can't make a good call here.

Younes-Charfaoui:
> > Suppose you have an app that supports right-to-left (RTL) languages, such as Arabic. In that case, the chucker app will also be RTL, which can sometimes hinder the readability of the response, especially if it is a large JSON. You depend on indentation to locate certain values for debugging purposes or to confirm that specific values have been received. For that reason, I suggest this change.
> 
> That's a good point, but maybe the header section on top should still respect the RTL? I'm not a RTL native so I can't make a good call here.

It also doesn't help @cortinico. As a native Arabic speaker, it would only help if the information were in Arabic. Since it is in English, it makes sense to keep it in the same direction because it follows the `key: value` format, like `date: 12/04/1997`. It would help if it were `12/04/1997: تاريخ`, but it isn’t, so it makes sense to have it as it is for all languages.

PR #1392, opened 2025-06-20
Migrate to version catalog

shivanandyadav11:


shivanandyadav11:
> awesome stuff!

Thanks @cortinico  
Working on the test case failure and code clean up, post that I will make it ready for review.

PR #1406, opened 2025-06-27
Initial commit to migrate from xml to compose.

shivanandyadav11:


shivanandyadav11:
@cortinico This is ready for review and merge.

library/api/library.api
```diff
@@ -4,13 +4,15 @@ public abstract interface class com/chuckerteam/chucker/api/BodyDecoder {
 }
 
 public final class com/chuckerteam/chucker/api/Chucker {
+	public static final field $stable I
```
cortinico, line 7: Why is this file changing?

library/build.gradle.kts
```diff
@@ -84,6 +86,11 @@ dependencies {
 
     api(libs.okhttp)
     api(libs.okhttp3.okhttp)
+
+    // Jetpack Compose
+    implementation(platform(libs.androidx.compose.bom))
+    implementation(libs.androidx.compose.material3)
```
cortinico, line 92: this shouldn't be needed

sample/build.gradle.kts
```diff
@@ -84,7 +81,7 @@ dependencies {
     implementation(libs.androidx.test.runner)
     androidTestImplementation(libs.androidx.espresso.core)
     androidTestImplementation(libs.androidx.junit)
-    debugImplementation(project(":library"))
+    implementation(project(":library"))
```
cortinico, line 84: Please don't change this

sample/src/main/kotlin/com/chuckerteam/chucker/sample/util/OpenUrlInBrowserUtil.kt
```diff
@@ -0,0 +1,52 @@
...
+ * 4. On [ActivityNotFoundException], logs the failure and shows a chooser for the generic intent.
+ */
+internal fun openUrlInBrowser(
+    context: Context,
+    url: String,
+) {
+    // Prepare Chrome-specific intent
+    val chromeIntent =
+        Intent(Intent.ACTION_VIEW, url.toUri()).apply {
+            setPackage("com.android.chrome")
```
cortinico, line 30: Why do we need to have a specific Chrome intent here?

cortinico (commented):
I've done an initial pass, but I'd like to do a more thorough review next week 👍

Looks great overall! Thanks for sending this over

library/build.gradle.kts
shivanandyadav11: We're keeping `ChuckerTheme` in the library module so it can be reused across both the library and sample modules. Since the theme uses `material3`, the dependency is needed here. Let me know if you'd prefer moving the theme to the sample module instead.

sample/build.gradle.kts
shivanandyadav11: The change to `implementation(project(":library"))` was necessary because `MainActivity.kt` (in `main/`) references `ChuckerTheme` from the library. Using only `debugImplementation` causes `./gradlew test` to fail since the library isn’t available to other variants that compile `main`. 

Happy to revert this if we move the `ChuckerTheme` to sample modules as per above comment.

sample/src/main/kotlin/com/chuckerteam/chucker/sample/util/OpenUrlInBrowserUtil.kt
shivanandyadav11: Good point, we don’t actually need the Chrome-specific intent here. I’ve refactored it to remove the Chrome-specific code and use `Intent.createChooser` instead.

library/api/library.api
shivanandyadav11: This file changed because we added Jetpack Compose. The `$stable` field is auto-generated by the Compose compiler for public APIs to help with UI stability. The pipeline failed since `apiCheck` detected this as a new public API change — fixed by running `apiDump` to update the baseline.

sample/src/main/kotlin/com/chuckerteam/chucker/sample/compose/ChuckerSampleControls.kt
```diff
@@ -0,0 +1,159 @@
...
+                        .weight(1f),
+                shape =
+                    androidx.compose.foundation.shape
+                        .RoundedCornerShape(4.dp),
+            ) {
+                Text(stringResource(R.string.export_to_file_har))
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
shivanandyadav11: Added

sample/src/main/kotlin/com/chuckerteam/chucker/sample/compose/ChuckerSampleTopBar.kt
```diff
@@ -0,0 +1,82 @@
...
+            Column {
+                Text(
+                    text = stringResource(R.string.app_name),
+                    style = MaterialTheme.typography.titleLarge,
+                )
+                Text(
+                    text = stringResource(R.string.app_name),
+                    style = MaterialTheme.typography.titleSmall,
+                    color = Color.White.copy(alpha = 0.7f),
+                )
```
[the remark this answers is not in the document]
shivanandyadav11: Yes and removed subtitle as they are same text.

sample/src/main/kotlin/com/chuckerteam/chucker/sample/util/OpenUrlInBrowserUtil.kt
```diff
@@ -0,0 +1,29 @@
...
+import android.util.Log
+import androidx.core.net.toUri
+
+/**
+ * Opens the specified [url] in the user's preferred browser.
+ *
+ * @param context The [Context] used to start the browser activity.
+ * @param url The web address to open.
+ */
+internal fun openUrlInBrowser(
```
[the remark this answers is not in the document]
shivanandyadav11: Updated.

sample/build.gradle.kts
shivanandyadav11: Done, moved `ChuckerTheme` to sample modules and changed back `implementation(project(":library"))` to `debugImplementation(project(":library"))`

library/build.gradle.kts
shivanandyadav11: Done

library/build.gradle.kts
```diff
@@ -87,6 +87,7 @@ dependencies {
 
     api(libs.okhttp)
     api(libs.okhttp3.okhttp)
+
```
cortinico, line 90: Can you amend this change?

cortinico (approved):
Just a small nit and we can merge this then 👍

library/build.gradle.kts
shivanandyadav11: Done

cortinico (approved):

PR #1411, opened 2025-07-02
Fix issue with large base_64 JSON (Request or Response)

Khudoyshukur:
## Large base64 Json loading issue
https://github.com/user-attachments/assets/6dd5c7ea-3262-40f7-92a1-fa3b3bc26266

## After the fix applied
https://github.com/user-attachments/assets/be0d750e-8ef6-4d59-a08f-61581eafe147

## :page_facing_up: Context
Actually I encountered to this issue at work while using chucker

## :pencil: Changes
Actually currently we are using recycler view to render the JSON. We are getting each line of the json (using line breaks like '\n') and giving each line to the recycler view.

The problem is that when we are working with large base64 JSONs, it does not contain line break ('\n') chars. It means that recycler view should render huge amount of text on the screen. This is causing ANR.

What I changed:
I checked each line length. If it exceeds some threshold (i put 500). Then I split this line into chunks preserving the span style of the line. With this, the recycler view item will get at most 500 chars to render, not whole base64 string

This is a bug-fix, and I think it does not affect the behavior of the app. So, i did not included this in CHANGELOG

## :paperclip: Related PR
No

## :no_entry_sign: Breaking
No

For testing purposes, i added one more request to sample app to test large base64

## :stopwatch: Next steps
-----


Khudoyshukur:
@cortinico Fixed, can you please check one more time

Khudoyshukur:
image

@cortinico 
I think there has been a network issue here. I did not touch any gradle related files

library/src/main/kotlin/com/chuckerteam/chucker/internal/ui/transaction/TransactionPayloadFragment.kt
```diff
@@ -527,3 +539,28 @@ internal class TransactionPayloadFragment :
...
+            val newEnd = if (end > endIndex) chunk.length else end - startIndex
+
+            chunk.setSpan(span, newStart, newEnd, flags)
+        }
+        result.add(chunk)
+
+        startIndex += size
+    }
+    return result
+}
```
shivanandyadav11, line 566: I think this extension function would be a great candidate for a dedicated Utils class/object. Extracting it there will make it easier to locate and write unit tests for it.

Khudoyshukur: Could not find predefined Utils class. So created one with a name 'SpannableStringExtension'. 
And I wrote few unit tests for the function.

Can you please check it again?

shivanandyadav11 (approved):

cortinico (approved):
That's great thank you very much for sending this over!

PR #1432, opened 2025-07-16
init

Dante291:
## :page_facing_up: Context

I have created a temp folder called 'util' (sorry couldn't come up with better name) where I have pasted my code.

**DummyChain**: custom implementation of OkHttp's Interceptor.Chain that simulates a network call by immediately returning a pre-defined Response instead of actually executing a real HTTP request

**HttpLogPayload**:  is a data structure designed to carry all the necessary details of an HTTP request and its corresponding response (like URL, method, headers, body, status, timestamps) from an external source (Flutter) to the Android host

**forwardHttpLogToChucker** :  is a method that converts an HttpLogPayload (external HTTP data) into OkHttp Request and Response objects, then uses a DummyChain to feed these objects to Chucker's ChuckerInterceptor for logging, effectively displaying external HTTP transactions in the Chucker UI


Dante291:
@cortinico integrated dummy http call with the chucker, you will be able build the sample app and will be able to see a button in screen "Do Flutter HTTP Call" you just have to click on it and you will be able too see a HTTP call in chucker, not able to see response body in this either.

PR #1539, opened 2025-12-15
Extend search functionality to include body and headers

erizoo:
- Add search support for requestBody and responseBody fields
- Add search support for requestHeaders and responseHeaders (both header names and values)
- Maintain existing GraphQL operation name search functionality
- Update SQL query in HttpTransactionDao to search across all relevant fields
- Update MainViewModel to pass searchQuery for both code and text searches
- Add comprehensive tests for body and headers search functionality

This enhancement allows users to search transactions not only by URL path and GraphQL operation name, but also by content in request/response bodies and headers, making it easier to find specific transactions.

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/data/repository/HttpTransactionDatabaseRepository.kt
@@ -15,15 +15,11 @@ internal class HttpTransactionDatabaseRepository(
         code: String,
         path: String,
     ): LiveData<List<HttpTransactionTuple>> {
-        val pathQuery = if (path.isNotEmpty()) "%$path%" else "%"
+        val codeQuery = if (code.isNotEmpty()) "$code%" else "%"
+        val searchQuery = if (path.isNotEmpty()) "%$path%" else "%"
         return transactionDao.getFilteredTuples(
-            "$code%",
-            pathQuery = pathQuery,
-            /*
-             * Refer <a href='https://github.com/ChuckerTeam/chucker/issues/847">Issue #847</a> for
-             * more context
-             */
-            graphQlQuery = pathQuery,
+            codeQuery = codeQuery,
+            searchQuery = searchQuery,
         )
     }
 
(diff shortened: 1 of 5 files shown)


ibrahim-iqbal (commented):
Read this carefully. The refactor half is a nice simplification (single `searchQuery` param, six columns behind it) and the tests cover the new surfaces well. A few things worth flagging on the SQL half before it merges:

**Performance — full scan on potentially large text columns**

`requestBody` and `responseBody` are `TEXT` columns that regularly hold multi-KB payloads (Chucker's whole point). `LIKE '%searchQuery%'` on those two columns, in a query that ORs across six columns, forces SQLite to:

- fetch every row's `requestBody` + `responseBody` blob into memory (not just the projected `HttpTransactionTuple` columns), and
- run a naive substring match through each.

Chucker DBs used in a debug session commonly reach a few thousand transactions. A search on 3-4 MB of accumulated body text will feel noticeably laggy on a real device, especially on emulator/older hardware, and it fires on every keystroke via `LiveData` recomposition. Worth measuring before shipping. A cheap first mitigation: keep header/body search behind a user-facing toggle (default off) so the fast path stays path+GraphQL-only.

**LIKE metacharacter escaping**

The user's query flows straight into `%query%` without escaping SQL LIKE metacharacters (`%`, `_`, `\\`). Real cases that break:

- Searching for `50%` (a percentage) matches every row that contains `50` followed by any characters.
- Searching for `id_123` matches `id0123`, `idA123`, etc.

SQLite supports `LIKE ... ESCAPE`, so `LIKE :searchQuery ESCAPE '\\\\'` in the query plus a `.replace(\"\\\\\", \"\\\\\\\\\").replace(\"%\", \"\\\\%\").replace(\"_\", \"\\\\_\")` on the query string at the repository boundary would make it literal-match. Non-blocking for the initial ship, but the current behaviour will surprise anyone who searches for a `%`-containing status code or an underscore-delimited request id.

**Header search on serialized string**

`requestHeaders` and `responseHeaders` are persisted as JSON (via `HttpHeader`'s serializer), so `LIKE '%text%'` on the raw column string matches both header names and header values transparently — which is the described intent. But it also matches on JSON structural chars, so a query of `\"name\"` accidentally matches every row. Same escape story.

**Repository rename**

Dropping `graphQlQuery` from the DAO signature is a public-adjacent internal change; anything reflecting on the DAO signature (unlikely, but detekt/binary-compat plugins do exist in this repo's CI) would notice. The rename to `searchQuery` reads clearer either way.

**Tests**

The 149 lines added to `HttpTransactionDaoTest` cover the four new columns (`requestBody`, `responseBody`, `requestHeaders`, `responseHeaders`) plus the merged path/GraphQL behaviour. A LIKE-escape regression case (e.g. inserting one row whose body contains `%` and searching for a literal `%`) would additionally lock in the behaviour if you take on the escape fix.

The user-facing value here is real — being able to grep for a header value across a session is exactly the thing Chucker is missing today. Just want the performance and escape questions answered before it goes live.

PR #1603, opened 2026-04-08
Fix NPE in Gson when serializing request headers on cold start

### LINKED ISSUE #1602: NPE in TypeVariableImpl.hashCode when serializing request headers (Gson) on cold start
### Description
Sporadic crash on app cold start (after the app has been unused for a long time) inside `ChuckerInterceptor` while serializing request headers with Gson. The crash originates from `libcore.reflect.TypeVariableImpl.hashCode()` returning NPE, which propagates through `GsonTypes.resolve` → `CollectionTypeAdapterFactory.create` → `HttpTransaction.setRequestHeaders`.

It is not reproducible after restarting the app — happens only on the very first request after a long idle period.

### Chucker version
4.3.1

### OkHttp / Gson versions
- OkHttp: 5.3.2 (transitive via io.ktor:ktor-client-okhttp:3.4.1)
- Gson: 2.13.2 (transitive via Chucker)

### Android version / Device
 Android 16, Pixel 7

### Stack trace
```
java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Object.hashCode()' on a null object reference
	at libcore.reflect.TypeVariableImpl.hashCode(TypeVariableImpl.java:47)
	at java.util.HashMap.hash(HashMap.java:337)
	at java.util.HashMap.getNode(HashMap.java:575)
	at java.util.HashMap.get(HashMap.java:563)
	at com.google.gson.internal.GsonTypes.resolve(GsonTypes.java:357)
	at com.google.gson.internal.GsonTypes.resolve(GsonTypes.java:400)
	at com.google.gson.internal.GsonTypes.resolve(GsonTypes.java:344)
	at com.google.gson.internal.GsonTypes.getSupertype(GsonTypes.java:290)
	at com.google.gson.internal.GsonTypes.getCollectionElementType(GsonTypes.java:311)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory.create(CollectionTypeAdapterFactory.java:50)
	at com.google.gson.Gson.getAdapter(Gson.java:627)
	at com.google.gson.Gson.toJson(Gson.java:927)
	at com.google.gson.Gson.toJson(Gson.java:898)
	at com.google.gson.Gson.toJson(Gson.java:847)
	at com.google.gson.Gson.toJson(Gson.java:824)
	at com.chuckerteam.chucker.internal.data.entity.HttpTransaction.setRequestHeaders(HttpTransaction.kt:164)
	at com.chuckerteam.chucker.internal.data.entity.HttpTransaction.setRequestHeaders(HttpTransaction.kt:160)
	at com.chuckerteam.chucker.internal.support.RequestProcessor.processMetadata(RequestProcessor.kt:36)
	at com.chuckerteam.chucker.internal.support.RequestProcessor.process(RequestProcessor.kt:24)
	at com.chuckerteam.chucker.api.ChuckerInterceptor.intercept(ChuckerInterceptor.kt:88)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:126)
	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:226)
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:574)
```

### Steps to reproduce
1. Install the app with `ChuckerInterceptor` in the OkHttp pipeline.
2. Leave the app unused for an extended period (days).
3. Open the app — the first network request crashes with the NPE above.
4. Reopening the app makes the issue disappear.

### Expected behavior
`ChuckerInterceptor` should not crash the request thread when Gson reflection on header types fails. The serialization of `Headers` should be guarded (try/catch) or use a more robust adapter.

### Notes
The crash happens inside `HttpTransaction.setRequestHeaders` (line 164) when calling `gson.toJson(...)` on the headers list. Wrapping that call defensively, or pre-registering a `TypeAdapter<List<HttpHeader>>`, would likely prevent the crash.

faraz152:
## Summary
- **Fixes #1602** — `ChuckerInterceptor` crashes on cold start with `NullPointerException` in `TypeVariableImpl.hashCode()` when Gson tries to serialize request headers via reflection
- Wraps `JsonConverter.instance.toJson(headers)` in try-catch in both `setRequestHeaders` and `setResponseHeaders` so the interceptor never crashes the request thread

## Root cause
On cold start (after long idle), Gson's internal reflection for `List<HttpHeader>` type resolution triggers a `NullPointerException` in `libcore.reflect.TypeVariableImpl.hashCode()`. This is a known JVM/Android runtime edge case where type metadata hasn't been fully initialized.

## Changes
- `HttpTransaction.kt` — Added defensive try-catch around both `setRequestHeaders` and `setResponseHeaders` serialization calls
- On failure, falls back to plain-text header representation (`name: value` per line) so the transaction is still recorded

## Test plan
- [x] Existing unit tests pass (`testDebugUnitTest`)
- [x] Build compiles successfully
- [x] Fallback produces readable header output instead of crashing
- [ ] Manual: Install with Chucker after long idle → first request should not crash

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
@@ -13,6 +13,7 @@ import androidx.room.Ignore
 import androidx.room.PrimaryKey
 import com.chuckerteam.chucker.internal.support.FormatUtils
 import com.chuckerteam.chucker.internal.support.FormattedUrl
+import com.chuckerteam.chucker.internal.support.HttpHeaderSerializer
 import com.chuckerteam.chucker.internal.support.JsonConverter
 import com.chuckerteam.chucker.internal.support.SpanTextUtil
 import com.google.gson.reflect.TypeToken
@@ -161,7 +162,7 @@ internal class HttpTransaction(
     }
 
     fun setRequestHeaders(headers: List<HttpHeader>) {
-        requestHeaders = JsonConverter.instance.toJson(headers)
+        requestHeaders = HttpHeaderSerializer.toJson(headers)
     }
 
     fun setGraphQlOperationName(headers: Headers) {
@@ -194,7 +195,7 @@ internal class HttpTransaction(
     }
 
     fun setResponseHeaders(headers: List<HttpHeader>) {
-        responseHeaders = JsonConverter.instance.toJson(headers)
+        responseHeaders = HttpHeaderSerializer.toJson(headers)
     }
 
     fun getResponseHeadersString(withMarkup: Boolean): String =
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/HttpHeaderSerializer.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/HttpHeaderSerializer.kt
@@ -1,0 +1,31 @@
+package com.chuckerteam.chucker.internal.support
+
+import com.chuckerteam.chucker.internal.data.entity.HttpHeader
+import com.google.gson.JsonArray
+import com.google.gson.JsonObject
+
+internal object HttpHeaderSerializer {
+    fun toJson(headers: List<HttpHeader>): String =
+        try {
+            JsonConverter.instance.toJson(headers)
+        } catch (
+            @Suppress("TooGenericExceptionCaught", "SwallowedException") e: Exception,
+        ) {
+            // Gson's default path resolves the List<HttpHeader> element type
+            // reflectively, which occasionally throws NPE on cold start (issue
+            // #1602). Build the JSON tree explicitly so callers still get a
+            // valid string that fromJson<List<HttpHeader>> can round-trip.
+            buildHeadersJson(headers)
+        }
+
+    private fun buildHeadersJson(headers: List<HttpHeader>): String {
+        val array = JsonArray(headers.size)
+        headers.forEach { header ->
+            val obj = JsonObject()
+            obj.addProperty("name", header.name)
+            obj.addProperty("value", header.value)
+            array.add(obj)
+        }
+        return array.toString()
+    }
+}
--- a/library/src/test/kotlin/com/chuckerteam/chucker/internal/support/HttpHeaderSerializerTest.kt
+++ b/library/src/test/kotlin/com/chuckerteam/chucker/internal/support/HttpHeaderSerializerTest.kt
@@ -1,0 +1,62 @@
+package com.chuckerteam.chucker.internal.support
+
+import com.chuckerteam.chucker.internal.data.entity.HttpHeader
+import com.google.common.truth.Truth.assertThat
+import com.google.gson.reflect.TypeToken
+import org.junit.jupiter.api.Test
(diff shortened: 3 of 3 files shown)


library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -161,7 +161,12 @@ internal class HttpTransaction(
     }
 
     fun setRequestHeaders(headers: List<HttpHeader>) {
-        requestHeaders = JsonConverter.instance.toJson(headers)
+        requestHeaders =
+            try {
+                JsonConverter.instance.toJson(headers)
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                headers.joinToString(separator = "\n") { "${it.name}: ${it.value}" }
```
cortinico: This won't work sadly. The caller will always assume the field contains a JSON string:

https://github.com/ChuckerTeam/chucker/blob/2054c4b99e228d437519f09645a3e76327850fa7/library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt#L174-L178

So you should store it as valid JSON, not as plain string.

What would happen is that you will show nothing to the user even if headers are there.

library/src/main/kotlin/com/chuckerteam/chucker/internal/data/entity/HttpTransaction.kt
```diff
@@ -194,7 +199,12 @@ internal class HttpTransaction(
     }
 
     fun setResponseHeaders(headers: List<HttpHeader>) {
-        responseHeaders = JsonConverter.instance.toJson(headers)
+        responseHeaders =
+            try {
+                JsonConverter.instance.toJson(headers)
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                headers.joinToString(separator = "\n") { "${it.name}: ${it.value}" }
```
cortinico: Once you do the other changes, can you refactor so that this code is inside a util function + add tests?

cortinico (changes requested):
Thanks for looking into this @faraz152 - there are a couple of things to fix though

faraz152:
Thanks @cortinico, good catch. Pushed an update that addresses both points:

- The fallback no longer stores a plain `name: value` string. It now builds a real JSON array via Gson's `JsonArray`/`JsonObject` tree API (which skips the reflective element-type resolution that was blowing up), so `getParsedRequestHeaders()` / `getParsedResponseHeaders()` still round-trip correctly.
- Extracted the logic into `HttpHeaderSerializer` in `internal/support` and added `HttpHeaderSerializerTest` covering: default path round-trip, empty list, fallback round-trip, and fallback escaping of quotes/control chars.
- ktlint + detekt are clean locally.

faraz152:
@cortinico — CI is also showing `action_required` here (needs a maintainer to approve the workflow run). Let me know if you'd like any changes to the `HttpHeaderSerializer` implementation while you're reviewing.

cortinico (approved):

Issue #1570, opened 2026-02-06
ANR triggered by thread waiting for a binder transaction


Marko2407:
## :writing_hand: Describe the bug

We receive ANR in our app and chrashlytics return this info
com.chuckerteam.chucker.internal.support.NotificationHelper.createClearAction
ANR triggered by thread waiting for a binder transaction

  main (native):tid=1 systid=22011 
#00 pc 0x886f0 libc.so (syscall + 32) (BuildId: 50ea6e178a2b5c764ec0df7ec85c6007)
#01 pc 0x201434 libart.so (art::ConditionVariable::WaitHoldingLocks + 136) (BuildId: 759e344ef0291bdf924e23387b5bf9ea)
#02 pc 0x95ad30 libart.so (artJniMethodEnd + 124) (BuildId: 759e344ef0291bdf924e23387b5bf9ea)
#03 pc 0x66ea5c libart.so (art_jni_method_end + 12) (BuildId: 759e344ef0291bdf924e23387b5bf9ea)
       at android.os.BinderProxy.transactNative(Native method)
       at android.os.BinderProxy.transact(BinderProxy.java:684)
       at android.app.IActivityManager$Stub$Proxy.getIntentSenderWithFeature(IActivityManager.java:7439)
       at android.app.PendingIntent.getBroadcastAsUser(PendingIntent.java:767)
       at android.app.PendingIntent.getBroadcast(PendingIntent.java:750)
       at com.chuckerteam.chucker.internal.support.NotificationHelper.createClearAction(NotificationHelper.kt:128)
       at com.chuckerteam.chucker.internal.support.NotificationHelper.show(NotificationHelper.kt:98)
       at com.chuckerteam.chucker.api.ChuckerCollector$onResponseReceived$1.invokeSuspend(ChuckerCollector.kt:78)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
       at android.os.Handler.handleCallback(Handler.java:959)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loopOnce(Looper.java:249)
       at android.os.Looper.loop(Looper.java:337)
       at android.app.ActivityThread.main(ActivityThread.java:9608)
       at java.lang.reflect.Method.invoke(Native method)
[...]

PR #1606, opened 2026-04-15
Lazy initialize and reuse clearAction in NotificationHelper

Khudoyshukur:
## :page_facing_up: Context
This PR is going to address this [issue-1570](https://github.com/ChuckerTeam/chucker/issues/1570)

## :pencil: Changes
In the issue it is reported that the ANR is coming from createClearAction() function within Notification Helper. 

From debugging with logs, I found out that every time new http request notification is posted, this function is called and a fresh pending intent is created via ActivityManagerService via a Binder IPC call. Under high concurrency (many simultaneous HTTP responses), this might be causing an ANR. 

Changes: 
1. I have moved this **clearAction** to a single variable (lazy initialized).
2. Also removed flag FLAG_ONE_SHOT since the pendingIntent will be reused multiple times after change 1

### DIFF
--- a/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
+++ b/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
@@ -48,6 +48,27 @@ internal class NotificationHelper(
         )
     }
 
+    private val clearAction by lazy {
+        val clearTransactionsBroadcastIntent =
+            Intent(context, ClearDatabaseJobIntentServiceReceiver::class.java)
+
+        val clearActionIntent =
+            PendingIntent.getBroadcast(
+                context,
+                INTENT_REQUEST_CODE,
+                clearTransactionsBroadcastIntent,
+                immutableFlag(),
+            )
+
+        val clearTitle = context.getString(R.string.chucker_clear)
+
+        NotificationCompat.Action(
+            R.drawable.chucker_ic_delete_white,
+            clearTitle,
+            clearActionIntent,
+        )
+    }
+
     init {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
             val transactionsChannel =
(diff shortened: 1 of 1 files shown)


library/src/main/kotlin/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
```diff
@@ -48,6 +48,27 @@ internal class NotificationHelper(
...
+
+        val clearActionIntent =
+            PendingIntent.getBroadcast(
+                context,
+                INTENT_REQUEST_CODE,
+                clearTransactionsBroadcastIntent,
+                immutableFlag(),
+            )
+
+        val clearTitle = context.getString(R.string.chucker_clear)
```
cortinico: no need to have this extra varialbe here, can you inline it?

library/src/main/kotlin/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
```diff
@@ -48,6 +48,27 @@ internal class NotificationHelper(
         )
     }
 
+    private val clearAction by lazy {
+        val clearTransactionsBroadcastIntent =
+            Intent(context, ClearDatabaseJobIntentServiceReceiver::class.java)
```
cortinico: no need to have this extra varialbe here, can you inline it?

cortinico (approved):
LGTM, just a couple of small nits ✅

library/src/main/kotlin/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
Khudoyshukur: done

library/src/main/kotlin/com/chuckerteam/chucker/internal/support/NotificationHelper.kt
Khudoyshukur: done

